PR #1, opened 2017-11-15
Add a minimal setup.py

abergeron:


setup.py
```diff
@@ -0,0 +1,35 @@
...
+from fnmatch import fnmatchcase
+from distutils.util import convert_path
+
+try:
+    from setuptools import setup
+except ImportError:
+    from distutils.core import setup
+
+
+def find_packages(where='.', exclude=()):
```
obilaniu: By sheer coincidence I was reading how to make a `setup.py` file myself just now.

Why not define `find_package()` only if it cannot be imported from `setuptools`? Why not use `os.walk()`? Also, given that Myia is going to be Python 3.6+-only (AFAIK), we shouldn't be hunting for `__init__.py` files anymore.

breuleux: It'll certainly be Python 3+, probably 3.6+ unless we have good reasons to support a lesser version. Can we simply assume `setuptools.find_package` exists in that case?

abergeron: @obilanu Where have you heard that python 3.6+ will not require `__init__.py`?

Apart from that I just copied the find_packages() from Theano, but it can certainly be improved.

Otherwise, if we assume setuptools, we can also drop it and just use that version.

obilaniu: @abergeron As of Python 3.3, _namespace packages_ are a thing: https://www.python.org/dev/peps/pep-0420/

PR #3, opened 2017-12-14
Graph-based ANF IR

bartvm:
Still WIP, but I made a start with the IR. Some points:

* The nodes have an object hierarchy so that you can use `isinstance`.
* I imagine that we might have situations in which we want to serialize the IR (e.g. create protobufs) and such, so trying to keep it pretty minimal.
* Edges are one-directional right now, since bidirectionality is only needed in some passes (and requires more bookkeeping when mutating the graph) I think it's best to try and keep def-use edges separate somehow
* I was thinking we can hide all the debug information in a single dictionary. Debug information would include things such as human readable names of the variable and the Python AST node that created the node from the parser.

PR #4, opened 2017-12-15
Parser

bartvm:
This is the converter/parser I committed [here](https://github.com/mila-udem/myia-private/blob/master/ast2anf.py), adapted to the current IR in #3.

Heavily WIP, just posting for early feedback.

- [x] Add documentation
- [x] Lots of tests
- [x] Get rid of `BLOCKS`/`RETURNS` hack and actually return the generated functions
- [x] Consider if an alternative for the `CONSTANTS` dictionary is needed


breuleux:
I think you're jumping the gun, here.

* We have not finalized the IR. I don't think we are close to finalizing the IR.
* I can't make this work. It would be a lot easier for me to review this code if I could see its output.
* In fact, nothing sets the graph's `return_` field, so it *can't* work.

I also see no infrastructure to properly track source code location, or to deal with global variables at all. Fortunately, I already tackled all of these issues in `master`, so perhaps it'd be better to start by incorporating the relevant code through a PR, then scaffold the `Parser` class, write a sequence of tests, then implement the operations we need.

bartvm:
This code isn't meant to be reviewed. It's simply meant to jumpstart a high
level discussion about the parser. Discussions are easier if you actually
have a diff to start from. This PR is incomplete and untested, which I
clearly mentioned in the description.

If you want to run it, you can run a function through the func2anf function
and you'll find a return node in the RETURNS constant afterwards. You can
manually walk the graph from there to confirm that it is doing the right
thing.

breuleux:
Okay, but I mean, if it's not meant to be reviewed, and you request reviews, that's a bit confusing. I don't think discussions are necessarily easier if you have a diff to start from either.

I guess my point is that I don't feel like I can review this code in this state, so I'm not sure what we're supposed to discuss here.

bartvm:
It is considered good practice to make pull requests early on to gather higher-level feedback: Are there features missing (you mentioned source code location tracking)? Are there features which are actually not necessary? Is the overall structure correct (perhaps it should be logically split in multiple PRs, multiple files, multiple classes)? Are there things that should be kept in mind going forward? It allows reviewers to steer the coding process, instead of simply nitpicking about the final result. This is sometimes called "early-pull strategy" or "continuous review". You'll find it advocated e.g. [here](https://github.com/blog/1124-how-we-use-pull-requests-to-build-github), but also in blogs like [here](https://ben.straub.cc/2015/04/02/wip-pull-request/) and [here](https://medium.com/practical-blend/pull-request-first-f6bb667a9b6).

> **Open a Pull Request as early as possible**
> Pull Requests are a great way to start a conversation of a feature, so start
one as soon as possible- even before you are finished with the code. Your
team can comment on the feature as it evolves, instead of providing all
their feedback at the very end.

I added you as a reviewer to clarify that this was intended to start a discussion. I can wait next time until the code is done.

bartvm:
@breuleux Once you have a chance, have a look at the current state of the code and let me know if you have any early feedback.

I introduced code to deal with global variables. The code is different from what you had, but only slightly, because I don't want to create actual global symbols (e.g. for testing it would be better if we can parse functions in such a way that they don't share nodes, even for built-ins).

It still uses SSA terminology, but I can easily change that if you think it will be clearer?

breuleux:
I skimmed it, I think your approach is good, but I'll read more closely today.

In the meantime, there are a few reasons I handled globals the way I did. They may not all be good reasons, but we can discuss them:

* Resolving globals during parsing means the globals must all be resolvable when parsing happens. This means that e.g. it can't occur in the `@myia` decorator, because a decorated function may need one that's defined after itself. Using global symbols, to be resolved in a later phase, relaxes that constraint and simplifies the parser.
* Delaying resolving globals might also make it easier to add some features or checks, e.g. verifying that the underlying references don't change from an execution to the next. This would otherwise require adding some bookkeeping in the parser, which isn't intrinsically bad, just less modular.
* I would rather not hardcode looking up globals in `function.__globals__` and provide the dictionary explicitly instead. In some instances we might want to evaluate a function in a hand-constructed global environment, notably when we have the function's source code or a simple expression as a string (although we can `exec` it, but that does pointless extra work).

Also, I'm not sure what you mean regarding sharing. The global symbols I have are not nodes and would live in the `value` field of a `Constant` and they are (in principle) immutable, so it isn't really any more problematic to "share" them than it is to share integers or strings. They don't have to be resolved the same way in different tests.

myia/anf_ir.py
```diff
@@ -42,6 +42,25 @@ def __init__(self) -> None:
         self.return_: Apply = None
         self.debug = GraphDebug()
 
+    def __str__(self) -> str:
```
breuleux: I'm a bit iffy on having this kind of string representation for graphs. It's tricky to do it properly, and it's even trickier to do it in a way that doesn't look like garbage for non-trivial graphs. For example, for:

```python
def f(x):
    a = x + 1
    b = x + 2
    return a + b
```

`str(parse(f))` currently gives me `f(x) -> add(a, b)` which is broken. It also busts the stack for some recursive graphs.

It's not trivial to fix either, since you only want to give `a`'s definition once even if it is used more than once, you may need to create temporary names for nameless nodes that are reused, and so on. It's work, basically. Graphical representations are easier to implement and more usable so I'm not sure it's worth bothering with textual ones at all. I'm working on porting the colourful graphs I had on master, for what it's worth.

myia/anf_ir.py
```diff
@@ -274,6 +307,20 @@ def __init__(self, graph: Graph) -> None:
...
+
+        The debugging `name` attribute will be used to output strings of the
+        form `x` if possible. Otherwise it will default to `arg0` where `0` is
+        the argument index.
+
+        """
+        if hasattr(self.debug, 'name'):
+            return self.debug.name
+        if self in self.graph.parameters:
+            return f"arg{self.graph.parameters.index(self)}"
```
breuleux: Under this scheme, an optimization that drops an argument would transform `f(arg0, arg1, arg2)` into `f(arg0, arg1)` regardless of which argument is dropped. That means optimizations can accidentally "rename" parameters and that could be confusing in some circumstances. I would prefer parameter names to be set in stone, although that does make avoiding collisions a little harder.

myia/parser.py
```diff
@@ -0,0 +1,434 @@
...
+    Functions parsed in the same environment will use the same Myia objects for
+    the same Python objects.
+
+    """
+
+    def __init__(self, object_map: Dict[Any, ANFNode],
+                 ast_map: Dict[Type[ast.AST], ANFNode]) -> None:
+        """Construct an environment."""
+        self.object_map = object_map
+        self.ast_map = ast_map
```
breuleux: Should these be different dictionaries? I mean, there's no chance of overlapping.

myia/parser.py
```diff
@@ -0,0 +1,434 @@
...
+    def process_if(self, block: 'Block', node: ast.If) -> 'Block':
+        """Process a conditional statement.
+
+        A conditional statement generates 3 functions: The true branch, the
+        false branch, and the continuation.
+
+        """
+        true_block = Block(self)
+        false_block = Block(self)
+        after_block = Block(self)
```
breuleux: You have to add the current block to `true/false_block.preds` here, or before the statements are processed in both branches. That seemed enough to make `if` work on my local copy, don't know if there's anything more to do.

myia/primops.py
```diff
@@ -0,0 +1,50 @@
...
+    execute one of the two functions, depending on the boolean value, and
+    return the result.
+
+    """
+
+    pass
+
+
+class Return(Primitive):
+    """Return primtive.
```
breuleux: *primitive

myia/parser.py
```diff
@@ -0,0 +1,434 @@
...
+            return self.block_map[block]
+        op = Constant(block.graph)
+        self.block_map[block] = op
+        return op
+
+    def read(self, varnum: str) -> ANFNode:
+        """Read a variable from the function's global namespace."""
+        assert isinstance(self.function, FunctionType)
+        if varnum in self.function.__globals__:
+            return self.environment.map(self.function.__globals__[varnum])
```
breuleux: I mentioned this elsewhere, but I think we should be able to set/override the globals dictionary we use here.

breuleux:
You may be aware of them, but to make sure, here are examples that currently don't work and should probably be the next priority (unless there's other problems I'm not aware of):

```python
def f(x):
    if x:
        return 1
    return 2
```

This will always return 2 because each branch is set to jump to the after block, but `return` is supposed to override that. This is more or less the same issue with `break` and `continue`, so ideally the same system should support all of these constructs.

```python
def f(x):
    a = 1
    if x:
        a = 2
    # else:
    #     a = 3
    return a
```

That gives a `KeyError` in `pred.jumps`, but only if the `else` branch is absent.

myia/anf_ir.py
bartvm: I agree that this quickly became too complicated. Even if we keep textual representations, I think they should be factored out into separate debugging tools.

That said, it would be useful to have a limited representation for when you're inspecting objects in `pdb` e.g. maybe something like `Graph(args=['x'], return=Node(inputs=['add', 'a', 'b']))` instead of `<myia.Graph at 0x1095da748>`?

myia/anf_ir.py
bartvm: Not sure what I think about this. Some things come to mind:

* Naming should be optional, since it is only for debugging. I don't want to have to give each variable a unique name when I'm just messing about in `pdb`, or be forced to deal with parameter naming when I'm quickly trying out an optimization that affects parameters. You can hide these requirements behind helper functions of course, but I'd like to avoid needing helper functions to instantiate nodes.
* Unique naming requires some form of scoping and/or globals, since each node name needs to be aware of all the other nodes that it cannot clash with and their names. Ew.

Given that, I am inclined to say:

* `__str__` should handle a lack of names gracefully
* But we can have our own parser and optimizations fix the name of each parameter, so that the parameter renaming problem you describe doesn't happen unless people write custom parsers/optimizations.

myia/parser.py
bartvm: Theoretically, there is:
```python
x = ast.If
def f():
    return x  # Returns the `cond` Myia primitive instead of the Python object
```
Right now we don't support arbitrary Python objects, but that could be a possibility in the future (kind of like Cython and Numba, we could just switch back and forth to the interpreter to handle Python code we can't handle).

More so than this theoretical overlap though, I prefer keeping them separate for typing/logic/clarity reasons.

lamblin:
Regarding globals, is there a way to be stricter than Python and refuse re-assignment of global variables accessed from decorated functions?

For instance, I think we discussed not allowing redefining variables that are closed over inside a decorated function, for instance:
```python
def f():
    x = 3
    def g():
        return x
    x = 4
    def h():
        return x
    return g, h

g, h = f()
assert g() == h()
```
has no chance to work exactly as Python, so we should just raise an error.

Can we do the same for globals?

myia/anf_ir.py
breuleux: Sure, we can have something like that, plus the graph's name if one is available.

myia/anf_ir.py
breuleux: New nodes can be named automatically either with the `uuid` package, or with a global monotonically increasing integer, or more simply by setting their name as `${id(parameter)}`. Also, insofar that the names are not significant in code generation, it's also fine (and more readable) to set them to random dictionary words and just ignore the small collision probability.

myia/parser.py
breuleux: Fair enough.

breuleux:
> Do we need "global symbols"? Python's variable binding in this case is entirely by name, so I think that all we need is a set of strings.

Each module has their own globals, though. Inlining a function `f` defined in `a.py` into `g` defined in `b.py` may create a graph that has to resolve some globals in one place and others in another place. That sounds a bit contrived, but imagine that e.g. we try to resolve and inline all globals that are functions at compile time, but keep global boolean flags as globals (since they are much more likely to change).

> Globals need to be resolved each time a function is called, which means recompiling. However, user-defined functions are globals, which means that for almost any program we will be endlessly recompiling programs. Can we accept that?

When a function is compiled, we can track the set of globals that were resolved and what their values were at that moment. We only need to recompile when there is a discrepancy. Function objects in Python are (I think?) immutable, so we can cache the result of compiling them.

> However, you get a really awkward coupling between the parser and the optimizer (i.e. the parser needs to be able to modify optimized code, so it can't be lowered too much).

I don't understand what you mean. In my view, the parser would produce a graph which contains `Global` nodes (or perhaps `Apply(primops.Resolve(), namespace, name)`). A different pass would resolve these globals and replace them on a *copy* of the graph, while keeping track of what it did so that it can be invalidated if needed. This requires no coupling, and the parse result can be reused.

bartvm:
> I don't understand what you mean. In my view, the parser would produce a graph which contains Global nodes (or perhaps Apply(primops.Resolve(), namespace, name)). A different pass would resolve these globals and replace them on a copy of the graph, while keeping track of what it did so that it can be invalidated if needed. This requires no coupling, and the parse result can be reused.

I was talking about the case where you want to cache an *optimized* graph with unresolved variables (to avoid reoptimizing when the globals changed). This means that the optimizer would be (a) forced to spit out a representation which can still be processed by whatever pass resolves global variables and (b) take into account the existence of unresolved variables.

If you don't optimize, then yes, what you are proposing is approximately what I had in mind as well (although instead of modifying the IR itself, I would simply use a standard graph with missing inputs and a separate list of unresolved inputs).

> When a function is compiled, we can track the set of globals that were resolved and what their values were at that moment. We only need to recompile when there is a discrepancy. Function objects in Python are (I think?) immutable, so we can cache the result of compiling them.

True, if we do the check efficiently this is probably a small overhead, so I guess it would be a minor optimization.

> Each module has their own globals, though. Inlining a function f defined in a.py into g defined in b.py may create a graph that has to resolve some globals in one place and others in another place.
 [...]

Not sure I follow this. An example of this case:

```python
# test.py
import test2
def g():
    return test2.f()
assert g() == 1
x = 2
assert g() != 2
test2.x = 2
assert g() == 2
```
```python
# test2.py
x = 1
def f():
  return x
```
When `g` is parsed, `test2` will be marked as an unresolved global. Likewise, `f` will be parsed with `x` marked as an unresolved global. When `g` is first called, the globals are resolved to the correct module (`test2`) and value (1). When `g` is called for the second time, we have to check whether the globals of `g` or any function in its call tree have changed. Since `test2.f` will be resolved to a function, we will check whether `test2.f.__globals__['x']` is still 1, which it will be. However, when `g` is called for the third time, this check fails, triggering a recompilation of `f` with the correct value for `x` (2).

Basically, the thing to note is that when we call a function, we need to know its entire call tree and check for each of the functions in there whether their `__globals__` have changed since we last compiled.

> For instance, I think we discussed not allowing redefining variables that are closed over inside a decorated function [...]

I think this is a good point. Maybe I'm wrong, but it looks like this is actually the consequence of a pretty fundamental difference in scoping and name resolution between Python and Myia's IR. A modified example of @lamblin's example that is slightly more realistic:
```python
def f(x):
    def g():
        return sum(x)
    x[:, -1] = 1
    return g()
```
Simplifying a bit, Python is lexically scoped (with forward declarations), and name resolution of free variables are resolved at runtime (using the "LEGB rule"). In our IR, free variables aren't names, they are direct references to variables (nodes) in an enclosing scope. This means the name resolution cannot happen at runtime. For globals, we can sidestep this issue by having the Python runtime in the loop and checking whether the name bindings of globals changed each time a function is called, but in general we cannot do this.

How should we address this? A language like Haskell sidesteps the issue by having immutable declarations, but I guess we cannot do this because of the functional update syntax (which requires mutable declarations). One way is to behave like Python in the cases where we can (for globals, as @breuleux suggested) and throw an error in the cases where can't (for closures, as @lamblin suggested). Are there other ways? I can't really think of any, considering we need to support forward declarations (e.g. for mutually recursive functions).

So the rule becomes "Once a variable is closed over, it can be assigned to exactly once (as an argument or assignment before or after the function definition). Any subsequent assignment will raise an error. So this works:
```python
def f():
    def g():
        return x
    x = 1
    return g
```
But this will throw an error:
```python
def f(x):
    def g():
        return x
    x = 1
    return g
```

breuleux:
> (although instead of modifying the IR itself, I would simply use a standard graph with missing inputs and a separate list of unresolved inputs).

I think it is simplest to just represent them as an application of some `resolve` primitive (which, in the context of the graph, can be considered to be a pure function). The current representation already supports this scheme, optimizations don't have to care about it at all, it is simple for a basic Python-based VM to implement that primitive, and it can eventually be optimized to an offset lookup in a global store.

```python
# test.py
import test2
def g():
    return test2.f()
```

```python
# test2.py
x = 1
def f():
  return x
```

I was talking about inlining: if `test2.f` is inlined into `g`, `g` will now return a global variable `x` that is in a different module from itself. If its body was e.g. `f() + x`, inlining would produce the graph `x@test2.py + x@test.py`. This means we either have to resolve globals prior to any inlining, or we embed a namespace in each global node, which lets us keep some of them unresolved all through optimization. That was one of the factors that weighed in the design in `master`.

> So the rule becomes "Once a variable is closed over, it can be assigned to exactly once (as an argument or assignment before or after the function definition). Any subsequent assignment will raise an error.

+1

bartvm:
I would do name resolution as the very first thing, and make sure all names
are resolved before doing any optimizations or transformations. What's the
benefit of keeping unresolved names around? The only case I can think of is
if we later on want to support runtime name resolution. Is that something
we want?

breuleux:
There are a few potential benefits:

* Several graph transforms and optimizations can operate on graphs that contain unresolved names, and their output could then be reused even when these variables change.
* If users make use of global variables that change often, and some probably will do that, we don't necessarily want to hardcode their values in the graph. Ideally, we would prefer the code we generate to be invariant to them (aside from e.g. their type). For the time being, the simplest solution for that is runtime resolution.

I mean, it's hardly complicated to keep that information around, and I've already suggested a way to do it that requires nothing more than a new primitive.

bartvm:
> I mean, it's hardly complicated to keep that information around, and I've already suggested a way to do it that requires nothing more than a new primitive.

I'm not convinced it is this trivial. You are coupling the output of the optimizations with the Python runtime. Right now this coupling is trivial, because we only have one IR and everything is done in Python, but this becomes less trivial in other cases. Just some random problems that come to mind because of this coupling:

* If speed becomes important, we might have to implement the IR and certain optimization passes in C++. We can share the IR between Python and C++ as a e.g. a protobuf, but only for primitive types. 
  * You could cast the `PyObject*` pointer of the function whose namespace you need to an `int` and use reference counting to make sure the function stays alive so that we can retrieve the global namespace of the function later during the name resolution phase (by going back to Python or calling `PyFunction_GetGlobals`). But my point is that it is a lot of Python specifics and internals that now need to be handled throughout the compiler. Ideally you just want to move to C++ and never go back.
* If `x` changes in value from 2<sup>32</sup> to 100 in Python, does it change its type from `int64` to `int32` and will we trigger recompilation? Knowing whether the "type changed" requires our IR type system to be comparable somehow with Python's.

I am not convinced that we won't see other issues over time by having this coupling. So I would still either (a) resolve names just-in-time each time a function is called from Python, or (b) turn globals into function arguments: 
```python
x = 2
@myia
def f():  # Parsed to f(global_x): return global_x
    return x
x = 3
# This is actually a call to f(f.__globals__['x'])
assert f() == 3
```
This means that besides the parser and Python calling mechanism, the IR, optimizer, codegen, etc. are completely ignorant of the existence of unresolved globals; they are just normal parameters. We already know that we need machinery to trigger recompilation when input types change. This same logic now can now automatically take care of handling a change in types in the global environment.

We then have two behaviors to choose from: treat globals as constants and recompile each time they change, or make them arguments. We can expose these two options through compiler flags or through something more intelligent.

---

A more complicated solution, but one that could eventually allow us to do name resolution at runtime even for closures, is to introduce environments. This is more similar to @breuleux's proposal of carrying around a namespace, but in a Python-agnostic manner.
```python
x = 2
@myia
def f():  # Parsed to f(env): return readenv(env, 'f.x')
    return x
x = 3
# This is actually a call to something like f(make_env([f]))
# i.e. it captures the environment of all relevant Python functions
assert f() == 3
```
Often when you define closures, you use the environment at the point of definition. To mimics Python's runtime name resolution, I think it might be enough to do something like this:
```python
def f():
    x = 3
    def g():
        return x
    x = g() + 1
    def h():
        return x
    return g, h

# Myia IR version
def g(env):
    return readenv(env, 'x')

def h(env):
    return readenv(env, 'x')

def f(env):
    env = writeenv(env, 'x', 3)
    new_x = g(env)
    env = writeenv(env, 'x', new_x)
    # The environment of `f` is finalized here, so we can create the closures with it
    return (g, env), (h, env)

g, h = f()
assert g() == h()
```
The downside of this approach is that the type inference and optimization passes would ideally both be aware of environments (so that the output type of `g` and `h` can be deduced, and ideally you want to actually optimize to `g = h = lambda: 4`).

breuleux:
> You are coupling the output of the optimizations with the Python runtime.

That depends on the resolution policy. You can still resolve all globals at the beginning if you want to, it's just not forced by the design. So I don't really understand the objection.

> If speed becomes important, we might have to implement the IR and certain optimization passes in C++. We can share the IR between Python and C++ as a e.g. a protobuf, but only for primitive types.

That may not have been clear, but my idea is to represent accessing the global `x` in module `m.py` as something like `Apply([primops.resolve, Constant("m.py"), Constant("x")])`. The namespace is a string. Some `Environment` object would serve as a middleman for resolution and would keep track of which globals are needed, which ones require recompilation, and so on.

> I am not convinced that we won't see other issues over time by having this coupling. So I would still either (a) resolve names just-in-time each time a function is called from Python, or (b) turn globals into function arguments:

Or Myia functions can have a global environment, which is quite a bit simpler than adding arguments all over the place. For example, every time a global variable is needed but its value is not relied upon, it may be associated to an integer index, then they can all be packed into an array for cheap lookup. Alternatively, put them in a closure.

bartvm:
> You can still resolve all globals at the beginning if you want to, it's just not forced by the design. So I don't really understand the objection.

I don't really follow this argument. If you want to have the _option_ of keeping names unresolved during optimization, you must still ensure that the output of the optimization pass can be readily handled by the Python name resolution pass, so my objection of introducing that constraint still stands.

> That may not have been clear, but my idea is to represent accessing the global `x` in module `m.py` as something like `Apply([primops.resolve, Constant("m.py"), Constant("x")])`. The namespace is a string. [...]

Resolving globals statically doesn't work because Python's lexical scope is entirely mutable. You will need the actual runtime environment of the calling Python interpreter.
```python
# foo.py
x = 1
def f():
    return x
```
```python
import foo
foo.x = 2
f = myia(foo.f)
f()  # Resolving x in foo.py will give you 1 instead of 2
```
Even if you decide not to support this mutability, you would still have to deal with e.g. import side-effects when doing `from m import x`.

> Or Myia functions can have a global environment, which is quite a bit simpler than adding arguments all over the place. For example, every time a global variable is needed but its value is not relied upon, it may be associated to an integer index, then they can all be packed into an array for cheap lookup. Alternatively, put them in a closure.

To do type inference with this you need each index to have a type, so you have a typed mapping from index to value, which is an environment. If you want to support Python's `global` keyword, this environment would have to be mutable too.

Having environments was my second proposal, and I think it might be a reasonable option, since it keeps our options open:

1. To support Python globals we can have a single read-only environment as a global. Type inference needs to support this (i.e. we need a corresponding type environment as well).
2. If we eventually want to support closures with free variables with runtime name resolution, we can make the environment writable and pass them as arguments to closures. The same machinery can be used.

---

So in summary, the options are:

1. Remove Python globals entirely from the compiler pipeline by just inlining them.
2. Turn Python globals into regular arguments.
3. Turn Python globals into a global read-only environment.

I think the advantage of 1 and 2 is that they are entirely local to the Python parser/calling mechanism. The advantage of 3 is that it might be useful on the long term (e.g. if we want to fully support closures or have support for the `global` and/or `nonlocal` keywords), but it will require some extra work for type inference.

abergeron:
@bartvm I have been following the discussion for a while, but I'm not following some of your points.  Maybe I'm just dense but here is what I don't understand:

1) How are we coupling the IR to the python runtime by adding a new primitive? It seems entirely feasible to add a new primitive and keying it with strings is language-neutral.  We would need something like this if we introduce a runtime environment anyway.

2) How could a python name resolution pass could break after optimizations? This pass would be something that turns a bunch of `Apply` into `Constant` (or maybe `Parameter`).

3) Why do we care which weird ways the users come up with to mutate their globals?  We can resolve the names during or just before the call and sidestep all the weird python semantics.  Or we can just resolve the names at a well-defined point (such as a `bind_globals()` method) and ignore future changes.

To be perfectly honest I'm not sure that the primitive approach is the best way to go, but I fail to understand the arguments above as to why one of the other approach is better.

In any case I don't think we want to use the same approach for all globals.  We can resolve things such as functions very early and delay the resolving of other things like ints to later.  It is possible that the user would turn a function into an int or the reverse, but we can just not support those cases.

bartvm:
@abergeron I'll try to be a clearer.

> 1. How are we coupling the IR to the python runtime by adding a new primitive. [...]

Because Python's namespaces are mutable, performing name resolution requires access to the Python interpreter at the time that the function is being executed (see my previous example). That is, executing `Apply(primitives.Resolve(), 'x', ...)` requires `...` to be a _live_ reference to the Python module in which `x` was defined. Carrying around this reference to a module is what causes the coupling.

You cannot replace this reference to a live Python object with a static reference (like @breuleux's suggestion of using `Apply(primitives.Resolve(), 'x', 'm.py')`), because any modifications to the module or built-ins namespace wouldn't be reflected.

My suggestions were trying to come up with an approach by which the name resolution becomes the responsibility of the calling code (which can be Python-specific) and is removed from the IR e.g. by turning the globals into arguments or by introducing environments.

> 2. How could a python name resolution pass could break after optimizations? This pass would be something that turns a bunch of Apply into Constant (or maybe Parameter).

Like the examples I gave: What if we move the optimization to C++? Or what if the IR gets dumped to disk and then reloaded? In those cases the references to Python objects will be either hard or impossible to keep alive.

> 3. Why do we care which weird ways the users come up with to mutate their globals? We can resolve the names during or just before the call and sidestep all the weird python semantics.

This is option (1) from the three options I gave in my previous comment i.e. just before the call we simply resolve all the globals, and then we treat them as constants from thereon out. In https://github.com/mila-udem/myia/pull/4#issuecomment-357465041 I asked whether that wasn't enough, after which @breuleux replied in https://github.com/mila-udem/myia/pull/4#issuecomment-357733143 that it would be beneficial to have actual support for globals in the IR instead. The rest of the discussion has been about what this support would look like then.

> It is possible that the user would turn a function into an int or the reverse, but we can just not support those cases.

I am happy not supporting this, especially in the beginning, with the caveat that we should throw errors rather than do something different from Python.

breuleux:
> That is, executing `Apply(primitives.Resolve(), 'x', ...)` requires `...` to be a *live* reference to the Python module in which `x` was defined.

I don't believe that's true. That live reference can be reliably retrieved using `vars(__import__(module_name))`. That requires the Python runtime to *evaluate*, of course, but there's no need to put a pointer in the IR.

> Like the examples I gave: What if we move the optimization to C++? Or what if the IR gets dumped to disk and then reloaded? In those cases the references to Python objects will be either hard or impossible to keep alive.

Neither of these seem to be problems at all. The IR only needs to contain the module name (a string), and to "retrieve" a global variable just means to load the module in which it is defined using the method I just described. I believe that's what `pickle` does.

> In #4 (comment) I asked whether that wasn't enough, after which @breuleux replied in #4 (comment) that it would be beneficial to have actual support for globals in the IR instead.

That was only the second point in the reply. Did you comment on the first point I made, which is that if the values of some globals are unnecessary for optimization, it saves work to only resolve them at the end, because when they change we only need to backtrack up to that point?

abergeron:
@bartvm Ok so I think the main disagreement is that you think we need a reference to the actual python module object to be safe whereas I assumed a name reference is enough.

With the name reference, none of the concerns you raised apply, which is why I was puzzled.  So my next question, why do you need we need a reference to the actual python module object?  Why is just the name of the module not enough?

bartvm:
> [...] why do you need we need a reference to the actual python module object? Why is just the name of the module not enough?

If you don't perform the name resolution in the same Python interpreter, you will need to re-import the module, which can have side-effects (i.e. it might trigger compiling Myia functions, Theano might allocate memory on the GPU, etc.)

> That live reference can be reliably retrieved using `vars(__import__(module_name))`

This still requires you to keep the Python interpreter alive and run this command in that same interpreter, otherwise you might trigger side-effects. I'd rather have a valid IR where the calling code is responsible for the name resolution.

> Neither of these seem to be problems at all. The IR only needs to contain the module name (a string), and to "retrieve" a global variable just means to load the module in which it is defined using the method I just described.

As you say, executing this IR _requires_ you to import a Python module. Basically, the IR you just defined is useless outside of Python, which is the coupling I keep talking about. Lifting the name resolution out of the IR into the calling mechanism avoids that. Random example of things you can do then:
```python
# block the GPU, or whatever other side-effects you can think of
# this makes it undesirable to import this outside of the original interpreter
import theano

# compile function to a shared library that can be called from C++
x: int = 1
def f():
    return x
myia.create_shared_lib(f, 'myia.so')
```
```c++
typedef int (*f)(*env);
struct env_t {
    int x;
} env;
env.x = 2;
void *myia = dlopen('myia.so', RTLD_NOW);
f = dlsym(myia, 'f');
int x = (*f)(&env);
dlclose(myia);
```

> Did you comment on the first point I made, which is that if the values of some globals are unnecessary for optimization, it saves work to only resolve them at the end, because when they change we only need to backtrack up to that point?

Yes, I agree it saves some work, which is why I've been proposing ways of handling globals that doesn't rely on having Python references inside of the IR.

bartvm:
As a sidenote, please note that the change I'm talking about is really very minor in practice. You basically replace `Apply(primops.Resolve(), varname, python_module)` with something Python-agnostic like `Apply(primops.ReadEnv(), global_env, f"{python_module}.{varname}")`. The only point is that we don't hardcode the existence of Python modules into the IR, but instead use something more generic that can naturally be fed data by a non-Python caller.

breuleux:
> As a sidenote, please note that the change I'm talking about is really very minor in practice. You basically replace `Apply(primops.Resolve(), varname, python_module)` with something Python-agnostic like `Apply(primops.ReadEnv(), global_env, f"{python_module}.{varname}")`. The only point is that we don't hardcode the existence of Python modules into the IR, but instead use something more generic that can naturally be fed data by a non-Python caller.

I see what you mean, but what I proposed *does not* hardcode the existence of Python modules into the IR: `Apply(resolve, module, name)` only means "look up this key,value pair in whatever the global environment is". The implementation of this *can* require a Python module and look at its globals live, but it doesn't *have* to do this.

Think of it this way: `global_env` is going to be the same object for every Myia function and closure, so why is it the IR's responsibility to pass it around? The interpreter can take care of it, and offer an interface to change these values.

bartvm:
> I see what you mean, but what I proposed does not hardcode the existence of Python modules into the IR [...]

Sure, although we're now entering the domain of semantics. "Resolve variable 'x' in module 'm.py'" sounds Python-specific to me since it is not obvious what the meaning of a "module" is outside of Python. But sure, a C++ runtime could create an environment which is really a nested struct with keys `m.py` and `x` and look up the value in there.

Hence my last comment: The practical impact is relatively minor. What I'm trying to hammer on is (a) the IR should be Python-agnostic and (b) separation of concerns should be maintained by making sure that only the parser and calling mechanism are concerned with Python globals.

> Think of it this way: `global_env` is going to be the same object for every Myia function and closure, so why is it the IR's responsibility to pass it around? 

I'm not saying it should be passed around, it can just be a single constant that all functions are closed over. The advantage of having an environment object (instead of hardcoding it in the interpreter) is that eventually we _can_ turn it into something that is passed around and/or mutable in order to support runtime name resolution in closures (my example in https://github.com/mila-udem/myia/pull/4#issuecomment-357769792).

On a more practical level, the type of this environment object will be a type map which can be naturally used by the type inference pass (else you need to special case the `Resolve` primitive somehow in order to determine its output type).

breuleux:
Okay, so do you mean something like this?

```python
# main.py
from m import g
x = 1
def f():
    return g(x)
...
f()
```

==>

```python
def GLOBAL(env):
    def f():
        return env.main.g(env.main.x)
    return {'main.f': f}
...
GLOBAL({'main.x': x, 'main.g': g}).main.f()
```

Where, basically, there is one global `Graph` that takes the global environment as an argument and contains every function from every module. That'd be fine by me. So would be feeding the globals as individual arguments to `GLOBAL`, if that makes things any simpler.

abergeron:
@bartvm Your proposal of something like `Apply(ReadEnv(), env, 'm.x')` sounds much like `Apply(Resolve(), 'm', 'x')`.  I like combining the names into one string, so we should go with that.

The only difference is if you are explicit with the environment in the IR.  I don't think we should be.  It could differ from call to call, but that should be the job of the interpreter.  So something like `Apply(ReadEnv(), 'm.x')` has my vote.

bartvm:
@breuleux Yeah. The other alternative is to just have it as a global constant, but then the caller would be responsible for setting its value correctly each call, while this way that requirement is made explicit.

@abergeron I'm not sure I follow your argument. The environment can be different from call to call by just passing a different environment object each time. I actually do think it's important that it is explicit, since it maintains the IR's purity. Why break this my making `readenv` an impure function that needs to be special cased?

For example, consider type inference. Only arguments and constants normally have types. If you make the environment explicit, this is completely natural:
```python
# Python
x = 1
def f():
    return x
```
```python
# An environment is just a mapping from variable names to values
class Env(types.SimpleNamespace):
    x: int

# Bind globals to functions
def main(env: Env):
    def f() -> int:
        return readenv(env, 'x')
    return f
f = main(Env(x=1))

# Call the function
assert f() == 1
```
```python
class ReadEnv(Primitive):
    ...
    def type_(self, args):
        """Type inference for the ReadEnv primitive."""
        # An environment's type is a type map (i.e. from variable names to types)
        env_t = args[0].type
        varname = args[1].value
        return env_t[varname]
```

abergeron:
All right I get your point about being explicit and I agree.

breuleux:
Okay, so it's settled, let's go for explicit. We can have a global graph in which every other is nested. Should it take an `env` parameter that we read from with `readenv`, or should each global be appended to the parameter list as a separate parameter?

I'll note an additional interesting aspect of being explicit, which is that some checking code can be offloaded to Myia: an optimization that decides to specialize on the value of some global can inject code in the global closure that verifies that the passed parameter is indeed equal to the value we specialized on.

bartvm:
@breuleux Maybe as a first implementation we can have each global be a separate parameter since it avoids having to implement the `readenv` primitive.

I'll leave the implementation of all of this to a next PR though; for now we can just do the very naive "parse each time the function is called", just so that this can be merged and tested alongside with the VM from @abergeron (once he implemented support for closures).

With that in mind, I tried finishing up this PR so that there is at least something in `clean` to work with:

* Removed the graph debugging code and string representations now that we have #7 and #9 merged.
* Added unit tests that test whether the code runs; correctness isn't tested yet, but maybe that is best done with the VM anyway.
* The globals environment can now be manually passed to the parser. Moreover, instead of just considering the global (module) namespace through `__globals__`, it by default now also considers the enclosed namespace and built-ins namespace (using `inspect.getclosurevars`). This means that the lookup in the builtins namespace is no longer hardcoded.

Here's a rough to-do list for subsequent PRs:

- [ ] Support forward declaration by doing name resolution in a separate just-in-time pass (right now `parse` must be called at the end)
- [ ] Enforce the immutable declaration rule (i.e. assignment to a variable that was closed over can only happen once) because right now it can lead to "wrong" results
- [ ] Handle globals more efficiently. In order of comprehensiveness:
  - [ ] Only reparse when globals have changed
  - [ ] Make globals explicit arguments as parameters
  - [ ] Introduce environment objects
- [ ] Integration tests using the VM

breuleux:
> Maybe as a first implementation we can have each global be a separate parameter since it avoids having to implement the `readenv` primitive.

Sounds good to me.

> Enforce the immutable declaration rule (i.e. assignment to a variable that was closed over can only happen once) because right now it can lead to "wrong" results

I wouldn't put a very high priority on that one, I'd leave it for the polish phase for before we release a beta.

What remains for the parser's functionality? Quite a bit, I assume.

bartvm:
> What remains for the parser's functionality? Quite a bit, I assume.

I think the most important functionality is there (name resolution, nested definitions, calls, loops, conditionals). I'd rather keep things to the bare minimum for now so that we can get to an MVP (alongside with the VM and gradient logic). We can add support for more Python constructs afterwards.

So I'd say this PR is done for now, unless there's something essential I'm missing. We can merge it into `clean`, or alternatively we can wait until `abergeron:vm` is done and merge that into `parser` so that we can add integration tests before merging into `clean`.

breuleux:
> I think the most important functionality is there (name resolution, nested definitions, calls, loops, conditionals).

I don't think nested definitions work:

```python
def f(x):
    def g():
        return x
    return g()
```

`parse(f).return_` gives me `None` in this case.

breuleux:
Here are other graphs that seem incorrect:

```python
def f(x):
    if x:
        return 1
    else:
        return 2
```

The continuation branch of the `if` is a graph with `return_ == None`. Most likely related to #13, but I really think that ought to be fixed.

```python
def g(x):
    return x

def f(x):
    return g(x + 1)  # works if g(x)
```

The above fails with `AttributeError: 'BinOp' object has no attribute 'id'`. It looks like the code that handles `ast.Call` assumes all arguments are variable names.

```python
def f(x):
    return g(x)

def g(x):
    return f(x)
```

Mutual recursion does not work, it leads to a stack overflow.

bartvm:
> The above fails with AttributeError: 'BinOp' object has no attribute 'id'. It looks like the code that handles ast.Call assumes all arguments are variable names.

I am assuming the input is ANF right now. I'm not sure whether it should be the parser's responsibility to handle code with non-trivial arguments. It might be better if have a separate pass that turns the code into something more similar to ANF (i.e. where all arguments are terminating expressions, this is what Tangent does, so we can just take the code from there).

> The continuation branch of the if is a graph with return_ == None. Most likely related to #13, but I really think that ought to be fixed.

Yup, I think that's the same issue. Will have a look when I have a chance.

> Mutual recursion does not work, it leads to a stack overflow.

Will make an issue for that one too.

breuleux:
> I am assuming the input is ANF right now. I'm not sure whether it should be the parser's responsibility to handle code with non-trivial arguments. It might be better if have a separate pass that turns the code into something more similar to ANF (i.e. where all arguments are terminating expressions, this is what Tangent does, so we can just take the code from there).

I don't think there is any need for this assumption at all. Just call `process_expression` recursively like you do for binops. The parser currently processes non-ANF expressions such as `x + y + z` or `g(x) + g(x)` anyway.

breuleux:
Regarding merging, I think we should probably merge this into `clean` regardless. The parser is in rough but reasonably working order and there's some critical components in this branch like the `Return` primitive that I'd like to merge sooner than later.

abergeron:
Regarding essentials, when try to parse this:
```
def test_fn1(x):
    def g(x):
	return x
    y = g(x)
    return y
```

I get this error:
```
self = <myia.parser.Parser object at 0x103c43780>
block = <myia.parser.Block object at 0x103c43f60>
node = <_ast.Call object at 0x103c430b8>

    def process_expression(self, block: 'Block', node: ast.expr) -> ANFNode:
        """Process an expression."""
        expr: ANFNode
        if isinstance(node, ast.BinOp):
            inputs_: List[ANFNode] = [self.environment.ast_map[type(node.op)],
                                      self.process_expression(block,
                                                              node.left),
                                      self.process_expression(block,
                                                              node.right)]
            expr = Apply(inputs_, block.graph)
            expr.debug.ast = node
        elif isinstance(node, ast.Name):
            expr = block.read(node.id)
        elif isinstance(node, ast.Num):
            expr = self.environment.map(node.n)
>       return expr
E       UnboundLocalError: local variable 'expr' referenced before assignment
```

Maybe it would be nice if the parser for expressions handled things like calls to other functions.  It can be later, but I would need that to properly test the VM.

breuleux:
@abergeron Support for `ast.Call` was added earlier, you'll need to merge these commits to your branch.

bartvm:
Okay, have to do some cleaning and testing, but I think a bunch of things are fixed:

* Multiple return statements
* Missing else branches
* Non-atomic arguments

Most of them required pretty minor changes. Still need to look at mutual recursion.

abergeron:
Yeah, I just rebased my branch and calls work.  I'll do some more tests.

abergeron:
Spoke too soon, I get a graph whose `return_` attribute is `None` with that same test.  I've pushed it to my PR so that you can try to look at that and figure out why.

bartvm:
@abergeron Try 840786c52d805a3c40a695a931fb7c2dfe051944 to see if it helps (the problem was that when there is a nested function definition, we have to go back to the original basic block it was defined in after processing the definition, but it was actually continuing with the last basic block of the nested function).

abergeron:
@bartvm that works.

breuleux:
I think there are issues with `while` loops:

```python
def f(x):
    ret = 0
    while x:
        ret = ret + x
        x = x + 1
    return ret
```

`f(-3)` returns `0` using @abergeron's vm, but it should return `-6`. I think this is a problem with the parser, though, because this is the graph I get:

while

The function `#1` only has two parameters (`#2` and `#3`, in green, which are both `x`) but it is called with four parameters. (Pink nodes are constants, and violet nodes are free variables: I now draw duplicate nodes so that graphs don't look like arrow spaghetti.) I get the feeling that the variables accessed by `while` are basically appended to the parameter list twice.

breuleux:
Quick thought about closures: right now, the parser generates a `Constant(graph)` when it encounters a nested function, which like all `Constant` objects has its graph is set to `None`. That doesn't seem right: that node represents a closure over the variables of its parent graph, so it *isn't* a constant. It's not necessarily a problem to use the `Constant` class, but I think the `graph` field should be set to the parent to signify that it is to be constructed in that context.

It's kind of a detail, just wanted to raise that point.

bartvm:
Non-closures also have corresponding `Constant` nodes. The `Constant` doesn't represent a closure but a function. All functions are top-level, and a closure is implicit by a function accessing another function's scope.

It is also not always true that a nested function in Python is still a closure when parsed (or after optimizations), so you would have to take care to always adjust `Constant.graph` attributes appropriately.

```python
def f():
    x = 2
    def g():
        return x
    return g

# Parsed to
def f():
    return g

def g():
    return 2
```

breuleux:
Yeah, I realized that while trying to figure out the nesting structure to list the free variables for `grad`.

myia/parser.py
```diff
@@ -313,9 +313,6 @@ def process_while(self, block: 'Block', node: ast.While) -> 'Block':
         header_block = Block(self)
         body_block = Block(self)
         after_block = Block(self)
-        header_block.preds.extend([block, body_block])
```
bartvm: @breuleux This is were things were added twice, the graph looks okay now to me, haven't tested it with the VM.

bartvm:
Fixed the recursive calls (just needed to move one line around). I think that fixes most of the things that @breuleux brought up. I'll wait for @abergeron to finish the new VM, easier to do testing and cleaning up once that's up.

breuleux:
Indeed, as far as I can tell all the issues I brought up are fixed. It's still not possible to have mutual recursion between nested functions, e.g.:

```python
def f():
    def g(): return h()
    def h(): return g()
    return g()
```

But I think that's a minor issue that we can ignore for now.

breuleux:
Small issue I just discovered writing some tests: it's not possible to put docstrings in Myia functions, because they produce `ast.Expr` nodes and these are not handled.

bartvm:
> Indeed, as far as I can tell all the issues I brought up are fixed. It's still not possible to have mutual recursion between nested functions, e.g.:

Yup, that relies on forward references (#10) and requires a few bigger changes (names can't be resolved immediately during parsing anymore).

> Small issue I just discovered writing some tests: it's not possible to put docstrings in Myia functions, because they produce ast.Expr nodes and these are not handled.

I guess we can ignore all expression statements for now, until we start implementing functions with side-effects (like `print`).

bartvm:
I'll update #5 by the way, so that other missing features can be detected by a syntax checking pass in the future.

bartvm:
I will squash and merge this in a bit if there are no further comments.

* Removed the graphing functions and `find_graphs` utility.
* I made `dfs` specific to the `ANFNode` class (necessary because of the `follow_graph` option), and renamed things appropriately (`ir_utils` is now `anf_ir_utils`).
* Added a test for `follow_graph` option
* I had to use one `type: ignore` because Mypy doesn't understand that `FunctionType` is a type of `Callable` (but I don't want to use the latter because we can only handle user-defined functions, not callables in general).

bartvm:
Quick note related to earlier discussions: This [blog post](http://hokstad.com/how-to-implement-closures) is a short read that is interesting in the context of our current closure support (spaghetti stack) and the environment object approach we discussed earlier (referred to as "rewriting").

Something else to consider: The `grad` operator might more naturally interact with the rewriting approach, since each backpropagator will have to return the derivative with respect to its environment. Having the environment be an explicit argument in the first place makes for a nice symmetry that might be easier to work with.

breuleux:
No objection to squash and merge.

PR #5, opened 2017-12-16
Syntax checker

bartvm:
Going through `parse.py` in `master` I noticed that the parser is responsible for checking whether we support the given AST, and if not, raising an informative syntax error. In Tangent we actually do this in a separate pass, which I think might be a good idea. It keeps the parser a bit cleaner (no need for checks) and separates two logical processes.

I also found a comment that says "maybe inherit from SyntaxError? Investigate." so I had a look, and it is a good idea. If you set the fields appropriately, it will automatically pretty print the error message with the filename, a little arrow, etc.

```pytb
bartvm-macbookpro3:myia bartvm$ python myia/fence.py
Traceback (most recent call last):
  File "myia/fence.py", line 120, in <module>
    fence.visit(ast.parse(textwrap.dedent(inspect.getsource(f))))
  File "myia/fence.py", line 88, in visit
    self.generic_visit(node)
  File "/Users/bartvm/anaconda3/lib/python3.6/ast.py", line 261, in generic_visit
    self.visit(item)
  File "myia/fence.py", line 88, in visit
    self.generic_visit(node)
  File "/Users/bartvm/anaconda3/lib/python3.6/ast.py", line 261, in generic_visit
    self.visit(item)
  File "myia/fence.py", line 93, in visit
    return visitor(node)
  File "myia/fence.py", line 98, in visit_Pass
    self.raise_("you shall not pass")
  File "myia/fence.py", line 109, in raise_
    self.lines[self.lineno - 1], message)
  File "myia/fence.py", line 113
    pass
    ^
__main__.MyiaSyntaxError: invalid syntax: you shall not pass
```

I attached a minimal proof of concept of this approach. The syntax checker is conservative i.e. it rejects everything unless it's been explicitly marked as supported.

### DIFF
--- a/myia/fence.py
+++ b/myia/fence.py
@@ -1,0 +1,119 @@
+"""Check Python code for compatibility with Myia."""
+import ast
+from typing import Sequence
+
+
+class MyiaSyntaxError(SyntaxError):
+    """Syntax error for Python code unsupported by Myia.
+
+    This class inherits from `SyntaxError` and pretty prints the relevant line
+    and code when raised.
+
+    """
+
+    def __init__(self, filename: str, lineno: int, offset: int, text: str,
+                 message: str = None) -> None:
+        """Construct a Myia syntax error.
+
+        Args:
+            filename: The filename from which the code was loaded. Use a string
+                like `<string>` for code that was not loaded from a file.
+            lineno: The 1-indexed line in the file or string with the offending
+                expressiong or statement.
+            offset: The column which should be pointed to in the error message.
+            text: The text to print, which should normally be the line in the
+                file given bye `lineno` and `filename`.
+            message: An optional message to clarify what is invalid about this
+                syntax.
+
+        """
+        full_message = "invalid syntax"
+        if message:
+            full_message = f"{full_message}: {message}"
+        super().__init__(full_message)
+        self.filename = filename
+        self.lineno = lineno
+        self.offset = offset
+        self.text = text
+
+
+class Fence(ast.NodeVisitor):
+    """Check if each Python AST node is supported.
+
+    This node visitor walks the Python AST, and raises a syntax error for each
+    AST node that is not explicitly supported.
+
+    If a node is fully supported, it can be added to the `supported` attribute.
+    If a node is partially supported, implement a `visit_Node` method which
+    calls the `raise` method when certain conditions are not met.
+
+    Attributes:
+        supported: A container of nodes that are fully supported.
+
+    """
+
+    supported = (ast.Module, ast.FunctionDef, ast.arguments, ast.arg)
+
+    def __init__(self, filename: str, lines: Sequence[str],
+                 line_offset: int) -> None:
+        """Construct a fence.
+
+        Args:
+            filename: The name of the file from which the AST was loaded.
+            lines: A sequence of lines whose indices correspond to the line
+                numbers in the AST.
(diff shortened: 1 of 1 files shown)


breuleux:
The advantage of having the checks in the parser is that they serve as scaffolding for when you actually add the missing features: if you want to support something new you can replace the error by the implementation, whereas if they are different passes you have to change code in two places. I don't feel particularly strongly about doing it one way or another, but it's something to think about.

Regarding extending `SyntaxError`, is there any way to remove all traceback entries except the last one? I recall spending a few infuriating hours trying to do this before giving up and replacing the exception hook, but I might have missed something obvious.

bartvm:
Not that I know of, but I can have a look. Then again, do we really want to do that? Seeing what triggered the syntax check might be useful. Perhaps largely theoretical, but there could be situations where Myia is invoked dynamically (think `compiled_f = myia(some_functions[randint(10)])` or `if compile_with_myia: compiled_f = myia(f)`). In those cases you might want to use `pdb` to walk up the call tree, and it would be good if that corresponded to the actual traceback you get.

breuleux:
That's true, the traceback could be useful in such cases. One way to reduce the noise from all the recursive calls to `visit` (there could be a whole lot of those) would be to catch the error downstack and re-raise it. That'd only leave the syntax annotation and the part of the stack the user might actually care about.

bartvm:
That might be a good idea. It looks like you need to copy the exception in order for exception chaining to be triggered, but `except MyiaSyntaxError as e: raise copy.copy(e)` does the trick.
```pytb
Traceback (most recent call last):
  File "myia/fence.py", line 127, in g
    fence.visit(ast.parse(textwrap.dedent(inspect.getsource(f))))
  File "myia/fence.py", line 93, in visit
    self.generic_visit(node)
  File "/Users/bartvm/anaconda3/lib/python3.6/ast.py", line 261, in generic_visit
    self.visit(item)
  File "myia/fence.py", line 93, in visit
    self.generic_visit(node)
  File "/Users/bartvm/anaconda3/lib/python3.6/ast.py", line 261, in generic_visit
    self.visit(item)
  File "myia/fence.py", line 98, in visit
    return visitor(node)
  File "myia/fence.py", line 103, in visit_Pass
    self.raise_("you shall not pass")
  File "myia/fence.py", line 114, in raise_
    self.lines[self.lineno - 1], message)
  File "myia/fence.py", line 120
    pass
    ^
__main__.MyiaSyntaxError: invalid syntax: you shall not pass

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "myia/fence.py", line 130, in <module>
    g()
  File "myia/fence.py", line 129, in g
    raise copy.copy(e) from e
  File "myia/fence.py", line 120
    pass
    ^
__main__.MyiaSyntaxError: invalid syntax: you shall not pass
```
It does make the traceback from the `visit` calls inaccessible in `pdb`, so we'll have to trust that a user or developer won't want to actually inspect the AST that triggered the syntax error. It's also a bit confusing that the error appears twice, but there is a lot less noise and `up` commands to type to get to the user's code, which is good.

breuleux:
I doubt anyone will want to inspect the AST instead of simply looking at the source, but we can just pass the node to `MyiaSyntaxError` and store it in `e.node` or something.

breuleux:
This is over a year old back from the beginning of the repo, I'm just going to close this.

PR #6, opened 2017-12-19
Pipeline

breuleux:
Myia will need to process functions along a pipeline, following (more or less) the following steps (each one using the result of the previous):

* Base level: A basic Python function.
* Parsing: Produces IR0.
* Inference: Checks types and produces IR1.
* Optimization: Optimizes the graph and produces IR1.
* VM: Lowers the IR and converts data to the VM's format.
* User: High level API for the user.

Primarily for testing purposes, it should be possible to build and run more than one pipeline in parallel on the same functions while sharing relevant parts, e.g. the parsing. So there should be a caching mechanism.

This PR contains code to implement a pipeline generator that satisfies these requirements. A `Universe` is a step in the pipeline that basically takes a representation one level down, transforms it, and (for certain types, like functions) caches the result:

```python
# Define names and constructors for the possible steps in the pipeline
pgen = PipelineGenerator(py=PythonUniverse, parse=ParseUniverse, ...)

# Build a pipeline -- the order of the steps can vary, some steps may be omitted
u = pgen('py->parse->inf->opt->vm->ui',
         opt={'passes': [...]},
         vm={'debugger': True}, ...)

# Index the resulting universe to get a compiled function
myia_f = u[f]
myia_f(x, y, z)
```

The base level is implemented as `PythonUniverse`, which:

* Transforms a myia-transformed function back into the original function (because that function may have been made by a different pipeline, and we may need to recompile it).
* Resolves `Symbol` instances that represent global variables (the `Symbol` class could be simplified a bit).

### DIFF
--- a/myia/pipeline.py
+++ b/myia/pipeline.py
@@ -1,0 +1,242 @@
+
+import inspect
+from types import FunctionType
+from typing import Tuple, Type, Any, Dict, Union
+
+
+class Symbol:
+    """
+    Represent a variable name in Myia's frontend AST.
+
+    Symbols should not be created directly. They should be created
+    through a GenSym factory: GenSym enforces a unique namespace and
+    keeps track of versions to guarantee that no Symbols accidentally
+    collide.
+
+    Attributes:
+        label (str or Symbol): the name of the variable. If
+            relation is None, this must be a string, otherwise
+            this must be a Symbol.
+        namespace (str): the namespace in which the variable
+            lives. This is usually 'global', 'builtin', or a
+            uuid created on a per-LambdaNode expression basis.
+        version (int): differentiates variables with the same
+            name and namespace. This can happen when there are
+            multiple writes to the same variable in Python.
+        relation (str): how this variable relates to some other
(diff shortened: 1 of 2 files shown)


bartvm:
Before reviewing, I have two questions/remarks.

Firstly, I have difficulty understanding the roles of these abstractions. For example, the `Symbol` class: What Myia AST? Which namespaces? What role does this play in AD? Is its functionality shared across steps in the pipeline? If so, why? And what is a `Universe` for? Why does every step need one? Why does this caching matter?

Secondly, without any of the steps of the pipeline in place, I have difficulty understanding the need for these abstractions. Why can't the pipeline just be a wrapper about something that effectively does `myia_f = compile(optimize(parse(fence(f)))`? Why do we need a `Universe` abstraction each step of the way, instead of just letting each two connecting steps in the pipeline decide on their abstraction?

In general, I think it might be better to wait with merging something as high level as a pipeline manager until we actually have the pieces of the pipeline in place. That said, in order for me to review the actual code I would first need a much clearer understanding of its purpose.

PR #7, opened 2018-01-13
Automatic node naming and string representations

breuleux:
* Add superclass Debug to GraphDebug and NodeDebug
* `Debug.force_name` names a node if it has no name
* `__str__` returns `node.debug.force_name()`
* `__repr__` returns a shallow representation

The basic idea is that names are lazily generated for nodes if they are needed: if a node has no `debug.name`, calling `str(node)` will generate one. It's a bit unusual for `str` to have a side effect, but it makes sense in this context, since it's only important for a node to have a name if a human is going to read it. `__str__` always just returns the name, whereas `__repr__` is more informative, but still shallow (it calls `str` on the inputs).

### DIFF
--- a/myia/anf_ir.py
+++ b/myia/anf_ir.py
@@ -27,22 +27,24 @@ class Debug(types.SimpleNamespace):
 
     Attributes:
         name: The name of the object.
+
     """
-    __curr_id__ = 0
 
-    def __init__(self, **kw):
-        self.name = None
-        super().__init__(**kw)
+    _curr_id = 0
 
-    def force_name(self):
+    def __init__(self, **kwargs):
+        """Construct a Debug object."""
+        self.name: str = None
+        super().__init__(**kwargs)
+
+    @property
+    def debug_name(self):
         """Return the name, create a fresh name if needed."""
         if self.name:
             return self.name
-        Debug.__curr_id__ += 1
-        self.name = f'#{Debug.__curr_id__}'
+        Debug._curr_id += 1
+        self.name = f'#{Debug._curr_id}'
         return self.name
-
-    name: str
 
 
 class Graph:
@@ -67,21 +69,24 @@ class Graph:
         self.debug = GraphDebug()
 
     def __str__(self) -> str:
-        return self.debug.force_name()
+        """Return string representation."""
+        pfx = f'{self.debug.name}=' if self.debug.name else ''
+        return f'{pfx}Graph(parameters={self.parameters})'
 
     def __repr__(self) -> str:
-        pfx = f'{self.debug.name}=' if self.debug.name else ''
-        ret = self.return_ and self.return_.inputs[1]
-        return f'{pfx}Graph(parameters={self.parameters}, return_={ret!r})'
+        """Return representation."""
+        return str(self)
 
     @classmethod
     def __hrepr_resources__(cls, H):
+        """Require the cytoscape plugin for buche."""
         return H.bucheRequire(name='cytoscape',
                               channels='cytoscape',
                               components='cytoscape-graph')
 
     def __hrepr__(self, H, hrepr):
-        from .debug.gprint import GraphPrinter, css
+        """Return HTML representation (uses buche-cytoscape)."""
+        from .debug.gprint import GraphPrinter, css  # type: ignore
         rval = H.cytoscapeGraph(H.style(css))
 
         options = {
@@ -188,9 +193,12 @@ class ANFNode(Node):
         return obj
 
     def __str__(self) -> str:
-        return self.debug.force_name()
+        """Return string representation."""
+        return self.debug.debug_name
 
-    __repr__ = __str__
+    def __repr__(self) -> str:
+        """Return representation."""
+        return str(self)
 
 
 class NodeDebug(Debug):
@@ -327,6 +335,7 @@ class Apply(ANFNode):
         super().__init__(inputs, APPLY, graph)
 
     def __repr__(self) -> str:
+        """Return representation."""
         pfx = f'{self.debug.name}=' if self.debug.name else ''
         return f'{pfx}Apply({[str(x) for x in self.inputs]})'
 
@@ -364,7 +373,12 @@ class Constant(ANFNode):
         super().__init__([], value, None)
 
     def __str__(self) -> str:
-        return str(self.value)
+        """Return string representation."""
+        if self.debug.name:
+            return self.debug.name
+        else:
+            return f'Constant({self.value!r})'
 
     def __repr__(self) -> str:
-        return repr(self.value)
+        """Return representation."""
+        return str(self)
--- a/myia/anf_ir.py
+++ b/myia/anf_ir.py
@@ -16,6 +16,7 @@ from typing import (List, Set, Tuple, Any, Sequence, MutableSequence,
 
 from myia.ir import Node
 from myia.utils import Named
+import json
 
 PARAMETER = Named('PARAMETER')
 APPLY = Named('APPLY')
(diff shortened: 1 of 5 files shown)


breuleux:
I pushed a graphical printer for `Graph`. @bartvm, @abergeron, @lamblin (and anyone willing), can you check if this works for you?

Install: https://github.com/breuleux/buche/releases/tag/v0.1.4

Run:

```bash
git checkout node_naming
curl -o test_gprint.py https://gist.githubusercontent.com/breuleux/09079a20805cd5146dbd0fdf07954f8b/raw/324c94951423d687ff54f5e37cd40ccd48cec01d/test_gprint.py
buche python test_gprint.py
```

It should ask you if you want to install `buche-cytoscape`, which you should accept, and then it should display a few graphs. If you see an error, try closing `buche` and try again.

myia/anf_ir.py
```diff
@@ -16,11 +16,35 @@
...
 APPLY = Named('APPLY')
 
 
+class Debug(types.SimpleNamespace):
+    """Debug information for an object.
+
+    Attributes:
+        name: The name of the object.
+    """
+    __curr_id__ = 0
```
bartvm: PEP8 on leading and trailing underscores: "Never invent such names; only use them as documented."

bartvm: Also, do we really need name mangling here?

myia/anf_ir.py
```diff
@@ -42,8 +66,47 @@ def __init__(self) -> None:
         self.return_: Apply = None
         self.debug = GraphDebug()
 
+    def __str__(self) -> str:
+        return self.debug.force_name()
```
bartvm: `#1` is not very readable, we should probably have something like `Graph(id=1)` for str, and maybe something `<Graph(name=1, parameters=[...], return_=... at ...>` for `__repr__`.

myia/anf_ir.py
```diff
@@ -292,3 +362,9 @@ class Constant(ANFNode):
     def __init__(self, value: Any) -> None:
         """Construct a literal."""
         super().__init__([], value, None)
+
+    def __str__(self) -> str:
+        return str(self.value)
+
+    def __repr__(self) -> str:
+        return repr(self.value)
```
bartvm: `<Parameter(value={self.value}) at ...>` or something would be better, else you can't distinguish between `repr(p)` and `repr(p.value)` when debugging.

myia/anf_ir.py
breuleux: If we want unique names for nodes that have no name, yes.

myia/anf_ir.py
```diff
@@ -16,11 +16,35 @@
...
+    Attributes:
+        name: The name of the object.
+    """
+    __curr_id__ = 0
+
+    def __init__(self, **kw):
+        self.name = None
+        super().__init__(**kw)
+
+    def force_name(self):
```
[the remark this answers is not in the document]
breuleux: That's what I had at first, but it means we can't check whether a node does have a name or not. `GraphPrinter` needs to know to avoid unnecessarily labelling nodes. Maybe it could be some `has_name()` method.

myia/anf_ir.py
breuleux: `Graph(name=1)` suggests that `graph.name == 1`, but it's `graph.debug.name == 1`. Other than that, I suppose a bit more verbosity would be fine.

myia/anf_ir.py
bartvm: Actually, `__curr_id__` doesn't get mangled. `__curr_id` would get mangled, but I don't see how that has anything to do with guaranteeing uniqueness. It simply renames the attribute from `__curr_id` to `_Debug__curr_id`.

myia/anf_ir.py
breuleux: Good point.

myia/anf_ir.py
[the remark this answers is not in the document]
bartvm: Why would you want to avoid it being triggered?

myia/anf_ir.py
bartvm: You can also make it `func 1` or `def 1: x, y -> z`, etc. My point was mostly that the string representation should make clear somehow that this is a function/graph.

myia/anf_ir.py
breuleux: Oh, that's what you meant. Yeah I don't particularly care whether it's mangled or not.

myia/anf_ir.py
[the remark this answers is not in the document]
breuleux: The graphical representation will label the node for `a = x + y` as `add->a`, to signify it's an addition and the result is called `a`. Labelling some anonymous addition `add->#1`, on the other hand, is unnecessary, since that temporary name means nothing for the user and we can draw arrows to signify uses.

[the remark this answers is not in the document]
bartvm: So you're trying to distinguish between "real" variable names and generated variable names. It sounds to me like that should then just be made explicit i.e. we can have `n.name` which is `a` or `None` and `n.debug_name` which returns `n.name if n.name else generated_name`.

Then, if you *need* a string representation (e.g. in `pdb` or `__repr__`) you use `n.debug_name`, whereas if you only want a name if it corresponds to some semantic variable name (in your visualizations) then you use `n.name`.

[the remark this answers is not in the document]
breuleux: I mean, it's more subtle than that: *if* the user ever requested to know the name of some node, and `#1` was generated, then when printing out the graphical representation, the corresponding node *should* be labelled `#1`, because the fact it was previously requested means that it is now meaningful. This is starting to feel like bikeshedding though.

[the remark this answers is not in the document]
bartvm: You can just change `def force_name` into `@property def debug_name`.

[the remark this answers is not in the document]
breuleux: I kind of prefer `force_name()` because I feel it's a bit more explicit about what it does, but sure.

bartvm:
I made these comments alongside with my others, but maybe I forgot to press submit because I don't see them anywhere... Sorry about that.

Can we move `__hrepr_resources__` and `__hrepr__` to a separate file? I think it is better to keep debugging code and core code separate. For example, this allows us to separate dependencies (`json` isn't required for our IR), and to have separate code coverage, unit tests, and style guide requirements (i.e. I would insist on 100% coverage for core code, but not necessarily for debugging libraries).

breuleux:
I'll merge this into `clean` once tests pass unless there are objections.

myia/anf_ir.py
```diff
@@ -21,6 +21,31 @@
...
+        self.name: str = None
+        super().__init__(**kwargs)
+
+    @property
+    def debug_name(self):
+        """Return the name, create a fresh name if needed."""
+        if self.name:
+            return self.name
+        Debug._curr_id += 1
+        self.name = f'#{Debug._curr_id}'
```
lamblin: Too late, but `#1` in error messages would interfere with GitHub issues if they ever get reported by users.

bartvm: I actually changed it in my local branch because #1 is also not a valid variable name, which gets a bit awkward in some object representations. We can change it.

breuleux: It shouldn't be a valid variable name: this lets us have terse names that nonetheless can never clash with the user's choices. Could be `$1`, though.

bartvm: Don't we only generate names if they don't already exist? So how could it clash with a user's choice?

Why should it not be a valid variable name? For example, the `__repr__` of a node should probably use the `__str__` representation of its inputs (otherwise you end up printing the entire graph recursively). You then end up with things like `Apply(inputs=[#1, #2])`, which I think looks weird.

breuleux: I mean it could clash with the names of other nodes. If e.g. we generate the name `arg0`, the user could have used that name elsewhere.

> You then end up with things like `Apply(inputs=[#1, #2])`, which I think looks weird.

Personally I think it looks weird in a good way, it tells me that these are automatically named temporaries better than valid variable names would. The "best" alternative would be `_1`, `_2`, etc. but I think that's just ugly.

bartvm: Let's continue discussing in #17.

PR #8, opened 2018-01-15
vm

abergeron:
In the spirit of early PRs, here is my code for the VM.

What I'm trying to do here is to execute the graph directly to make easier to convert into a debugger.  I'm still handling tail calls correctly so that we don't need an infinite amount of memory.

This is currently very inefficient for temporary storage of the computed values (as in all values are kept until the function returns, with the exception that tail calls are not considered returns).  There can also be situations where we will recompute values more than once when they are used across internal functions.  The results should always be good though.

### DIFF
--- a/myia/vm.py
+++ b/myia/vm.py
@@ -4,7 +4,7 @@ This VM will directly execute a graph so it should be suitable for
 testing or debugging.  Don't expect stellar performance from this
 implementation.
 """
-from typing import List, Dict, Any, Union
+from typing import List, Dict, Any, Union, Set
 
 from myia.anf_ir import Graph, ANFNode, Constant, Parameter, Apply
 from myia.ir_utils import dfs
@@ -141,9 +141,24 @@ class VMFrame:
     def wrap_closure(self, value):
         """Wrap graphs that are closures."""
         if isinstance(value, Graph):
-            if any(n.graph and n.graph is not value
-                   for n in dfs(value.return_)):
-                return self.Closure(value, self)
+            subgraphs: Set[Graph] = set([value])
+            seen: Set[Graph] = set()
+            targets: Set[Graph] = set()
+            while len(subgraphs) != 0:
+                g = subgraphs.pop()
+                seen.add(g)
+                for n in dfs(g.return_):
+                    if n.graph and n.graph is not self and n.graph not in seen:
+                        targets.add(n.graph)
+                    if (isinstance(n, Constant) and
+                            isinstance(n.value, Graph)):
+                        subgraphs.add(n.value)
+            if len(targets) > 0:
+                frame = self
+                while frame and frame.graph not in targets:
+                    frame = frame.parent
+                    assert frame is not None
+                return self.Closure(value, frame)
         return value
 
     def do_call(self, graph: Union[Graph, 'VMFrame.Closure'], args: List[Any]):
--- a/myia/vm.py
+++ b/myia/vm.py
@@ -57,6 +57,7 @@ class VMFrame:
 
         def __init__(self, value: 'VMFrame' = None) -> None:
             """Which value to return."""
+            super().__init__()
             self.value: VMFrame = value
 
     class Closure:
@@ -85,7 +86,14 @@ class VMFrame:
         self.todo: List[ANFNode] = [self.graph.return_]
 
     def run(self) -> 'VMFrame':
-        """Run the graph to completion and return the result."""
+        """Run the graph to completion and return the result.
+
+        Returns:
+            The next frame to execute. Can be either the frame of a (tail)
+            call, the frame of the parent (if the current frame completed), or
+            `None` (if the program terminated).
+
+        """
         while True:
             try:
                 res = self.advance()
@@ -101,12 +109,12 @@ class VMFrame:
                 return res
 
     def tail(self) -> bool:
-        """Are we in a tail call?."""
+        """Determine whether the current call is a tail call."""
         return self.todo[-2] is self.graph.return_
 
     def done(self) -> bool:
-        """Are we done?."""
-        return len(self.todo) == 0
+        """Determine whether this function has fully evaluated."""
+        return not self.todo
 
     def result(self):
         """Get this frame's result."""
@@ -115,13 +123,25 @@ class VMFrame:
     def get_value(self, node: ANFNode):
         """Get the value for a node.
 
-        This will check if we've already computed a value for it and
-        add it to the queue otherwise. If there was no value, we stop
-        the evaluation of the current node and restart the step, since
-        no useful work was done.
+        If the value for a node is not available, the node will be added to the
+        todo list. Evaluation of the current node in `eval_node` will be halted
+        (by raising `Continue`) and control will be returned to the `advance`
+        method in order to evaluate the last item on the todo list.
 
-        If the value we want is in another graph, we will resume
-        execution in a different frame and jump back to this one later.
+        If the value is not available and the node belongs another graph, we
+        will evaluate that node in a different frame (by raising `Return`). A
+        `Jump` instruction will be added to that frame to jump back to the
+        current one.
+
+        Returns:
+            The value of the node.
+
+        Raises:
+            Continue: If the node's value is not available and its calculation
+                has been added to the todo list.
+            Return: If the node's value is not available and must be retreived
+                from an enclosing frame, or from a closure.
+
         """
         frame = self
         while frame and node.graph and node.graph is not frame.graph:
@@ -135,7 +155,7 @@ class VMFrame:
             if frame is not self:
                 raise self.Return(frame)
             else:
-                raise self.Continue()
+                raise self.Continue
         return frame.values[node]
 
     def wrap_closure(self, value):
@@ -166,6 +186,10 @@ class VMFrame:
 
         This will handle tail calls and closures.
 
+        Raises:
+            Return: Giving control to the new frame in which the function
+                (graph) will be evaluated.
+
         """
         if self.tail() and self.parent is not None:
             parent = self.parent
@@ -179,16 +203,25 @@ class VMFrame:
         raise self.Return(new_frame)
 
     def advance(self) -> 'VMFrame':
-        """
-        Take a step.
+        """Take a step.
 
-        This will perform one "step" of execution.  This corresponds
-        roughly to evaluating one node of the graph.  In the case of
-        function calls, we single-step through the called function
-        even though the apply is a single node in our graph.
+        This will perform one "step" of execution.  This corresponds roughly to
+        evaluating one node of the graph. In the case of function calls, we
+        single-step through the called function even though the apply is a
+        single node in our graph.
+
+        Returns:
+            The next frame to execute. Can be either a new function call, a
+            tail call, or a parent frame in the case of closures/accessing the
+            enclosed scope. Returns `None` when the program terminated.
+
+        Raises:
+            StopIteration: If there are no more steps to take and control
+                should be returned to the `run` function.
+
         """
         if self.done():
-            raise StopIteration()
+            raise StopIteration
 
         while True:
             node = self.todo[-1]
@@ -205,11 +238,16 @@ class VMFrame:
             return None
 
     def eval_node(self, node: ANFNode):
-        """
-        Compute the value for a node.
+        """Compute the value for a node.
 
-        This may fail and ask for a missing value to be computed or
-        enter into a call.
+        This may fail and ask for a missing value to be computed or enter into
+        a call.
+
+        Raises:
+            Return: If the node is a jump instruction (see `get_value`),
+                returning control to another frame.
+            ValueError: If the node is the application of an unknown primitive.
+
         """
         if isinstance(node, Constant):
             self.values[node] = self.wrap_closure(node.value)
@@ -217,6 +255,9 @@ class VMFrame:
             idx = self.graph.parameters.index(node)
             self.values[node] = self.args[idx]
         elif isinstance(node, self.Jump):
+            # Normally `advance` pops the last node after `eval_node`
+            # completes, but here we do it ourselves because control will be
+            # returned to `run` directly
             self.todo.pop()
             raise self.Return(node.value)
         elif isinstance(node, Apply):
--- a/myia/vm.py
+++ b/myia/vm.py
@@ -4,16 +4,14 @@ This VM will directly execute a graph so it should be suitable for
 testing or debugging.  Don't expect stellar performance from this
 implementation.
 """
-from typing import List, Dict, Any
+from typing import List, Dict, Any, Union
 
 from myia.anf_ir import Graph, ANFNode, Constant, Parameter, Apply
 from myia.primops import Primitive, Add, If, Return
 
 
 class VM:
-    """
-    Virtual Machine interface.
-    """
+    """Virtual Machine interface."""
 
     def evaluate(self, graph: Graph, args: List[Any]):
         """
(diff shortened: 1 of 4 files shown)


myia/vm.py
```diff
@@ -0,0 +1,193 @@
...
+        elif isinstance(node, Parameter):
+            idx = self.graph.parameters.index(node)
+            self.values[node] = self.args[idx]
+        elif isinstance(node, Apply):
+            fn = self.get_value(node.inputs[0])
+            if isinstance(fn, Primitive):
+                if isinstance(fn, Add):
+                    args = [self.get_value(i) for i in node.inputs[1:]]
+                    self.values[node] = sum(args[1:], args[0])
+                    raise
```
bartvm: What's this raise for?

bartvm:
Awesome, looks good! Did you try it on the graphs that the parser spits out? If not, I'll try it and use it to write the first few integration tests.

myia/vm.py
```diff
@@ -0,0 +1,193 @@
...
+        return self.values[node]
+
+    def do_call(self, graph: Graph, args: List[Any]):
+        """
+        Perform a call.
+
+        This will handle tail calls correctly by reusing the current VMFrame.
+        """
+        if self.tail():
+            self.reset(graph, args)
```
breuleux: Because (of course) of these damned closures, I don't think you can actually do that: whenever you encounter a `Constant` that contains a `Graph`, and that this graph has the current graph as a parent, you need to attach the frame to it somehow (as an env), so that it can look up its closure variables. This will sometimes entail further computation in this frame, so you can't reuse it for a different graph.

myia/vm.py
```diff
@@ -0,0 +1,193 @@
...
+        This may fail and ask for a missing value to be computed or
+        enter into a call.
+        """
+        if isinstance(node, Constant):
+            self.values[node] = node.value
+        elif isinstance(node, Parameter):
+            idx = self.graph.parameters.index(node)
+            self.values[node] = self.args[idx]
+        elif isinstance(node, Apply):
+            fn = self.get_value(node.inputs[0])
```
breuleux: Since no primitive is lazy or short-circuiting with respect to its arguments, you can safely do

```python
fn, *args = [self.get_value(x) for x in node.inputs]
```

Which will simplify the code below, and will make it easier to implement a generic system for primitives.

myia/vm.py
breuleux: `raise` what? Is that supposed to be there?

abergeron: I belive that is a mistake that I left in.

abergeron:
I haven't run any tests yet.  I'll do that tomorrow, while writing integration tests. If you find some problems before then, don't hesitate to tell me.

myia/vm.py
```diff
@@ -0,0 +1,193 @@
...
+        """Run the graph to completion and return the result."""
+        while True:
+            try:
+                self.advance()
+            except StopIteration:
+                return self.values[self.graph.return_]
+
+    def tail(self):
+        """Are we in a tail call?."""
+        return self.todo[-2] is self.graph.return_
```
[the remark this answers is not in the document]
abergeron: More or less since the last node on the list should always be a Return and the code that handles it doesn't call this.

myia/vm.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def done(self):
+        """Are we done?."""
+        return self.graph.return_ in self.values
```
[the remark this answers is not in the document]
abergeron: Not with the current design, since we put the return node as the last one on the todo list.

If you have an example of a graph that could need to evaluate some other nodes after the return, please show me an example.

myia/vm.py
abergeron: I don't have any form of variable mapping in the frame, so I fail to see how that would help closures.

Also, if we can't reuse frames, then we can't do tail calls.

myia/vm.py
abergeron: Right I did it this way to make sure that If didn't trigger any computing on the branch not taken, but since the branches are functions, then I could do as you say.

myia/vm.py
abergeron: A mistake.

myia/vm.py
```diff
@@ -0,0 +1,193 @@
...
+from myia.anf_ir import Graph, ANFNode, Constant, Parameter, Apply
+from myia.primops import Primitive, Add, If, Return
+
+
+class VM:
+    """
+    Virtual Machine interface.
+
+    Attributes:
+        env: evaluation environement (currently unused)
```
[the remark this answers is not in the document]
abergeron: fixed

myia/vm.py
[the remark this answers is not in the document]
breuleux: I'm thinking of something like this:

```python
def f(x):
    a = x * x
    def g():
        return a
    return g

def h():
    g = f(9)
    return g() + g()
```

There is no direct path from `f.return_` to the node for `a`: it is the graph for `g` that has a pointer to `a` which belongs to `f`. Nonetheless, it needs to be evaluated in `f`'s environment, at most once per `g` closure, and should reuse anything computed when evaluating `f`. However, both calls to `g` happen after `f` is done evaluating, so unless we walk `g` to find `a` and add it to the todo list immediately (i.e. closure conversion), `a` is going to be found after the frame for `f` exited.

myia/vm.py
breuleux: Closures here are graphs that have pointers to nodes that belong to other graphs. The semantics at execution is that if, when evaluating graph `G`, you encounter graph `H` such that `H.parent is G`, then you must construct a closure for `H` that lets it access the values for `G`'s nodes. Each execution of `G` produces a fresh closure for `H` which is a first class value that can be passed around, and as long as that closure lives, so does `G`'s frame or whatever structure holds these values.

A tail call just means that the current frame can be popped from the stack immediately. It can still be kept alive by a closure. If no closure points to it, then the GC will eliminate it.

myia/vm.py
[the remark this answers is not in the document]
abergeron: Oh yeah, I see the problem.  It will attempt to recompute `a`, but the parameters will no longer be available so it will fail or produce bad results.

myia/vm.py
abergeron: I see what you mean, and I will reorganize the inner workings so that this can actually support closures correctly.  As a bonus it should waste less memory when there are no closures.

abergeron:
I've made some major changes to the internals following @breuleux's comments.

Also I've started on some tests, but I'm currently blocked on some expressions that make the parse choke.

abergeron:
I've rebased once again and added a bunch of tests.

While adding tests I've discovered problem when mixing closures and tail calls which are now fixed.  This should be solid now, so I consider it ready for an in-detail review.

myia/vm.py
```diff
@@ -0,0 +1,222 @@
...
+                   for n in dfs(value.return_)):
+                return self.Closure(value, self)
+        return value
+
+    def do_call(self, graph: Union[Graph, 'VMFrame.Closure'], args: List[Any]):
+        """Perform a call.
+
+        This will handle tail calls and closures.
+
+        """
```
breuleux: You should check that there are exactly as many arguments as the graph takes parameters. This would have caught the bug in https://github.com/mila-udem/myia/pull/4#issuecomment-358711256 immediately, and it is virtually certain that it will catch other similar bugs in the future.

myia/vm.py
```diff
@@ -0,0 +1,222 @@
...
+        elif isinstance(node, Parameter):
+            idx = self.graph.parameters.index(node)
+            self.values[node] = self.args[idx]
+        elif isinstance(node, self.Jump):
+            self.todo.pop()
+            raise self.Return(node.value)
+        elif isinstance(node, Apply):
+            fn = self.get_value(node.inputs[0])
+            args = [self.get_value(i) for i in node.inputs[1:]]
+            if isinstance(fn, Primitive):
```
breuleux: We need to figure out where to put primitive implementations. I'd suggest a dictionary mapping primitives to functions, so that we can easily experiment with different implementations. One slight complication is that the implementation for `if`, as it stands, requires the VM.

breuleux:
This crashes with `AttributeError: 'NoneType' object has no attribute 'values'`:

```python
def f(x):
    def g():
        def h():
            return x
        return h
    return g()()
```

The graph looks correct to me, but maybe the multiple nesting is throwing the VM off.

myia/vm.py
```diff
@@ -0,0 +1,222 @@
...
+                raise self.Return(frame)
+            else:
+                raise self.Continue()
+        return frame.values[node]
+
+    def wrap_closure(self, value):
+        """Wrap graphs that are closures."""
+        if isinstance(value, Graph):
+            if any(n.graph and n.graph is not value
+                   for n in dfs(value.return_)):
```
breuleux: Pretty sure this is causing the issue with https://github.com/mila-udem/myia/pull/8#issuecomment-358799988. In short, this will miss free variables: if this is graph `G`, another graph `H` is nested inside `G`, and `H` accesses a variable that's outside the scope of both `G` and `H`, then that variable is a free variable of both graphs. But you're not looking inside the graphs encountered in the `dfs`, so you will miss it.

I think that for this interpreter it's best not to try to be clever: create a closure all the time.

abergeron: If we don't have a reliable way to detect closures, this is the safest choice, yes.  However I fear it will keep all the stacks around for loops and cost a lot of memory.

myia/vm.py
abergeron: We can special case If and have a dict or something for the rest.  I've implemented add inline for now, but I'll move it when we have more.

myia/vm.py
abergeron: added

myia/vm.py
breuleux: It's not the stack per se it's supposed to save, just the environment, so a closure should never keep more "frames" alive than the lexical nesting depth. I think it'd be safe to set a frame's parent to `None` once it returns, so that the GC can reclaim it.

breuleux: Or maybe it'd be better to split functionality into `Environment` (containing `values` and a parent `environment`) and `Frame` (containing a `return_node`, an `environment` and a `todo`), and have closures store the environment only. I think it'd be a bit clearer, and then you can safely reuse `Frame` objects in tail calls and whatnot.

breuleux: But if I understand correctly, `parent` is the caller frame. The caller frame is not part of the environment. What it represents essentially is "what are we doing once we've computed the `return_` node?" If we're missing some nodes and we need to compute them later, we're doing something different afterwards: we're not returning to the parent twice.

So, the way I think it should work is, suppose you have:

```python
def f(x):
    a = x + x
    def g():
        return a
    return g
```

When `f` is called, we create `Frame(f.env, f.return_)`. When `g` is encountered, we create `Closure(g, new_env(f.env))`. Now, when that closure is called and it sees it needs `a`, and it appears `a` has not been computed, then it creates a *new frame*, `Frame(f.env, f.a)`. That frame uses `f.env` so it has access to everything previously computed, but the "return node" is a different one. In other words, you have one environment for `f`, but one frame for every time you need to compute something there.

breuleux: > The more I think about it, the more I realize that we probably need to do a pass to mark which graphs are closures and up to where do we need to preserve the environment to capture everything they need.

Also, just by the way, a precise account of free variables *is* needed for `grad`, so I am working on this right now. I don't think there's any need for this in the VM, but in either case, don't do it.

abergeron: > But if I understand correctly, parent is the caller frame. The caller frame is not part of the environment. What it represents essentially is "what are we doing once we've computed the return_ node?" If we're missing some nodes and we need to compute them later, we're doing something different afterwards: we're not returning to the parent twice.

Yes, that is mostly correct.  But the "missing" nodes could need to be computed in the middle of a nested call.  They could also be required to compute the return value later.  So we need to ensure that all the nodes that are evaluated for a specific application of a graph happen in the same environment.

> When f is called, we create Frame(new_env(f.env), f.return_). When g is encountered, we create Closure(g, f.env). Now, when that closure is called and it sees it needs a, and it appears a has not been computed, then it creates a new frame, Frame(f.env, f.a) (no call to new_env here). That frame uses f.env so it has access to everything previously computed, but the "return node" is a different one. In other words, you have one environment for f, but one frame for every time you need to compute something there.

What is `f.env` in your example?  As far as I know, `f` is a Graph and those don't have an env.  And even if they did it couldn't be the basis for the frame environment since that one needs the parameters for a specific application and the mapping of the values for the nodes that might depend on the values of the parameters.

breuleux: Sorry, I was being loose with notation. You get the env from the frame when you create a closure, and from the closure when you call it. I mean something like this:

```python
def call(fn, args):
    graph = fn.graph if isinstance(fn, Closure) else fn
    parent_env = fn.env if isinstance(fn, Closure) else None
    env = Environment(parent_env=parent_env, graph=graph)
    for param, arg in zip(graph.parameters, args):
        env[param] = arg
    frame = Frame(env=env, ret=graph.return_)
    exec_frame(frame)

def wrap_closure(current_frame, graph):
    return Closure(current_frame.env, graph)

def eval_node(current_frame, node):
    if node.graph is not current_frame.graph:
        env = current_frame.env
        while env.graph is not node.graph:
            env = env.parent_env
        frame = Frame(env=env, ret=node)
        exec_frame(frame)
    ...
```

Is that any clearer? It's a bit difficult to explain, but I think the idea of saving "frames" muddied the waters. It's a red herring, it's just environments that you need to save.

abergeron: I see what you are getting at.  I'll try to rework the code to follow that pattern.

breuleux: Actually, I thought about it more, and I think this won't work, but for a different reason than I thought. Basically, consider:

```python
def f(x):
    def g(z):
        return h(z + x)
    def h(z):
        return g(z * x)
    return g(3)
```

This doesn't terminate, but that's besides the point -- the calls are tail calls, so this should loop forever without growing memory.

However, when you call `g` and you encounter `h`, the code I gave will naively give it `g` as its environment, so it'll have environment chain `[f, g]`. This is wrong, of course, because `h` is not nested in `g`, but this isn't written anywhere in the representation. Nonetheless, when `h` is called, it will see `g`, and it will naturally assume that `g` is nested in `h`, so it will create the environment `[f, g, h]`. Then `[f, g, h, g]`, and so on. The stack won't grow, but the chain of environments will, because both functions think they are nested in each other.

Maybe it'd kinda-work if the closures are cached, but honestly I think we should forget about evaluating this representation directly and preprocess with closure conversion. Then it'll be way more straightforward. I'm already working on that because of `grad`, it's a bit more complicated I thought it'd be, but I think it should be done by next meeting.

abergeron: Yes, it would be much easier to work on a closure-converted version where everything is passed around explicitly.  If it is required for grad to work properly, then we *can* also make it a requirement for the vm, but I would prefer if it wasn't.

We can talk about it before the next meeting, yes.

bartvm: If possible, I agree it would be nice not to have to rely on closure conversion (debugging closure conversion and the parser would be easier that way).

bartvm:
@abergeron Maybe you were aware of this already, but I wasn't: It looks like what this VM does is called a "[spaghetti stack](https://en.wikipedia.org/wiki/Parent_pointer_tree#Use_as_call_stacks)"

abergeron:
I wasn't aware of that, but thanks for the reference.

abergeron:
I did one last rebase and changed the target branch.  This should be mostly good to go.

PR #9, opened 2018-01-17
Make __repr__ valid Python expressions where possible

bartvm:
@breuleux Few small suggested changes to your node naming PR before merging (`__repr__` should be valid Python expressions "if at all possible" according to the docs). Also removing `myia/debug` from code coverage requirements (tests pass right now but only because Codecov doesn't know of the existence of `myia/debug` yet).

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,4 +2,5 @@ ignore:
   - "tests"
   - "docs"
   - "setup.py"
+  - "myia/debug"
 comment: false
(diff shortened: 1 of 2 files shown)


breuleux:
I think `str` on a `Constant` should ideally print the value, not a temporary name.

bartvm:
Ah, yes.

breuleux:
All right, we're all set, I think?

PR #17, opened 2018-01-19
Better string representations of nodes and graphs

bartvm:
Factored this out of #4 so that it is a bit easier to review. This fixes a bunch of issues with the node naming. The current naming:
```pycon
>>> def f(x):
...     return x
>>> parse(f)
<f = Graph(parameters=[x = Parameter(<f = Graph(parameters=[...])>)])>
>>> parse(f).return_
Apply(Inputs([Constant(<myia.primops.Return object at 0x106271630>), x = Parameter(<f = Graph(parameters=[x = Parameter(<f = Graph(parameters=[...])>)])>)]), <f = Graph(parameters=[x = Parameter(<f = Graph(parameters=[...])>)])>)
```
With this PR:
```pycon
>>> parse(f)
<myia.anf_ir.Graph(name=f, parameters=[x], return_=_apply1) object at 0x105c72da0>
>>> parse(f).return_
<myia.anf_ir.Apply(name=_apply2, inputs=[_constant3, x], graph=f) object at 0x105c82eb8>
```
Main changes:
* Make generated names valid variable names
* Base generated names on type of object we are naming
* Use `str` representation where needed to avoid recursively printing the entire graph
* Ensure uniqueness of `repr` by printing object ID like CPython does
* Only represent constant with value if its a literal (to avoid printing really long strings when its value is e.g. a graph)

### DIFF
--- a/myia/anf_ir.py
+++ b/myia/anf_ir.py
@@ -15,10 +15,12 @@ from typing import (List, Set, Tuple, Any, Sequence, MutableSequence,
                     overload, Iterable)
 
 from myia.ir import Node
-from myia.utils import Named
+from myia.utils import Named, repr_, list_str
 
 PARAMETER = Named('PARAMETER')
 APPLY = Named('APPLY')
+
+LITERALS = (bool, int, str, float)
 
 
 class Debug(types.SimpleNamespace):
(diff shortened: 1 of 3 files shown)


breuleux:
I think we have a bit of a problem of having different workflows, whereas you use these string representations, and I use buche. That's fine in principle, but regarding e.g. automatic naming, I don't actually want the node type in there because I have it color coded. So ideally maybe find a way to make it so that `str(node)` can return `_apply2`, but buche can display something terser like `#2`, considering we can visually tell what kind of node it is.

bartvm:
When I plot stuff using graphviz I don't use str, but a separate naming
function that names things so that they make sense in the plot. Probably
makes sense to do something like that for buche, and keep repr/str for
printing in the REPL, logs, etc.

bartvm:
I changed it so that each object has a unique ID in an `id` attribute that can be used for debugging (you could use `id(obj)` I guess, but 0, 1, 2... is easier to track). So now we have:

`id`: Unique, sequential number to identify each node/graph.
`debug_name`: A variable name-friendly tag that always exists (not enforced to be unique). Can be used to refer to nodes (e.g. `Apply(inputs=[debug_name1, ...])`).
`name`: A variable-name friendly tag which is meaningful (i.e. it was the original variable name or something the parser named), not enforced to be unique, can be `None`. If it's given, we assume it means something to the user.
`str`: A printable string for each object (often `debug_name`, but for literals we use the value)
`repr`: Unique, informative string for debugging

breuleux:
Sure, that works, except I think the id should be created on demand. If I only require these ids for a few nodes, I'd rather see 1, 2, 3, ... than 3, 7, 55, ...

bartvm:
Done.

breuleux:
OK, merge it. (Also, are you amending the one commit each time? I'd prefer not so I can see individual diffs, you can squash when merging).

bartvm:
I squashed it already since it was such a minor change.

PR #18, opened 2018-01-22
Add output property to Graph

breuleux:
The purpose of `graph.return_` is so that each graph has at least one node that belongs to it, so that various operations like node replacement are facilitated. This node is always the application of the `Return` primitive, so the real output is in `graph.return_.inputs[1]`. This adds an `output` property that serves as a shortcut to get and set that real output.

### DIFF
--- a/myia/anf_ir.py
+++ b/myia/anf_ir.py
@@ -97,7 +97,10 @@ class Graph:
     @output.setter
     def output(self, value: 'ANFNode') -> None:
         """Set the graph's output."""
-        self.return_ = Apply([Constant(RETURN), value], self)
+        if self.return_:
+            self.return_.inputs[1] = value
+        else:
+            self.return_ = Apply([Constant(RETURN), value], self)
 
     def __str__(self) -> str:
         """Return readable string representation."""
--- a/myia/anf_ir.py
+++ b/myia/anf_ir.py
@@ -16,11 +16,14 @@ from typing import (List, Set, Tuple, Any, Sequence, MutableSequence,
 
 from myia.ir import Node
 from myia.utils import Named, repr_, list_str
+from .primops import Return
 
 PARAMETER = Named('PARAMETER')
 APPLY = Named('APPLY')
 
 LITERALS = (bool, int, str, float)
+
+RETURN = Return()
 
 
 class Debug(types.SimpleNamespace):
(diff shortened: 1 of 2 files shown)


myia/anf_ir.py
```diff
@@ -79,6 +82,26 @@ def __init__(self) -> None:
...
+            raise Exception('Graph has no output.')
+        return self.return_.inputs[1]
+
+    @output.setter
+    def output(self, value: 'ANFNode') -> None:
+        """Set the graph's output."""
+        if self.return_:
+            self.return_.inputs[1] = value
+        else:
+            self.return_ = Apply([Constant(RETURN), value], self)
```
bartvm: There should probably be only one node `Constant(RETURN)` in the entire graph, not a different node for each function. One the other hand, we should be able to create multiple `Constant(RETURN)` objects if we want to keep two programs entirely separate (allowing us to e.g. look at `Constant(RETURN).uses` without mixing up two programs).

This is general thing: We don't want a new `Constant(1)` node each time the constant 1 is used somewhere. I am not entirely sure where this logic should be handled though...

abergeron: I don't think it's a major problem for now if we have multiple `Constant(RETURN)` for now.  We can push that for later.

breuleux: Yeah I was wondering about what to do about that, but I don't think merging constants and other nodes immediately is a particularly important or pressing issue.

bartvm: Okay, let's merge it then but please make an issue about duplicate nodes. Note that it's not just constants, any node whose `(inputs, value, graph)` values are identical shouldn't be duplicated (that's the idea of doing CSE at graph construction).

myia/anf_ir.py
```diff
@@ -16,12 +16,15 @@
 
 from myia.ir import Node
 from myia.utils import Named, repr_, list_str
+from .primops import Return
```
bartvm: Missed this, but please don't do relative imports. (It's quite strongly discouraged.)

abergeron: Why do you say it's discouraged.  I've heard that about old-style relative imports (the kind that don't work in python 3), but never about the new style.

bartvm: > Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path)

> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose.

Regardless, we should at least be consistent. So either all the imports should be relative, or all should be absolute, not a mix.

breuleux: What's discouraged are implicit relative imports like `from primops import ...`, not explicit relative imports like `from .primops import ...`. Your quote says these are an acceptable alternative. Personally I vote for relative imports.

abergeron: I just want to point out that non-explicit relative imports (as in `from primops import ...`) are no longer possible with python 3.  What @bartvm quoted is old advice that only applies to python 2.

In the current situation I slightly favor absolute import because we are bound to transform some files into modules and absolute imports will keep working if we do that.  In the long run I usually favour relative import because that's less typing.

breuleux: I'm going to change it for absolute import for consistency, we can figure out the final policy in another issue.

bartvm: @abergeron Advice I quoted is for the current PEP8 for Python 3. As @breuleux pointed out, guideline is that implicit relative imports are evil, explicit relative imports are "acceptable alternatives", absolute imports are "recommended".

PR #19, opened 2018-01-22
Retrieve graph nesting structure and free variables

breuleux:
This adds the `myia.cconv.NestingAnalyzer` class, used as follows:

```python
def f(x):
    def g():
        return x
    return g

f_graph = parse(f)

na = NestingAnalyzer().run(f_graph)
assert na.parents == {f_graph: None, g_graph: f_graph}
assert na.fvs == {f_graph: set(), g_graph: {x_node}}
```

(This won't run because `g_graph` etc. are not set to the corresponding graphs and nodes, but it should illustrate the point nonetheless.)


myia/cconv.py
```diff
@@ -0,0 +1,165 @@
...
+
+from typing import Set, Dict, Optional, Union
+from collections import defaultdict
+from .anf_ir import ANFNode, Constant, Graph
+from .anf_ir_utils import dfs
+
+
+def is_graph(x: ANFNode) -> bool:
+    """Return whether x is a Constant with a Graph value."""
+    return isinstance(x, Constant) and isinstance(x.value, Graph)
```
abergeron: This should be in anf_ir_utils

myia/cconv.py
```diff
@@ -0,0 +1,165 @@
...
+            it has no parent.
+        fvs: Maps a graph to a set of nodes it refers to from other
+            graphs, including nodes nested graphs have to access.
+
+    """
+
+    def __init__(self) -> None:
+        """Initialize the NestingAnalyzer."""
+        self.processed: Set[Graph] = set()
+        self.parents: Dict[Graph, Optional[Graph]] = {}
```
abergeron: It would be nice to have the inverse of parents.

myia/cconv.py
```diff
@@ -0,0 +1,165 @@
...
+                       for g, gs in self.deps.items()}
+        self.parents.update(new_parents)  # type: ignore
+
+        # We find all free variables (could be interleaved with other
+        # phases, probably)
+        self.associate_free_variables()
+
+        return self
+
+    def compute_deps(self) -> None:
```
abergeron: I undertand that your logic has different steps, but you might not want to expose them as public functions.

bartvm:
I mentioned in the gradient transform PR several changes that I think would be good to have the ability to get the free variables and the implicit nesting separately, and I'd like clarified whether `_simplify` does topological sorting, and if so, in the correct complexity, and if that is true too, ideally I would like a more reusable implementation of toposort.

At the very least, I would like the bottom line of the long discussion we had on the other PR regarding the two different types of nesting reflected in the documentation here.

abergeron:
I have a simple toposort implementation in my branch for the type inference.  I intend to commit that along with the rest, but if it is useful, I can make a separate PR with it.

As for the rest I think the free variables are accessible through `.fvs` and nesting is in `.parents`/`.children`

breuleux:
`_simplify` doesn't do topological sorting per se, it does specifically what it is supposed to do, which is to find the closest parent of a graph, under the assumption that there cannot be more than one parent. You could do it with a topological sort, but it's less direct and probably less efficient.

`fvs` contains a set of free variables for each graph. However, it takes implicit nesting into account: if `f` is a parent of `g`, `g` is considered a free variable of every graph nested in `f`. The non-implicit dependencies are done by `_compute_deps` but `_resolve_proxies` dumps the implicit ones in the same place.

Anyway, I'll fix it up.

breuleux:
@abergeron I do think you should make a PR for `toposort`, that way it'll already be there when someone else inevitably needs it.

bartvm:
@breuleux Not entirely sure what you mean by "finding the closest parent". My thought process:
```python
def f(x):
    def g(y):
        def h():
            return x + y
        return h
    return g(x + 1)
```
Then we have something like `free_vars_deps = {h: {f, g}}` and `closure_deps = {h: {g}, g: {f}}`. Combined, these relationships define a small DAG, of which you want to find . You can do a topological sort, *O(n)*, and if a function has multiple parents you pick the first one from the topological sort. My guess is that will give you the tree you want. The other option is to find the transitive reduction of the graph, but that's *O(nm)* in general, so my guess is that you're just doing a toposort. How is `_simplify` more efficient or more direct than that?

breuleux:
> if a function has multiple parents you pick the first one from the topological sort

Sure, but what's the complexity of *that?* Maybe I'm missing something or misrepresenting what you have in mind, but if you do one topological sort for the whole graph, there can be up to O(n) unrelated nodes sandwiched between a node and its closest parent in the sort, so naive looping to find it would lead to O(n^2) in the worst case for the complete algorithm (to find all relationships). If, on the other hand, you're running one topological sort for each node, restricted to its dependencies, then it's easy, but you have to preprocess and you're performing redundant work (so it'd be O(dn) where d is the outdegree -- to be fair that's also what I think the complexity of `_simplify` is, after optimizing it a little).

bartvm:
You can construct the mapping from parents to children in linear time. Then
you just walk the nodes in topological order and for each child of the
current node you check if a parent has been defined yet. If not, the
current node is set to be the parent. That's linear time.

breuleux:
I have remade NestingAnalyzer to provide more analyses, including the "direct" list of free variables. It is still a class-based design, where each step is memoized, because I think that's the best interface if you need to call several methods and want to reuse any intermediate results.

I also added a generic `dfs` that takes a successor function and rewrote the one `anf_ir_utils` to use that (it also fixes a bug with it where it sometimes yielded the same nodes multiple times). The generic graph utils should go in `myia.graph_utils`.

breuleux:
Is this good to merge?

bartvm:
Sorry, I haven't had the time to look at this in detail yet. Is the memoization really a good idea? Seems to me that if the graph is mutated, and you call any of these methods again, you might get incorrect results i.e. the user must keep in mind that whenever they mutate a graph, the `NestingAnalyzer` is invalidated and a new one must be constructed. This seems like a bug waiting to happen...

breuleux:
The memoizing is so that relevant intermediate steps can be shared if you call more than one method on the object.

I could add a set of top-level functions that correspond to each method and take the root graph as an argument as a safer alternative when you only need one thing. Or a function that takes a list of analyses and returns a dictionary.

bartvm:
Can we just do the naive but always correct thing for now and add optimizations later on at the point that we actually need them?

breuleux:
I'm using more than one method in grad, so it saves some work there.

bartvm:
"Need" as in when we get to the point that parsing and compilation time becomes an issue.

PR #22, opened 2018-01-24
Gradient transform

breuleux:
This provides the `Grad` transform on graphs. Call `Grad().process_graph(graph)` to obtain the transformed graph.

**Current state**: `Grad` outputs something that looks sensible for all graphs I've tested, including closures and graphs with if/while loops. I also tried up to sixth order gradient on a while loop and it didn't crash, which is encouraging (sixth order takes 30 seconds, so I stopped there). However, I didn't run any of these graphs (can't, since the necessary operations are not implemented), so I don't know if they are indeed correct (the graphs can be a bit hard to read, unfortunately).


myia/cconv.py
```diff
@@ -0,0 +1,165 @@
...
+        self.nodes = set(dfs(Constant(root), True))
+
+        # We get all graphs possibly accessed.
+        self.graphs = {node.value for node in self.nodes if is_graph(node)}
+        self.processed |= self.graphs
+
+        # We compute the dependencies:
+        # * If a node in graph g has an input from graph g', then g depends
+        #   on g'
+        # * If a node in graph g refers to a graph g', then g depends on
```
bartvm: Is this true? For example:
```python
def f(x):
    return x

def g():
    return f
```
I think you only need to look at the free variables.

myia/cconv.py
```diff
@@ -0,0 +1,165 @@
...
+                else:
+                    parents.add(dep)
+            return parents - {g}
+
+        newdeps = {}
+        for g in self.graphs:
+            newdeps[g] = seek_parents(g)
+        self.deps = newdeps
+
+    def simplify(self) -> None:
```
bartvm: If you have a full dependency graph and you only want to keep the closest one, that sounds to me like you're trying to do a series of topological sorts (and I think this function kind of does that, but not very efficiently).

It's worth noting that I don't think we can always guarantee a unique ordering e.g. both of these functions are valid nestings of the same graph.
```python
def main():
    def f(x):
        def g(y):
            def h():
                return x + y
            return h
        return g(1)
    return f(2)

def main():
    def g(y):
        def f(x):
            def h():
                return x + y
        return f(2)
    return g(1)
```

myia/cconv.py
breuleux: It's true: the minimal possible nesting of `g` is the minimal nesting of `f`'s parent, which in this case happens to be `global` (if it was nested below, it wouldn't be able to see `g`). I suppose the main reason I did it this way is that it allows part of the algorithm to be run non-recursively and it made it easier for me to see what's going on (some free variables belong to no graphs, because they *are* graphs that we have yet to place in the hierarchy, and that confused me a bit at first). All this being said, I'm pretty sure it can be simplified, if need be.

myia/grad.py
```diff
@@ -0,0 +1,284 @@
...
+def is_constant(x):
+    return isinstance(x, Constant)
+
+
+def is_graph(x):
+    return isinstance(x, Constant) and isinstance(x.value, Graph)
+
+
+add = Constant(primops.Add())
+J = Constant(primops.J())
```
bartvm: I might have asked this before, but I can't remember the answer: Why do we want to have `J` as a primitive? Why not "evaluate" `J` directly at parse-time here? It only seems to be applied to constants after all.

myia/cconv.py
breuleux: These are equivalent graphs, but they are not the same graph...

myia/grad.py
```diff
@@ -0,0 +1,284 @@
...
+
+        if len(contribs) == 0:
+            # No contributions means a gradient of zero, naturally.
+            sens = Constant(0)  # TODO: should be zeros_like(node)
+        elif len(contribs) == 1:
+            # A single contribution can be used directly.
+            sens, = contribs
+        else:
+            # More than one contribution must be added trogether.
+            sens = Apply([add, *contribs], bg)
```
bartvm: Just a note that although `add` is currently implemented as `sum` in the VM, it's still an `add`, and we probably should stick to these simpler semantics for now and leave `sum` for a later part in the optimization pipeline.

myia/grad.py
```diff
@@ -0,0 +1,284 @@
+
+from typing import Set
+
+from .anf_ir import Apply, Constant, Parameter, Graph, ANFNode
```
bartvm: Let's use absolute imports for now, or change everything to relative imports in a single PR, but not mix the convention across files.

myia/grad.py
```diff
@@ -0,0 +1,284 @@
...
+from collections import defaultdict
+from .anf_ir_utils import dfs
+from .cconv import NestingAnalyzer
+
+
+def is_constant(x):
+    return isinstance(x, Constant)
+
+
+def is_graph(x):
```
bartvm: These two helper functions should probably go into `anf_ir_utils` (and `is_graph` should probably be `is_graph_node`, can get a bit confusing otherwise).

myia/grad.py
breuleux: `J` will not always be applied to constants. It will also be applied to e.g. arguments to higher order functions, in which case we need to do some value propagation. We might need a runtime version of `J` that looks up the proper pre-transformed function in a static map.

Also, it's quite a bit simpler to implement and debug a transformation if you don't have to worry about recursively expanding macros. Many operations we wish to eliminate can exist in the IR transiently, I don't think this is a problem.

myia/grad.py
breuleux: Sure.

myia/grad.py
breuleux: Yeah, some cleanup's definitely needed here.

myia/grad.py
```diff
@@ -0,0 +1,284 @@
...
+        tg = node.graph and self.tagged_graphs[node.graph]
+
+        if is_graph(node):
+            # We will have to process this graph too.
+            tagged, bprop = self.scaffold_graph(node.value)
+            self.graph_to_ct[node.value].add(node)
+        elif is_constant(node):
+            # Note that this application will have its graph set to None, which
+            # makes sense since it's basically a constant expression.
+            tagged, bprop = Apply([J, node], tg), None
```
bartvm: I'm not sure about changing the semantics this way. It seems to complicate things like VMs and scheduling i.e. when do you actually evaluate these expressions? And when do they go out of scope? Ideally we should just "evaluate" them at compile time, or otherwise we can wrap the entire graph in a global graph whose only job it is to evaluate these nodes and then start the rest of the calculations.

breuleux: Well, there was already the idea before of wrapping all graphs in a global graph which would be fed the global variables. That seems like an appropriate scope for all these computations.

myia/cconv.py
bartvm: What's the difference after parsing them?

myia/grad.py
bartvm: Do you have a case in which a runtime `J` primitive is required? If you want to keep `J` around for one pass and remove it another, sure, but I'm wondering whether we can get rid of `J` before runtime/codegen, in which case we probably should. Targeting different backends would be easier if we don't have AD-specific primitives floating around.

myia/grad.py
```diff
@@ -0,0 +1,284 @@
...
+        else:
```
bartvm: Reading this code in isolation makes it look like there's a mistake because it ignores the possibility of `node` being a `Parameter` instance. This isn't possible because they were added to `tagged_nodes` by `scaffold_graph`, but an extra `assert` or `isinstance` test would make that a bit clearer for people reading the code.

myia/cconv.py
breuleux: Well, for one, in the first one, `main.output.inputs` is `[f, 2]`. In the second one, it is `[g, 1]`. In the first case, `g` has to build the closure for `h`, so it needs `x`, and has to be in `f`. In the second case it is `f` that builds the closure and must be in `g`.

myia/grad.py
breuleux: I don't know, and I'm not too concerned about that for now. The Python VM can implement that primitive to run the transformer, which could be nice for debugging, and the same system if tasked with constant prop can take care of eliminating it. As for a potential C or Rust VM, it would basically be just a lookup, but I think for now we can optimistically assume we won't need it.

myia/grad.py
breuleux: That's a good point. I will make this clearer.

myia/grad.py
```diff
@@ -0,0 +1,284 @@
...
+            # a = f(x, y) -> ↑a, ♢a = ↑f(↑x, ↑y)
+            tagged_args = [self.phi(n) for n in node.inputs]
+            app = Apply(tagged_args, tg)
+            # ↑a (the first element)
+            tagged = Apply([index, app, Constant(0)], tg)
+            # ♢a (the second element)
+            # Note that ♢a is not part of the forward graph, however,
+            # it will be a free variable of the backpropagator graph.
+            bprop = Apply([index, app, Constant(1)], tg)
```
bartvm: Do I understand this code correctly that primitives and functions are treated equally? Should we treat them equally? It seems like it's best to assume that the adjoints of primitives should always be inlined in the backpropagator, and perform this inlining immediately. I imagine this simplifies the graph a lot. So instead of this:
```python
def pow(x, y):
    def dpow(dz):
        return y * x ** (y - 1) * dz, x ** y * log(x) * dz
    return x ** y, dpow

def f(x):
    y, dpow = pow(x, 2)
    def df(dy):
        dx, _ = dpow(dy)
        return dx
    return y
```
can we generate this?
```python
def f(x):
    y = x ** 2
    def df(dy):
        return 2 * x * dy
    return y
```
After all, we know that primitives never have free variables, and we don't want to incur function call overhead each time a primitive's adjoint is used.

myia/cconv.py
bartvm: > In the first case, g has to build the closure for h, so it needs x, and has to be in f. In the second case it is f that builds the closure and must be in g.

I'm not sure I see how that follows. We have `Apply([RETURN, h], g)` in one graph, and `Apply([RETURN, h], f)` in the other. As I said in one of my other comments, returning a function doesn't imply nesting. Hence, these return statements don't tell you that `f` is in `g` or vice versa. What you do know, from `h`'s free variables, is that `h` needs to be in `f` _and_ in `g`.  Now, if Python code, your functions must have a total order (i.e. `g -> f -> h` or `f -> g -> h`), but in our graph representation, both of these nestings are valid for this graph.

breuleux: Well, let's keep it simple: if `f` is nested in `g`, `main` cannot possibly call `f` directly, because there is a different `f` for every call of `g`. So if `main` calls `f`, the nesting `g->f->h` can't possibly work. There's no need to look further than that, I think.

In any case, I don't think it is possible for `h` to depend on `f` and `g` without either `f` or `g` depending on the other. There is no rule which instances of `f` and `g` to pair for an instance of `h` in that situation, so I don't think it has any semantics.

myia/grad.py
breuleux: We can optimize the graph into that. I don't see a good reason to do it immediately: I think that's unnecessary clutter and coupling with the optimization framework, and the less `Grad` does, the easier it will be to debug.

myia/cconv.py
bartvm: How is `f` nested in `g`? In the Python code `f` is explicitly nested in `g`, but in the graph this is no longer the case. Nesting is implicit by using a variable from another function's scope, and `f` and `g` do not access each other's scopes.

If you would try to turn the graph back into Python code, then there is only one valid nesting, but that is a Python limitation: In Python, a function must have exactly one parent i.e. the nesting graph is a tree. This is not true in Myia: In Myia, the nesting is only constrained to be a non-cyclical graph. In Myia-land, it's perfectly reasonable to say that `h` is nested in both `f` and `g`, and that `f` and `g` are not nested in each other. So why would we want to enforce a tree? Maybe it is useful, but that's not immediately obvious to me that it is.

This is a Python illustration of what the Myia graph really looks like:
```python
def main():
    return f(2)
def g(y):
    def h():
        return x + y
    # when the closure gets created, h is bound to the scope of both f and g
    # f and g must both be in the call tree, but we don't care in what order
    return h
def f(x):
    # this h is the same as the h above; no way of writing this in Python
    @in_this_scope_too
    def h():
        pass
    return g(1)
```

breuleux: > Nesting is implicit by using a variable from another function's scope, and `f` and `g` do not access each other's scopes.

In the operational semantics, `g` has to access `f`'s scope. It's the only arrangement that allows `h` to be instantiated. Let me put it this way:

Background premises:

1. `f` is nested in `g` if `f` must access one of `g`'s free variables.
1. A closure must have access to a pointer to its free variables.
1. A closure must be constructed: *some* function (possibly more than one) has to pack the function pointer with the environment pointer.
1. The function that constructs the closure must therefore access the closure's environment pointer, and by extension every free variable in the closure.
1. If `f` accesses closure `g`, but does not build `g`, then `g` is a free variable in `f` and must point to its instantiation in a different scope.
1. Two different functions cannot be nested in each other.

Now:

* Suppose that `g` builds `h`. Then `g` must have access to `x` and therefore it is nested in `f` (1). That seems fine.
* Suppose that `f` builds `h`. Then `f` must have access to `y` and therefore it is nested in `g` (1). That seems fine so far. But wait! `g` accesses `h`, but `h` is built by `f`. Therefore, `h` is a free variable in `g`, belonging to `f` (5), which means that `g` must be nested in `f` (1). So this is impossible (6).
* Suppose that neither `f` nor `g` is nested in each other. By contraposition of (4) this entails that `h` is built by neither `f` nor `g`. But then where and how is `h` built? This is not runnable (3).
* Suppose that both `f` and `g` build `h`. Then both must access each other's variables and are nested in each other (1). This is impossible (6).

> `# f and g must both be in the call tree, but we don't care in what order`

Consider that `f` calls `g`, but `g` doesn't call `f` or even refers to it. Clearly there is only one order they can possibly have in the call tree.

bartvm: You are confusing two types of nesting: The lexical scope, and the call tree. In a traditional representation these are constrained in ways that they are not in our representation.

> Suppose that g builds h. Then g must have access to x and therefore it is nested in f (1). That seems fine.

If `g` builds `h`, then we know that `f` must be in the call tree at the time of construction, so that the scopes of `f` and `g` are both live. In terms of lexical scope, we only know that `h` is lexically nested in `f` and in `g`. I think you are confusing these two constraints.

> Suppose that f builds h. Then f must have access to y and therefore it is nested in g (1). That seems fine so far. But wait! g accesses h, but h is built by f. Therefore, h is a free variable in g, belonging to f (5), which means that g must be nested in f (1). So this is impossible (6).

If `f` builds `h`, then we know that `g` must be in the call tree at this point. From looking at `h`'s free variables, again, we know that `h` must be lexically nested in both `g` and `f`. So we haven't learned anything new, and there is no contradiction: `h` must be lexically nested in both `f` and `g`, and at the time that `h` is constructed, `g` and `f` must be in the call tree.

> Suppose that neither f nor g is nested in each other. By contraposition of (4) this entails that h is built by neither f nor g. But then where and how is h built? This is not runnable (3).

`f` and `g` can be not lexically nested, while still being part of the same call tree. This means that when `h` is constructed by either function, as long as both functions are in the call tree at the point of construction, `h` has access to the scopes it needs to capture its free variables.

> Suppose that both f and g build h. Then both must access each other's variables and are nested in each other (1). This is impossible (6).

`f` and `g` can build `h` if they are nested in the cal ltree, which doesn't mean they are nested lexically, so this is not impossible.

bartvm: This is an example of `f` and `g` both building `h`:
```python
def main():
    h1 = f(False, 2)  # A call with c = True would be invalid
    h2 = g(False, 1)  # These two calls return the same closure
    assert h1() == h2()
def g(c, y):
    def h():
        return x + y
    if c:
        return h
    else:
        return f(True, 2)
def f(c, x):
    # h is lexically nested in both f and g
    # there is simply no way of writing this in Python
    def h():
        return x + y
    if c:
        return h
    else:
        return g(True, 1)
```

breuleux: `f` and `g` are lexically nested in each other in that example. It's clearer if you rewrite it like this:

```python
def main():
    h1 = f(False, 2)(True, 1)  # A call with c = True would be invalid
    h2 = g(False, 1)(True, 2)  # These two calls return the same closure
    assert h1() == h2()
def g(c, y):
    def h():
        return x + y
    if c:
        return h
    else:
        return f
def f(c, x):
    # h is lexically nested in both f and g
    # there is simply no way of writing this in Python
    def h():
        return x + y
    if c:
        return h
    else:
        return g
```

When `f` returns `g`, it needs to pack `x` with it, otherwise `g` won't be able to build `h`, and likewise the other way. Both `f` and `g` need to access each other's variables in certain circumstances. I suppose the moral of this story is that Myia can represent cyclical nesting patterns, and a VM could be made to run it. Not that surprising, considering it can also represent this directly cyclical pattern, which would work just as well as your example:

```python
def f(c, x):
    return x + y if c else g(True, 1)
def g(c, y):
    return x + y if c else f(True, 2)
```

bartvm: > When f returns g, it needs to pack x with it, otherwise g won't be able to build h, and likewise the other way. Both f and g need to access each other's variables in certain circumstances. [...] I suppose the moral of this story is that Myia can represent cyclical nesting patterns, and a VM could be made to run it.

I think we are slightly closer to understanding each other. You say `f` and `g` are cyclically nested, I say they are not at all nested. Both are correct depending on what you call "nested". I say that `f` is nested in `g` if it accesses a variable from `g`'s scope. Your definition of nested is wider, because you say that `f` is also nested in `g` if calling `f` involves the construction of a closure which contains a reference to `g`. (This is why I said: One function returning another function--constructing a closure--does not imply nesting.)

For sake of semantics: Lexical scope is separate from the call stack. Whether `g` accesses `f`'s scope or vice versa depends on the call stack, so I wouldn't refer to this as lexical scope. I would say that `f` and `g` are not lexically nested, `h` is lexically nested in both `f` and `g`, but `f` and `g` have a cyclical _runtime dependency_ (i.e. one must call the other at runtime in order to be a valid program).

breuleux: >  I say that f is nested in g if it accesses a variable from g's scope. Your definition of nested is wider, because you say that f is also nested in g if calling f involves the construction of a closure which contains a reference to g.

A reference to g's free variables, but notwithstanding that detail, these definitions are the same. The construction of a closure involves accessing all the variables that the closure accesses, or a structure through which they can all be accessed, which amounts to the same thing.

> Whether g accesses f's scope or vice versa depends on the call stack, so I wouldn't refer to this as lexical scope.

No, it depends on whether `c` is true or false. It just so happens that if `c` has the wrong value, the function will try to read from an uninitialized environment and it will crash. In practice, the generated code for `f` will wrap its reference to `g` as `make_closure(g, self.env)` and `g` will wrap its reference to `f` as `make_closure(f, self.env)`. That is the implementation of lexical scoping semantics.

If you actually got free variables from the call stack, you would have dynamic scoping, like Lisp used to have in the old days. This is not what we have with Myia. What we have is lexical scoping.

> f and g have a cyclical runtime dependency (i.e. one must call the other at runtime in order to be a valid program).

Neither `f` nor `g` has to *call* the other. `f` can return `g` without calling it.

bartvm: > A reference to g's free variables, but notwithstanding that detail, these definitions are the same.

I don't see how these are the same. In our graph representation, `f` being lexically nested in `g` because `f` accesses `g` is explicit because there will be a node in `f` with `n.inputs[i].graph == g`.

Accessing a variable by constructing a closure is entirely different: If a function `g` returns some function `f`, that does not imply lexical nesting. In general, you statically don't even know what the function `f` is, so you don't know what the free variables are, so you don't know what should be nested in what.

> No, it depends on whether c is true or false. [...]

I'm going by the Wikipedia definition here, which says that "In languages with lexical scope (also called static scope), name resolution depends on the location in the source code and the lexical context, which is defined by where the named variable or function is defined." 

If `c` determines whether `f` accesses `g`'s scope or vice versa, and `c` is potentially only known at runtime, then saying that `f` and `g` are lexically (statically) nested doesn't sound right. What we know statically from the graph is that `h` accesses `f` and `g`; this is the lexical nesting.

> Neither f nor g has to call the other. f can return g without calling it.

Sure, the point is at the time that `h` is constructed, for the program not to crash, it must have access to environments of both `f` and `g`, which means that both must have been called at one point.

breuleux: > In our graph representation, `f` being lexically nested in `g` because `f` accesses `g` is explicit because there will be a node in `f` with `n.inputs[i].graph == g`.

What I'm saying is that there is also *implicit* lexical nesting in the graph. The absence of a direct occurrence of a node in `f` that belongs to another graph `g` *does not entail* that `f` is not nested in `g`.

> In general, you statically don't even know what the function `f` is, so you don't know what the free variables are, so you don't know what should be nested in what.

We're not talking about functions that are given as a parameter, obtained as the result of a call, or are free variables. All closures obtained through these means have already been constructed. The problem, *specifically*, is when you encounter a `Constant` containing a `Graph` which points to free variables. That graph is a specification that in order to exist, this function needs this and that variable from this and that scope. In and of itself, though, *that does not tell you how the variables are supposed to get there.* This needs to be defined, and depending on how you do it, you may find new nesting relationships that are necessary under the semantics chosen. So there are at least two options:

* You can say that when the function is used, the free variables are fetched from the call stack. That's dynamic scoping, that's completely inconsistent with Python semantics, and we are not doing it.
* Or you can say that when a reference to the function is obtained, all the free variables are accessed and packed with the function. This means a construction operator must be inserted in the graph. And once it is inserted, the graph may suddenly have new free variables, so you keep going until everything stabilizes. This is closure conversion, and it is in line with Python semantics.
* Whatever else it is you have in mind, but if you're not inserting construction operators and adding free variables and nesting relationships as needed, I don't know how you're going to preserve Python semantics.

I suppose we could say that our graph representation is underspecified: it contains information about which functions need variables from which other functions, but it doesn't tell us how that's supposed to work. Any methodology in line with Python semantics requires additional work to be performed in certain scopes, and when that work is made explicit, additional nesting relationships are revealed.

bartvm: > What I'm saying is that there is also implicit lexical nesting in the graph.

So we agree that there are two types of nesting. My questions are about trying to understand what you call "implicit nesting". It is clearly of a different nature, for example, "explicit" nesting can never be cyclical, but this type of nesting can be.

> That graph is a specification that in order to exist, this function needs this and that variable from this and that scope. In and of itself, though, that does not tell you how the variables are supposed to get there. This needs to be defined, [...]

Why does this need to be defined? Most of your references are to "closure conversion" or "Python semantics", but I have been purely thinking about automatic differentiation.

For AD it is obvious to me that I care that `h` has references to `x` and `y` in `f` and `g`, because that means that we need to return environment sensitivities for those two variables when constructing the backpropagator of `h`.

However, why do we care about the "implicit", cyclical nesting of `f` and `g`? In your AD code I only see a call to `nested_in` for variables that have uses in another graph i.e. explicit nesting.

If I try to think of what the backpropagator of `f` looks like in the examples we discussed, I don't immediately see the use of knowing what the cyclically, implicit nesting relationship of `f` and `g` is.
```python
def df(denv):
    dx = denv.pop('x', 0)
    if c:
        return 0, dx, denv
    else:
        _, _, denvg = dg(denv)
        return 0, dx, denv + denvg
```
The main thing that comes to mind is unpacking your environment sensitivities: In this example `denv` can either contain just `x`, or both `x` and `y`, depending on whether `g` called `f` or vice versa. Knowing that would allow you to write something static like `denv[0]` instead of something runtime dependent like `denv.pop('x', 0)` (with some unique node ID, not just `'x'`). But (a) is that useful? and (b) I don't think the current code handles this correctly?

breuleux: > Why does this need to be defined? Most of your references are to "closure conversion" or "Python semantics", but I have been purely thinking about automatic differentiation.

It needs to be defined because the return value of the function depends on it, and if the return value depends on it, so does the AD, obviously. Take the following program, for example:

```python
def f(idx, g1):
    def g2():
        return idx
    def g3():
        return g2
    def g4():
        return g3
    return [g1, g2, g3, g4][idx]()

f(0, f(0, f(3, None)))
```

What does this return? Under lexical scoping, `f(3, None)` calls `g4`, which returns the closure `g3{idx=3}`, then `f(0, _)` calls `g3{idx=3}` yielding `g2{idx=3}`, and then that's called and the result is `3`. We say that `g3` and `g4` are "nested" in `f` because they both have to carry one of `f`'s variables, `idx`, to `g2`.

Now, you have said a few things that have confused me: "f and g do not access each other's scopes," "If g builds h, then we know that f must be in the call tree at the time of construction," "Whether g accesses f's scope or vice versa depends on the call stack," and so on. So let's see what this means for our little program above. First, I call `f(3, None)`. This has to return `g3`. By your criteria, `g3` isn't nested in `f`, so we're just returning `g3`, right? Then we call `f(0, _)` and we have to return `g2`, which *is* nested in `f`, so we construct a closure. But we need `idx`. That "depends on the call stack," so `f` must be in it, and will you look at that, it *is*! And the value of `idx` in `f` on the call stack is `0`. So the result is `0`.

I have no idea if this is what you had in mind. If it is, this is flat out inconsistent with both Python semantics and the semantics of the language our AD transform is defined on. If it isn't, I have to say you have completely lost me, from the very beginning. The above is what the words you used suggest to me.

> In your AD code I only see a call to `nested_in` for variables that have uses in another graph i.e. explicit nesting.

Notice that `graph` is an explicit argument to `rho`: for each scope in which some variable `x` is used, we need to compute a gradient contribution, but a nested scope isn't going to add contributions from a parent scope (lest you want to end up with an *explicit* nesting cycle in the backpropagators), nor from sibling scopes, although it will need to add contributions from children scopes. `node.uses` contains everything, though, so what `nested_in` does in this case is filter out uses in parent and sibling scopes. I'm adding comments in the code to clarify.

Incidentally, this is also wrong, because this code should only be triggered for graphs that are *immediately* nested in the current graph (because they are already taking care of any graphs that are nested further). I'll have to fix that. And then test it, definitely test it...

bartvm: I repeated this many times now, but it seems to me that there are two different ways of what you call "nesting". It would probably help a lot if you tried to understand and acknowledge the disambiguation I am trying to make.

One is represented in Myia through direct edges in the graph, the other one is not. One can have what you call "cyclical nesting", the other one does not. I am not attached to the precise terminology, as long as they are disambiguated, but I have been referring to the former as "lexical (static) nesting", because from the graph you can directly see what the nesting is due to edges between functions. Dependencies due to returning closures are different: As you mentioned yourself, you can easily construct graphs that will try to read from an uninitialized environment and crash. Hence I have been calling these "runtime dependencies".

In the example I gave early on, `h` is lexically scoped in `f` and `g`, `f` and `g` have a cyclical runtime dependency. In the example you just gave: `idx` is a reference to `f`'s body, so`g2` is lexically nested in `f`. On the other hand, `g3` needs access to `f`'s environment in order to construct a closure for `g2`, so there is a runtime dependency here. Similarly, `g4`  has a runtime dependency on `f` through `g3`.

> This has to return g3. By your criteria, g3 isn't nested in f, so we're just returning g3, right?

No. I never discussed closure creation semantics. All I said is that `g3` isn't lexically scoped in the sense that it has no references to `f`'s body. Instead, it has a runtime dependency on `f` through `g2` (which is lexically nested in `f`).

Maybe I didn't clarify this enough in https://github.com/mila-udem/myia/pull/22#discussion_r164237322. Wherever I said "call stack" I meant runtime as opposed to lexical i.e. `g3` accesses `f` not lexically, but because at runtime it is asked to construct a closure, which in turn is lexically nested.

> It needs to be defined because the return value of the function depends on it, and if the return value depends on it, so does the AD, obviously. 

Not obvious to me. Just to clarify: What I mean is that runtime dependencies don't need to be known in order to apply the AD transformation. In my previous comment I gave an example of a backpropagator for `f` that does not rely on knowing the cyclical runtime dependency between `f` and `g`. This is a similar working example for your case. The only I need to know is that `idx` is a closed over variable, so that we extract it from the environment sensitivities. Otherwise you can just blindly pass through the environment sensitivities of `g2` and `g3`, without needing to know who is runtime dependent on who ("nested").

```python
class Env(dict):
    """Environment sensitivities can be combined."""
    def __add__(self, other):
        added = Env()
        for k in self.keys() | other.keys():
            null = type(self.get(k, other.get(k)))()
            added[k] = self.get(k, null) + other.get(k, null)
        return added

# Environment sensitivities default to 0
def f(idx, g1):
    def g2():
        def dg2(didx):
            return Env(idx=didx)
        return idx, dg2
    def g3():
        def dg3(dg2):
            return dg2
        return g2, dg3
    def g4():
        def dg4(dg3):
            return dg3
        return g3, dg4
    gs = [g1, g2, g3, g4]
    g = gs[idx]
    y, dg = g()
    def df(dy):
        # We get the environment sensitivities of the closure g
        denv = dg(dy)
        # The adjoint of indexing
        dgs = [Env()] * 4; dgs[idx] += denv
        # The adjoint of constructing a list
        dg1, dg2, dg3, dg4 = dgs
        # The environment sensitivity of f is the combined
        # sensitivities of all the constructed closures
        denv = dg2 + dg3 + dg4
        # idx is a closed over variable, so we extract that
        didx = denv.pop('idx', 0)
        # Return correct partials
        return (didx, dg1), denv
    return y, df

# Forward pass
g3, df1 = f(3, None)
g2, df2 = f(0, g3)
idx, df3 = f(0, g2)

# Backward pass
(_, dg2), denv = df3(1)
(_, dg3), denv = df2(dg2)
(didx, _), denv = df1(dg3)
assert didx == 1
```

So to summarize: There are two kinds of nesting: lexical scope and runtime dependencies (I don't care much for the terminology). For the AD transform, I can see that you need to know the lexical scope. I don't see why you need to know the runtime dependencies. I worked out two examples of AD transforms that don't seem to rely on this. Am I overlooking something?

breuleux: > I repeated this many times now, but it seems to me that there are two different ways of what you call "nesting". It would probably help a lot if you tried to understand and acknowledge the disambiguation I am trying to make.

I'm trying, but it's deep in a sea of red herrings.

> One can have what you call "cyclical nesting", the other one does not.

There is no possible Python program that will produce a graph that exhibits either of these cycles. And if you're looking at all graphs that can be constructed using our IR, both "kinds" of nesting can in fact be cyclical. Here's an explicit one, using made up syntax:

```python
def f(x): return g(x + g::y)
def g(y): return f(f::x + y)
```

Where is it written that we aren't supposed to compile this? Nowhere. We just assume we're not going to, just like I assume that we want to ban implicit cycles on the grounds of having a principled way to prove that we will not access undefined variables (not to mention the lack of syntax and use cases). I don't know why we've even been entertaining this scenario, to be honest.

---

Anyway, what you mean I think is that it is possible to implement AD using a opaque universal representation for environments. Then a function that does not explicitly read a variable can sort of just chuck the whole pack of them around without looking in it, forward or backward. Sure. Notwithstanding the fact that this is a question of representation and has little to do with the idea of "nesting", there are a couple issues with the approach:

* It is a more complex data structure.
* Seems like every reference to a top-level function will produce a spurious addition operation of an empty environment. To eliminate these, you will need an analysis akin to what I wrote.

bartvm: Just a quick summary of what we discussed offline earlier today, and then let me add some stuff that I thought of later.

We agreed that there are two approaches. The example we worked out was for:
```python
def f(x, y):
    def g():
        def h():
            return x
        def i():
            return y + x
        return h if ... else i
    c = g()
    return c
```
The AD transform in my head was this:
```python
def f(x, y):
    def g():
        def h():
            def dh(dx):
                return Env(x=dx)
            return x, dh
        def i():
            def di(dz):
                return Env(x=dz, y=dz)
            return y + x, di
        def dg(denv):
            # denv is either Env(x=dx) or Env(x=dx, y=dy)
            # g has no free variables -> just pass through
            return denv
        return h if ... else i, dg
    c = g()
    def df(denv):
        denv = dg(denv)  # denv is either Env(x=dx) or Env(x=dx, y=dy)
        dx = denv.pop('x', 0)
        dy = denv.pop('y', 0)
        return (dx, dy), denv  # empty environment
    return c, df
```
The example that @breuleux worked out was:
```python
def f(x, y):
    def g():
        def h():
            def dh(dx):
                return Env(x=dx)
            return x, dh
        def i():
            def di(dz):
                return Env(x=dz, y=dz)
            return y + x, di
        def dg(denv):
            # denv is either Env(x=dx) or Env(x=dx, y=dy)
            # we look at what variables were captured by the returned values
            dh, di = Env(x=0), Env(x=0, y=0)  # requires knowing that h and i access x and y
            if bla:
                dh += denv
            else:
                di += denv
            # We know the structure of the environment sensitivities here, so we can unpack
            dx, = dh
            dx, dy = di
            # construct new environment, need to know all variables that can be captured by g
            return Env(x=dx, y=dy)
        return h if bla else i, dg
    c = g()
    def df(denv):
        denv = dg(denv)
        dx, dy = denv  # we know the structure of denv returned by dg
        return (dx, dy), Env()
    return c, df
```
However, thinking about it on the way home, I'm having some problems with the example @breuleux worked out.  Notably, `g` might receive a function as argument which is then passed on to be called somewhere else. This means that we still have no idea idea what the argument `denv` to `dg` looks like. An example of such a case:
```python
def f(x, y, z):
    def g():
        return y + z
    def h(i):
        # h's backpropagator either receives Env(x=dx, y=dy) or Env(y=dy, z=dz)
        # (or Env(), actually, if the returned closure was never actually called)
        # figuring out that z might be in the environment is not trivial at all
        def j():
            return x + y
        return i if ... else j
    k = h(g)
    return k
```
I can work out the backpropagators for this my way pretty easily (see below) but I don't readily see how you write a backpropagator for `h` that statically unpacks or constructs environment sensitivities, because you don't know what it looks like. The only thing you can do is just "chuck the whole pack of them around without looking in it".
```python
def dh(denv):
    return denv, Env() if ... else Env(), denv
```

breuleux: `dh` doesn't return a sensitivity for `z` at all. That sensitivity will be returned in the *parameter* sensitivity, not the environment sensitivity, and it will be extracted by `f`, not by `h`. Basically:

```python
def bprop_h(dout):
    di, dj, dx, dy = 0, 0, 0, 0
    if ...:
        di += dout
    else:
        dj += dout
    dx, dy += dj
    return (dx, dy,), di

def bprop_f(dk):
    dh, dg, dx, dy, dz = 0, 0, 0, 0, 0
    dh, dg += bprop_h(dk)
    dx, dy += dh
    dy, dz += dg  # We get dz here, from *dg*
    return (), dx, dy, dz
```

Basically, sensitivities that come from parameters don't need to be unpacked, they are chucked along like in your method.

bartvm: But `di` is initialized to `Env()` (you can't do anything more clever) and `dout` is `Env(y=dy, z=z)`, so here you aren't just doing an element-wise addition on two tuples, you are doing the more complicated "combine two arbitrary environments", right?

If we do need a more complex environment object that can be added/merged in this way, then what is the advantage of your `bprop_h` example compared to my `dh`?

breuleux: `di` is initialized as `zeros_like(i)` (which `bprop_h` plucks from the scope of `h`). That operation essentially creates a zeroed out copy of whatever `i`'s closure is, if we can assume that the gradient tuple has the same structure as the closure environment (which would be sensible) -- otherwise I suppose it would have to look up the template in some runtime table. Either way, there is no need for a complex environment object.

bartvm: Okay, that makes more sense than `di = 0`. Keeping references to closures around doesn't seem ideal though, you will want to free those sooner rather than later.

And to clarify, `di` is not a tuple, right? It's got a variable number of elements after all, all with different types. So although it's not as complex as an unordered dict that must be merged, it's still a special kind of object that has a different type. This means we need a special `add_envs` primitive, which does element-wise addition on these types of objects, to perform `di += dout`.

So I understand now that you do not need an arbitrary read-write object; we can guarantee that two environments will have the same elements when they are added. On the other hand, in general we do not know the structure of environment sensitivities, so we can't statically unpack them outside of the function that constructed the closure, and we can't combine environment sensitivities using the standard `add` operator, which is what I guess I've been trying to get at.

breuleux: > Keeping references to closures around doesn't seem ideal though, you will want to free those sooner rather than later.

It's arguably a problem with both approaches. When you do something like `denv.pop('x', 0)`, what you really want is `denv.pop('x', zeros_like(x))`, so you need to keep a reference to `x` and that may very well be 99% of the closure right there.

I don't expect most of these cases to remain after optimization -- even then, there's the simple trick of pulling `zeros_like` out of scope, so that the free variable is the zeroed-out structure rather than the closure. That would take up less memory since you can use zero strides for tensors and whatnot. Or something like `zeros(abstract_shape(i))`.

> And to clarify, `di` is not a tuple, right? It's got a variable number of elements after all, all with different types.

You mean it'll have a different type depending on the caller? That's true, that's part of the annoyance of typing grad. I wouldn't say it's necessarily "not a tuple", though -- it can be a union of tuples (`(int, str)|(str, int, int)|...`). Coupled with the fact that we only unpack when we know the function and can narrow down the right type, and maybe something like type variables, it could work.

This is a rough sketch and I'd have to think more to see if it'd work, but for every variable `v` that has a function type, you could associate a special type variable `v::env`. So if you have a call `z = f(x, y)` the backpropagator call `bprop_f(dz)` would return something with type `(f::env, x::env, y::env)`. These would not be types per se, they would be type variables: if you know the exact identity of `f`, then you can unify `f::env` with a specific tuple type and that lets you unpack it; if `f` could be either `g` or `h` then `f::env = Union[g::env, h::env]` (and you can't unpack it safely). `add` would just require that its two inputs are unifiable and would map to an implementation of structural map that does runtime type dispatch (I think it's more like an unspecialized add than a different primitive altogether). I think what grad generates is mostly straightforward and doesn't do anything particularly weird, so I imagine this would probably suffice? Something similar would have to be done for the env approach, I reckon, if you want the type system to be able to prove that the element you're popping has the right type.

bartvm: > it can be a union of tuples ((int, str)|(str, int, int)|...)

Determining what this type is could require finding out all the possible values of an argument though. That seems like it could get complicated.

I was wondering if there is a way of moving the adding of partials to the backpropagator of the function where the original variable was defined. At that point of the code we know the type after all, and in a way it seems more symmetric. The major downside though is that it would cost a lot of memory keeping all those partials around separately longer than needed... I guess that's the reason that the symmetry between primal and adjoint goes out the window in this case, which means that we are operating on data not local to the adjoint, and that gives these typing issues.

Your suggestion for typing seems reasonable. Alternatively, the environment sensitivities could just be typed at runtime for runtime dispatch, but then you'd have to runtime compile the addition operators as well, so not a great idea.

bartvm: To get back to the original discussion by the way, because we digressed a lot: Ideally I think the two types of nesting (free variables, closure construction) should be separate passes. In some cases I might just want to know that `h` accesses `f`'s scope, instead of getting the nesting `h->g->f` because `g` constructs `h`. The relationships from both these passes can then be combined into the nesting relationship you want by doing a topological sort on the combined dependencies (which should replace the `simplify` pass).

breuleux: They already are separate passes in `NestingAnalyzer`: `compute_deps` fills `deps` with direct dependencies, plus a few edges toward `ParentProxy` objects (which you can just ignore in this case). `resolve_proxies` then adds the implicit dependencies, although it overrides `deps` in doing so. It'd be straightforward to modify `NestingAnalyzer` to provide both analyses.

I don't know if a topological sort is any better than the method `simplify` uses -- by optimizing `simplify` a little it essentially has the same complexity as a single topological sort over the data.

bartvm: What I mean is to expose them as separate passes, preferably as separate classes and/or functions if possible. Likewise, the `simplify` might or might not be doing topological sort, but it would help if it said that, and even better if it would be a reusable implementation of toposort.

breuleux:
Progress report.

I've implemented backpropagators for a few primitives and `if`, ported the finite difference code from `master`, and added a `zeros_like` primitive. Then I ported several tests from `master`.

The good news is that the algorithm seems to be correct: correct gradients are obtained on simple arithmetic, if statements, closures and nested while loops. Worked on first try, too. There are, however, a few issues:

* Using the current VM, `zeros_like` can't get the shape of a closure's gradient, because that involves fetching all closure variables, and some of them will cause `Continue` exceptions to be raised. I used a simple workaround: `zeros_like(closure)` generates the `ZERO` object, which has `__add__` and `__radd__` overriden to return the other argument, and `__getitem__` to return `ZERO`. That seems to do the trick, but I used the same trick in `master` (though more as an optimization) and I recall it causing a few issues with higher order gradients. So I don't trust that solution very much. Might be worth looking into the `Env` idea.

* Tuples are made via the `make_tuple` primitive, which takes a variable number of arguments. This means it's impossible to build a graph for `J(make_tuple)` (so I can't test second order right now). I can see the following solutions:
  * Add varargs support. An inference pass could try to eliminate them for optimization.
  * Special case `make_tuple` in the gradient pass.
  * Generate `J(fn, n)` where `n` is the number of arguments passed to the function. This would let `J` specialize on the number of arguments, in which case a graph could be generated for `make_tuple` with `n` arguments.
  * Replace `make_tuple(a, b, c)` by the composition of a binary function, for example `cons(a, cons(b, cons(c, ())))`.

breuleux:
For now I'm trying the last option: replacing `make_tuple` by the composition of `cons_tuple` calls. Seems to be working so far, just need to scaffold the tests for second order gradients to make sure everything's in order.

tests/test_grad.py
```diff
@@ -0,0 +1,196 @@
...
+
+        def x4(c):
+            return b
+        return x4
+    x2 = x1(a)
+    x3 = x2(1)
+    return x3
+
+
+# TODO: test when a == b (finite diff won't return anything sensible)
```
lamblin: We could have an option for finite diff to only test the left slope or the right slope, and pass the one that will return the right thing.

myia/parser.py
```diff
@@ -208,12 +210,15 @@ def make_location(self, node: ast.AST) -> Location:
             # we basically just pass through them.
             return None  # pragma: no cover
 
-    def parse(self) -> Graph:
+    def parse(self, clean=True) -> Graph:
```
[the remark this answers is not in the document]
breuleux: I can, if I decide to keep it. This branch is in a rough WIP state and this is a bit of a hack, so it might not remain. But basically it determines whether dead code elimination is triggered or not, and I left an option not to trigger it so that DCE could be tested.

myia/grad.py
```diff
@@ -0,0 +1,257 @@
...
+        self.backpropagator_graphs: Dict[Graph, Graph] = {}
+
+        # x -> ↑x
+        self.tagged_nodes: Dict[ANFNode, ANFNode] = {}
+        # x -> ♢x
+        self.backpropagator_nodes: Dict[ANFNode, ANFNode] = {}
+        # (x, g) -> ∇x (in the context of that graph)
+        self.sensitivity_nodes: Dict[Tuple[ANFNode, Graph], ANFNode] = {}
+        # x -> ♢x(∇x) (for x an Apply node)
+        self.step_nodes: Dict[Apply, Apply] = {}
```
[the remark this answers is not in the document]
breuleux: P&S uses a bunch of arrows and semi-arrows over the symbols, which I'm not sure how to represent here. But yes, that would be good to document.

myia/grad.py
```diff
@@ -0,0 +1,257 @@
...
+        bprop = self.backpropagator_nodes[node]
+        if bprop:
+            rval = self.apply(bg, bprop, self.rho(node, node.graph))
+            self.step_nodes[node] = rval
+            rval.debug.about = About(node, 'grad_bprop_step')
+            return rval
+        else:
+            return None
+
+    def rho(self, node, graph):
```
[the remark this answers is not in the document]
breuleux: They call the forward and backward transforms phi and rho in the paper. I'm not sure why, and I agree it's a bit obtuse. I'll probably change it to `make_sensitivity` or something like that.

myia/py_implementations.py
```diff
@@ -159,3 +162,77 @@ def setattr(data, attr, value):
...
+    elif isinstance(x, (int, float)):
+        return x
+    elif isinstance(x, tuple):
+        return smap(J, x)
+    else:
+        raise TypeError(f'J is not defined on {type(x)}')
+
+
+@register(primops.Jinv)
+def Jinv(x):
```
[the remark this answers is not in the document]
breuleux: It's the inverse of `J`, `J^{-1}`. When applied on a function, it undoes the transform. This is used in the gradient of `J`.

tests/test_grad.py
breuleux: Right. I think the best option, which Bart mentioned he used for Tangent, would be to compare with autograd. It was simpler for me to just use the finite diff code I already had, but ultimately I think we should compare with autograd or tangent.

tests/test_grad.py
```diff
@@ -0,0 +1,196 @@
...
+def test_while(x, y, z):
+    rval = 0
+    # Cannot compare to 0 or finite diff is unstable
+    while x > -0.1:
+        rval += y
+        x -= z
+    return rval
+
+
+# @grad_test((2,), (4,), (8,))
```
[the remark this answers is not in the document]
breuleux: `for` is not supported by the parser and there is no `range` primitive. The test was in master, though, so I pasted it anyway and commented it out for later.

myia/parser.py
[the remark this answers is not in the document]
lamblin: OK.

myia/grad.py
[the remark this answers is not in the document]
lamblin: I think even LaTeX embedded and rendered in a sphinx / autodoc page would be nice.

myia/grad.py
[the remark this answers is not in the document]
lamblin: OK, I did not remember that.

tests/test_grad.py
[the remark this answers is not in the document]
lamblin: OK, then I think we need a more formal roadmap on what needs to be ported from master, since I did not realize `for` had not been ported in the "clean" parser yet..

myia/grad.py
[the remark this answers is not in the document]
breuleux: That looks pretty awful if you're just looking at the source code though.

myia/py_implementations.py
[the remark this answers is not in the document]
breuleux: Yes. It applies recursively on data structures and is a no-op on scalars.

myia/grad.py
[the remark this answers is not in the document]
lamblin: I mean, we could have a webpage where this is rendered somewhere (be it through embedded LaTeX or pre-rendered images or whatever), and have a link to that page here as a glossary between their notations and Myia's.

breuleux:
The functionality should be complete, and the tests pass, but there are still a few things I want to improve, like notation and documentation.

Note that this PR includes the modifications in #35, #45 and #46. You should review these before this one.

breuleux:
I'm closing this PR and making a new one.

PR #24, opened 2018-01-25
DebugInfo framework

breuleux:
This reorganizes the `Debug` objects in `anf_ir` in a new file, `myia.info`, and the `DebugInfo` and `AutoNamedDebugInfo` classes. `DebugInfo` is a context manager and behaves essentially like this:

```python
with DebugInfo(a=1):
    with DebugInfo(b=2):
        info = DebugInfo(c=3)
assert info.a == 1
assert info.b == 2
assert info.c == 3
```

The idea is that a block of code can be executed in a context where certain properties are set, and every node or graph created within that code will inherit these properties. This lets us keep some information without having to pass it around or setting it explicitly. It can also be used as a kind of dynamically scoped store, I suppose.

`DebugInfo` defines the optional attributes `about: DebugInfo` and `relation: str`. For example, the graph for the true branch of the first `if` in function `f` will have `about = f.graph` and `relation = 'if_true'`. This can be used to derive a more informative name for this graph, like `f.if_true`, `✓f`, or something else. I modified `Parser` to set these fields.

It also saves the trace of the creation of the object in `trace` if `save_trace == True`. That was useful to me in the past, so I just threw it in there.

### DIFF
--- a/myia/anf_ir.py
+++ b/myia/anf_ir.py
@@ -14,7 +14,7 @@ from typing import (List, Set, Tuple, Any, Sequence, MutableSequence,
 
 from myia.ir import Node
 from myia.utils import Named, repr_, list_str
-from myia.info import AutoNamedDebugInfo
+from myia.info import NamedDebugInfo
 from myia import primops
 
 PARAMETER = Named('PARAMETER')
@@ -42,7 +42,7 @@ class Graph:
         """Construct a graph."""
         self.parameters: List[Parameter] = []
         self.return_: Apply = None
-        self.debug = AutoNamedDebugInfo(type=self.__class__.__name__.lower())
+        self.debug = NamedDebugInfo(type=self.__class__.__name__.lower())
 
     @property
     def output(self) -> 'ANFNode':
@@ -106,7 +106,7 @@ class ANFNode(Node):
         self.value = value
         self.graph = graph
         self.uses: Set[Tuple[ANFNode, int]] = set()
-        self.debug = AutoNamedDebugInfo(type='graph')
+        self.debug = NamedDebugInfo(type=self.__class__.__name__.lower())
 
     @property
     def inputs(self) -> 'Inputs':
--- a/myia/anf_ir.py
+++ b/myia/anf_ir.py
@@ -42,7 +42,7 @@ class Graph:
         """Construct a graph."""
         self.parameters: List[Parameter] = []
         self.return_: Apply = None
-        self.debug = NamedDebugInfo(type=self.__class__.__name__.lower())
+        self.debug = NamedDebugInfo(self)
 
     @property
     def output(self) -> 'ANFNode':
@@ -106,7 +106,7 @@ class ANFNode(Node):
         self.value = value
         self.graph = graph
         self.uses: Set[Tuple[ANFNode, int]] = set()
-        self.debug = NamedDebugInfo(type=self.__class__.__name__.lower())
+        self.debug = NamedDebugInfo(self)
 
     @property
     def inputs(self) -> 'Inputs':
--- a/myia/anf_ir.py
+++ b/myia/anf_ir.py
@@ -9,13 +9,12 @@ implicitly creates a nested function. Functions are first-class objects, so
 returning a nested function creates a closure.
 
 """
-import types
-from ast import AST
 from typing import (List, Set, Tuple, Any, Sequence, MutableSequence,
                     overload, Iterable)
 
 from myia.ir import Node
 from myia.utils import Named, repr_, list_str
+from .info import AutoNamedDebugInfo
 from myia.primops import Return
 
 PARAMETER = Named('PARAMETER')
@@ -24,41 +23,6 @@ APPLY = Named('APPLY')
 LITERALS = (bool, int, str, float)
 
 RETURN = Return()
-
-
-class Debug(types.SimpleNamespace):
-    """Debug information for an object.
-
-    Attributes:
-        name: The name of the object.
-
-    """
-
-    _curr_id = 0
-
-    def __init__(self, obj, **kwargs):
-        """Construct a Debug object."""
-        self.obj: Any = obj
-        self.name: str = None
-        self._id: int = None
-        super().__init__(**kwargs)
-
-    @property
-    def id(self):
-        """Generate a unique, sequential ID number."""
-        if self._id is None:
-            self._id = self._curr_id
-            self.__class__._curr_id += 1
-        return self._id
-
-    @property
-    def debug_name(self):
-        """Return the name, create a fresh name if needed."""
-        if self.name:
-            return self.name
-        prefix = self.obj.__class__.__name__.lower()
-        self.name = f'_{prefix}{self.id}'
-        return self.name
 
 
 class Graph:
@@ -80,7 +44,7 @@ class Graph:
         """Construct a graph."""
         self.parameters: List[Parameter] = []
         self.return_: Apply = None
-        self.debug = GraphDebug(self)
+        self.debug = AutoNamedDebugInfo(type=self.__class__.__name__.lower())
 
     @property
     def output(self) -> 'ANFNode':
@@ -111,21 +75,6 @@ class Graph:
         return repr_(self, name=self.debug.debug_name,
                      parameters=list_str(self.parameters),
                      return_=self.return_)
-
-
-class GraphDebug(Debug):
-    """Debug information for a graph.
-
-    Any information that is used for debugging e.g. plotting, printing, etc.
-    can be stored in this class.
-
-    Attributes:
-        ast: The AST node that created this graph.
-        name: The function name.
-
-    """
-
-    ast: AST
 
 
 class ANFNode(Node):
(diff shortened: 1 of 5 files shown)


myia/anf_ir.py
```diff
@@ -157,7 +106,7 @@ def __init__(self, inputs: Iterable['ANFNode'], value: Any,
         self.value = value
         self.graph = graph
         self.uses: Set[Tuple[ANFNode, int]] = set()
-        self.debug = ANFNodeDebug(self)
+        self.debug = AutoNamedDebugInfo(type='graph')
```
bartvm: `'graph'`? Also, it's hardcoded here whereas it's dynamic for `Graph`

myia/info.py
```diff
@@ -0,0 +1,103 @@
...
+    def __init__(self, **kwargs):
+        """Construct a DebugInfo object."""
+        self.about = None
+        self.relation = None
+        self.save_trace: bool = False
+        self.trace: Any = None
+        top = current_info()
+        if top:
+            # Only need to look at the top of the stack
+            self.__dict__.update(top.__dict__)
```
bartvm: So you're copying `about` and `relation` as well? Does that mean that two objects can have the same relationship to the same object if `about` and `relation` isn't overwritten elsewhere?

myia/info.py
```diff
@@ -0,0 +1,103 @@
...
+
+        self._id: int = None
+        self.__dict__.update(kwargs)
```
bartvm: `super().__init__(**kwargs)`? Otherwise not much point in inheriting from `SimpleNamespace`

myia/info.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        _about.stack.append(self)
+
+    def __exit__(self, type, value, tb):
+        """Exit the context of this `DebugInfo`."""
+        assert _about.stack[-1] is self
+        _about.stack.pop()
+
+
+class AutoNamedDebugInfo(DebugInfo):
```
bartvm: `NamedDebugInfo`? They can be manually named after all, and it's shorter.

myia/info.py
```diff
@@ -0,0 +1,103 @@
...
+        """Return the name, create a fresh name if needed."""
+        if self.name:
+            return self.name
+        prefix = self.type or ''
+        self.name = f'_{prefix}{self.id}'
+        return self.name
+
+
+def about(obj, relation=None):
+    """Set context as being about the given object."""
```
bartvm: A slightly longer docstring might be good here, it's very abstract right now. Maybe just give an example?

myia/anf_ir.py
breuleux: Yeah, that's because I wanted to preserve the same string representation you had, but I didn't want `DebugInfo` to hold a pointer to the object (because holding a `DebugInfo` shouldn't keep the object alive, especially if they're going to be chained in the `about` field), so I just used a string instead.

This being said, thinking about it, the object could be held in a weak reference. I think that would be perfect.

myia/info.py
breuleux: Potentially, yes. If a node is replaced by two nodes, you can say that they both have the "replaces" relation with respect to the original. That's coarse information, but better than none.

myia/info.py
breuleux: I think that's basically what `SimpleNamespace.__init__` does. Just wanted to make sure that we're updating `__dict__` and not doing some other weird thing. I guess that defeats the point somewhat yeah.

myia/info.py
breuleux: Sure.

myia/info.py
breuleux: Will do.

myia/anf_ir.py
breuleux: Yeah I just realized what you meant. That's obviously the wrong type, oops.

myia/parser.py
```diff
@@ -42,9 +42,31 @@
...
+        line: The line number.
+        column: The column number.
+
+    """
+
+    def __init__(self, url, line, column):
+        """Initialize a Location."""
+        self.url = url
+        self.line = line
+        self.column = column
```
bartvm: Since this is basically just a struct, maybe a good idea to make it immutable (and leave the pretty string printing to e.g. the logger).
```python
class Location(NamedTuple):
    url: str
    line: str
    column: int
```

myia/parser.py
```diff
@@ -42,9 +42,31 @@
...
+    Attributes:
+        url: The filename.
+        line: The line number.
+        column: The column number.
+
+    """
+
+    def __init__(self, url, line, column):
+        """Initialize a Location."""
+        self.url = url
```
bartvm: It's a file, not a URL

myia/parser.py
```diff
@@ -138,6 +162,15 @@ def __init__(self, environment: Environment,
             globals_.update(free_vars.nonlocals)
         self.globals_ = globals_
 
+    def make_location(self, node: ast.AST) -> Location:
+        """Create a Location from an AST node."""
+        if hasattr(node, 'lineno') and hasattr(node, 'col_offset'):
+            return Location(self.url,
+                            node.lineno + self.line_offset - 1,  # type: ignore
+                            node.col_offset)  # type: ignore
+        else:
```
bartvm: Can we fail early here instead of returning `None`?

myia/parser.py
breuleux: That's a good idea.

myia/parser.py
breuleux: True. I intended to have something more general, but I didn't really bother making it an actual url, as you can see.

myia/parser.py
breuleux: OK.

myia/info.py
bartvm: What I meant more is:
```python
with about(about=x, relation='child'):
    with about(about=y):
        n = Constant(2)  # this node is now 'child' of x?
```

myia/info.py
```diff
@@ -19,6 +19,15 @@ def current_info():
 class DebugInfo(types.SimpleNamespace):
     """Debug information for an object.
 
+    If used with the `with` statement, any `DebugInfo` created
+    while inside the `with` body will inherit all attributes:
+
+        with DebugInfo(a=1, b=2):
```
bartvm: Code snippets in docstrings should probably use `>>>` and `...` so that they are compatible with e.g. [`doctest`](https://docs.python.org/3.6/library/doctest.html) and Sphinx.

myia/info.py
breuleux: In this case, no, `relation` will be `None` (the default value for that argument in `about`). Would it be better to require that argument?

`with DebugInfo(about=y)` would have the behavior you describe, though.

myia/info.py
breuleux: Ah, right.

myia/info.py
bartvm: It's a bit opaque that `DebugInfo` inherits everything from its parent, except if constructed through `about`, in which case the attributes `relation` and `about` are required, or overwritten.

Is it obvious that a child should inherit everything from its parent? For example, using `with DebugInfo(about=y)` copying the relationship of its parent while the `about` field is different, is confusing. Setting `_id` to `None` seems to be another example where this inheriting behavior has to be manually overridden to prevent weird behavior, and in doing so it actually leaks implementation details from the `NamedDebugInfo` subclass.

Is it possible to not inherit everything, but instead just set a `parent` attribute? Or maybe we can have a subclass of `DebugInfo` with a list of attributes that should be inherited (e.g. `location` and `ast`, I imagine).

breuleux: The idea is that if you write `with DebugInfo(...)` you want every `DebugInfo` created in that context to have properties `...`. That's the whole point of doing it. Going through parents would be an annoyance. A whitelist wouldn't work, since you may want to make up new debug properties to help you debug something e.g. `with DebugInfo(my_transform=parameter): my_transform(parameter, graph)`.

A cleaner implementation, though, might be to use a different class than `DebugInfo` as a context manager: then it would never have any fields you don't mean to inherit.

bartvm: That might be a good idea. It seems weird that any assigment to a debug object propagates to children. Only the ones given in the context manager should propagate.

The `about` and `relationship` are an odd one out, but maybe they should just be one field i.e. `about = (x, 'child')`. That way it either fully propagates, or not at all, but you can't have the thing where a relationship suddenly applies to a new object. In that case, maybe you can also get rid of the `about` constructor and just use `DebugInfo` directly?

breuleux:
Is there anything else to address before merging this?

myia/parser.py
```diff
@@ -129,6 +146,8 @@ def __init__(self, environment: Environment,
         """Construct a parser."""
         self.environment = environment
         self.function = function
+        _, self.line_offset = inspect.getsourcelines(function)
+        self.url: str = inspect.getfile(function)
```
bartvm: url -> filename

bartvm:
LGTM (just url -> filename was missed in one place)

PR #25, opened 2018-01-30
Make Primitive subclass of Named, add primitives

breuleux:
Primitives are now individual instances of the Primitive class, a subclass of Named. A whole bunch of them were also added, with implementations pending.


abergeron:
I'm ok with this and I need it for the type inference work, so I'll merge this as-is and we can refine later if we find some problem.

PR #26, opened 2018-01-31
Primitive implementations for the VM

breuleux:
This modifies VM to take a dictionary of implementations for primitives. A set of default implementations is provided in `myia.impl_py` for all the primitives we have at the moment.

Tests remain to be added. More language support will be added to `parser` as needed to test the primitives (whenever it is easy to do so).


myia/impl_py.py
```diff
@@ -0,0 +1,160 @@
+"""Implementations for the debug VM."""
+
+
+from typing import Dict, Callable
+from copy import copy
+from myia import primops
+
+
+implementations: Dict[primops.Primitive, Callable] = {}
```
bartvm: Globals should be in capitals.

myia/impl_py.py
```diff
@@ -0,0 +1,160 @@
...
+    """Create a decorator to associate a function to a primitive.
+
+    The function name will need to start with `pfx`, followed by
+    the primitive's name in `primops`. It will then be associated
+    to that Primitive in `store`.
+    """
+    def deco(fn):
+        """Decorate the function."""
+        name = fn.__name__
+        assert name.startswith(pfx)
```
bartvm: No `assert`, but `raise ValueError`

myia/impl_py.py
```diff
@@ -0,0 +1,160 @@
...
+        name = name.split(pfx)[1]
+        prim = getattr(primops, name)
+        store[prim] = fn
+        return fn
+    return deco
+
+
+impl = registrar('impl_', implementations)
```
bartvm: `impl` isn't really a universal abbreviation, so maybe just write out `implementation`

myia/impl_py.py
```diff
@@ -0,0 +1,160 @@
...
+
+from typing import Dict, Callable
+from copy import copy
+from myia import primops
+
+
+implementations: Dict[primops.Primitive, Callable] = {}
+
+
+def registrar(pfx, store):
```
bartvm: Should we insist on the prefix? It should be clear that the `add` function is an implementation because it's in the `impl_py` module, so we shouldn't have to enforce the `impl_` prefix everywhere.

myia/impl_py.py
```diff
@@ -0,0 +1,160 @@
+"""Implementations for the debug VM."""
+
+
+from typing import Dict, Callable
+from copy import copy
```
bartvm: If you really want to enforce immutability, this might have to be `deepcopy` instead of `copy`?

myia/impl_py.py
```diff
@@ -0,0 +1,160 @@
...
+from myia import primops
```
bartvm: PEP8; imports should be grouped standard, third party, local, each with a blank line in between.

myia/impl_py.py
breuleux: I think that's for constants, not dictionaries?

myia/impl_py.py
breuleux: `@impl` gets the primitive name from the function name by assuming it has the form `impl_{name}`, so checking the prefix is a sanity check of sorts.

Also at some point I was pondering having a second type of implementation, which would take `frame` as an extra argument and could implement primitives like `if`. That would be a different prefix in the same file.

myia/impl_py.py
breuleux: This is for the implementation of `setitem` and `setattr`, which only modify the object shallowly.

myia/impl_py.py
bartvm: Not sure I follow what "sanity check" enforcing `impl_{name}` instead of `{name}` is. It just checks whether I named a function a certain way.

If you have multiple things for a single primitive in the same file, then sure, although I thought the point of this exercise was to separate implementations of certain aspects into different modules after all. That said, can we at least have something readable like `py_add` or `add_implementation`? `impl_add`, and `impl_py` for that matter, aren't very readable names.

myia/impl_py.py
bartvm: Well, that depends on the object. Which ones are you trying to support here? `setitem` is for lists and arrays I guess, but what is `setattr` for?

breuleux: `setattr` would be for records, but I suppose that may be a bit premature.

myia/impl_py.py
breuleux: Getting the primitive name from the function name saves some typing and prevents some copy-paste errors. Adding a prefix is useful since otherwise the implementation of certain primitives that happen to have the same name as a builtin will shadow that builtin, so it needs to be aliased to something else. On the other hand, I suppose it would be nice if you could import `add` from `myia.impl`, use it in a Myia-decorated function, and have it resolve to the primitive.

bartvm: On the other hand, `@implementation(primops.add)` is much more explicit than `@impl` plus magic string matching, which forces me to name my function `impl_bla`. It also enables static checking e.g. `impl_Named` would happily parse right now, but if you type `implementation` you could get a static type error.

Mostly pushing back on this because although this isn't too bad, I think that libraries which add magic in order to save a few keystrokes quickly end up being hard to maintain and difficult to understand for outsiders. Without reading the implementation of `registrar` someone would be surprised to find out that things crash when they change function names, or end up wondering how to register the implementation of their new primitive. All of these small bits of magic quickly introduce significant startup costs to new developers.

breuleux: I don't particularly care about this, so I'll just make it explicit.

I think you are exaggerating the issue, though. The point of this kind of magic isn't *just* to save typing, but also to prevent easy copy/paste errors that would see the primitive changed but not the function name. That has happened to me multiple times, which caused a few really bizarre errors, but that's mostly an issue with rapid development cycles. I agree that the magic needs not be user-facing, though (I was thinking of this mostly for internal use, to be honest).

myia/impl_py.py
breuleux: It's a pretty common abbreviation, I think. What else would it mean? Implication?

bartvm: Implicit, for example.

myia/impl_py.py
breuleux: Done.

breuleux:
Quick mention that I reorganized the tests:

* Tests in `test_parser` that were checking valid functions were moved to `test_lang` and now also test that the functions return the right value on one or more inputs.
* Tests in `test_parser_vm` were also moved to `test_lang`.
* Several tests from `master::test_front` were also ported and moved to `test_lang`, since the primitives they depend on are now available.
* Redundant tests were eliminated.
* Each primitive is tested in `test_prim`.

abergeron:
I'll leave some time for @bartvm to do another pass if he wants before merging.

myia/api.py
```diff
@@ -50,10 +51,14 @@ def default_object_map() -> Dict[Any, ANFNode]:
         operator.ge: Constant(P.ge),
         operator.not_: Constant(P.not_)
     }
+    for prim, impl in implementations.items():
```
bartvm: I would prefer not exposing this right now. I think the decision as to whether we want shadow NumPy's API, or have our own API, and if so, what that API should look like, it still an open question. And for now I think all the primops are covered by Python operators, right?

myia/py_implementations.py
```diff
@@ -0,0 +1,144 @@
...
+
+from typing import Dict, Callable
+from copy import copy
+from myia import primops
+
+
+implementations: Dict[primops.Primitive, Callable] = {}
+
+
+def make_registrar(store):
```
bartvm: Maybe move this to `utils`?

bartvm:
LGTM, except that I would prefer keeping the Python implementations as an internal implementation detail, and not expose them to the user.

myia/api.py
breuleux: `getitem/setitem` aren't at the moment, but it's true they are in the `operator` package. I wouldn't really call this "exposing" them, though -- this only makes it so that if we import `myia.py_implementations.add` and use it, it'll resolve to the primitive instead of naively building a graph for it. There is an internal use case for this, which is that if we define some primitive's gradient as a Myia graph, and it needs a primitive that isn't covered by Python's standard library, we can simply get it from there. So while I agree it probably shouldn't be a user-facing API, it's still internally useful.

myia/py_implementations.py
breuleux: Sure. I think I might make this a class, actually.

PR #27, opened 2018-01-31
Type system

abergeron:
This is my work-in-progress code for the type inference.  I made a PR because I realize that it will take a bunch of time to complete.

Don't spend too much time reviewing the details or anything because all of it is subject to change and rebases.

I currently would like comments on the overall structure of the type stuff (in type/dtype.py) and the unification stuff (in type/unity.py).

### DIFF
--- a/myia/dtype.py
+++ b/myia/dtype.py
@@ -75,19 +75,20 @@ class Type(metaclass=TypeMeta):
     """Base class for all Types.
 
     This class brings a number of unusual behaviour for its subclasses
-    compared to normal python classes.
+    compared to normal Python classes.
 
-      * Arguments handled to the constructor will be passed to a
-       `_parse_args` class method if it exists and the value this
-       returns is what will be used for `__init__`.
+      * Arguments passed to the constructor will be passed to the
+       `_parse_args(cls, args, kwargs)` class method if it exists.
+       That method must return a tuple of values, one for each
+       declared class attribute.
 
-      * Instance are unique based on the parsed version of the
-        constructor arguments.  This means that `is` and `==` mean the
+      * Instances are unique based on the parsed version of the
+        constructor arguments. This means that `is` and `==` mean the
         same thing.
 
       * If the `__new__` method is not defined, one will be generated
-        which takes one parameter for each declared class attributes
-        that does not start with an underscore (`_`) and strores them
+        which takes one parameter for each declared class attribute
+        that does not start with an underscore (`_`) and stores them
         in a read-only container that can be referred to by attribute
         access.
 
--- a/myia/dtype.py
+++ b/myia/dtype.py
@@ -72,7 +72,31 @@ class TypeMeta(type):
 
 
 class Type(metaclass=TypeMeta):
-    """Base class for all Types."""
+    """Base class for all Types.
+
+    This class brings a number of unusual behaviour for its subclasses
+    compared to normal python classes.
+
+      * Arguments handled to the constructor will be passed to a
+       `_parse_args` class method if it exists and the value this
+       returns is what will be used for `__init__`.
+
+      * Instance are unique based on the parsed version of the
+        constructor arguments.  This means that `is` and `==` mean the
+        same thing.
+
+      * If the `__new__` method is not defined, one will be generated
+        which takes one parameter for each declared class attributes
+        that does not start with an underscore (`_`) and strores them
+        in a read-only container that can be referred to by attribute
+        access.
+
+    Notes
+    -----
+        In the current implementation, Type is a subtype of tuple via
+        metaclass magic, but this may change in the future.
+
+    """
 
     _root = True
     _fields: DictT[str, Any]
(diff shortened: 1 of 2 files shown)


myia/type/dtype.py
```diff
@@ -0,0 +1,145 @@
...
+    @property
+    def _props(self) -> TupleT:
+        """Return the value of the comparison properties."""
+        return tuple(getattr(self, p) for p in self._props_names)
+
+    def __repr__(self) -> str:
+        """."""
+        name = type(self).__name__
+        args = ', '.join(repr(p) for p in self._props)
+        return "{}({})".format(name, args)
```
breuleux: We target 3.6+ so I'd use `f'{name}({args})'` which I think is more readable.

myia/type/infer.py
```diff
@@ -0,0 +1,141 @@
+"""Type inference."""
```
breuleux: Remove the unification code that's in this file.

myia/type/unity.py
```diff
@@ -0,0 +1,173 @@
+from typing import (Dict as DictT, Tuple as TupleT, Iterable, Union,
```
breuleux: I'd rename the file to `unify` or `unification`.

myia/type/unity.py
```diff
@@ -0,0 +1,173 @@
...
+
+    elif isinstance(w, Callable) and isinstance(v, Callable):
+        if len(w.arguments) != len(v.arguments):
+            raise TypeError("Function with different number of arguments")
+        for we, ve in zip(w.arguments, v.arguments):
+            equiv = unify_raw(we, ve, equiv)
+        return unify_raw(w.retval, v.retval, equiv)
+
+    else:
+        raise TypeError("Type match error")
```
breuleux: I think a custom exception `UnificationError` would be safer to use (since you don't want to catch `TypeError`s that arise from genuine implementation bugs).

myia/type/infer.py
abergeron: Should be done.

myia/type/unity.py
abergeron: I like the name `unity`, so I'll keep it for now.  @bartvm can be the tiebreaker.

abergeron:
I've finally finished the type inference code.

I'm probably missing some docstrings and some tests, but from manual testing all the code should work and you can get the return type for a graph with any set of arguments, ensuring that all the operations along the way support the given types.

This should be good for an in-depth review,  but I will probably reorganize the commits before this is merged.

tests/type/test_unity.py
```diff
@@ -0,0 +1,218 @@
...
+    assert v3.matches(Float(32))
+    assert not v1.matches(v3)
+    assert v3.matches(v1)
+    v4 = RestrictedVar('1', (Float(64), Float(32)))
+    assert v4 is v1
+    v5 = RestrictedVar('1', (Float(32), Float(16)))
+    assert v5 is not v1
+
+
+def test_FitlerVar():
```
lamblin: "Filter" I guess

abergeron:
I've added proper support for function arguments.  They have one limitation: the type of the function must be fixed for a single call.  This means that:

```
def f(fn):
    fn(1, 2)
    fn(1.0, 2.0)
```
would be rejected because `fn(1, 2)` has type `i64, i64 -> i64` and `fn(1.0, 2.0)` has type `f64, f64 -> f64` which are not compatible.

However something like:
```
def f(fn, x, y):
    return fn(x, y)

f(add, 1, 2)
f(add, 1.0, 2.0)
```
is ok, because within a single call to `f`, `fn` has a single type.  (note that in this case the type of `f` is `(a, b -> c), a, b -> c`)

abergeron:
This is now only a representation for types.  Inference is in a follow-up PR.

Arrays are not there yet, will come in another PR.

myia/dtype.py
```diff
@@ -0,0 +1,188 @@
...
+
+class Number(Type):
+    """Numerical values."""
+
+    bits: int
+    _valid_bits: TupleT[int, ...] = ()
+
+    @classmethod
+    def _parse_args(cls, args, kwargs):
+        assert cls is not Number
```
breuleux: A more informative error message would be nice.

myia/dtype.py
```diff
@@ -0,0 +1,188 @@
...
+            assert len(kwargs) == 0
+        key = (self, args)
+        if key not in self._instances:
+            obj = super().__call__(*args)
+            self._instances[key] = obj
+        return self._instances[key]
+
+
+class Type(metaclass=TypeMeta):
+    """Base class for all Types."""
```
breuleux: Could you document a bit how Type works? Just some high level information that's enough to help someone implement a new Type. From what I understand:

* Type instances are cached so that `Type(x, y, z) is Type(x, y, z)`
* Types are tuples.
* `_parse_args(cls, args, kwargs)` is a class method and serves as an init function. It receives constructor arguments (positional and keyword) which it must check. It must return the values of the attributes declared in the class, in the order that they are declared, except those that start with `_`.

myia/dtype.py
```diff
@@ -0,0 +1,188 @@
...
+    @classmethod
+    def _parse_args(cls, args, kwargs):
+        assert cls is not Number
+        assert len(kwargs) == 0
+        assert args[0] in cls._valid_bits
+        return args
+
+
+class Float(Number):
+    """Represents float values."""
```
breuleux: Document how to instantiate a `Float`, e.g. "`Float(nbits)` where `nbits in (16, 32, 64)`". The same applies for all the other types.

myia/dtype.py
```diff
@@ -0,0 +1,188 @@
...
+
+class Tuple(Type):
+    """Represents a set of ordered values with independant types."""
+
+    elements: TupleT[Type, ...]
+
+    @classmethod
+    def _parse_args(cls, args, kwargs):
+        assert len(kwargs) == 0
+        if len(args) == 1 and isinstance(args[0], collections.Iterable):
```
breuleux: Given that types are tuple instances, `Tuple(Float(16))` will trigger this branch and will produce `Tuple((16,))`. That's probably not what someone intends when they write that.

myia/dtype.py
```diff
@@ -0,0 +1,188 @@
...
+        return tuple.__new__(cls, (dict(elements),))  # type: ignore
+
+    def __getattr__(self, attr):
+        if attr in self.elements:
+            return self.elements[attr]
+        raise AttributeError
+
+
+class Tuple(Type):
+    """Represents a set of ordered values with independant types."""
```
breuleux: *independent

myia/dtype.py
```diff
@@ -0,0 +1,188 @@
...
+    @classmethod
+    def _parse_args(cls, args, kwargs):
+        assert len(kwargs) == 0
+        if len(args) == 1 and isinstance(args[0], collections.Iterable):
+            return (tuple(args[0]),)
+        else:
+            return (args,)
+
+
+class Callable(Type):
```
breuleux: `Callable` feels like Python jargon, I'd prefer to call this `Function`.

tests/test_dtype.py
```diff
@@ -0,0 +1,77 @@
...
+
+def test_Tuple():
+    t = Tuple((Float(64), Int(8), Bool()))
+    assert t.elements[1] is Int(8)
+    t1 = Tuple((Bool(), Int(8)))
+    t2 = Tuple([Bool(), Int(8)])
+    assert t1 == t2
+    assert t1 is t2
+    t3 = Tuple(Bool(), Int(8))
+    assert t1 is t3
```
breuleux: Add a test for `Tuple(Float(16)) is Tuple((Float(16),))` (which will fail right now).

myia/dtype.py
abergeron: done

myia/dtype.py
abergeron: fixed

myia/dtype.py
abergeron: done

myia/dtype.py
abergeron: done

myia/dtype.py
abergeron: I'm ok with that.

breuleux:
@bartvm @lamblin I'd like to merge this tomorrow (Wednesday), so tell me if you want to review it before that happens.

PR #30, opened 2018-02-01
Add support for more language features

breuleux:
* Tuples, getitem, getattr
* Tuple unpacking, assign to item or attribute
* Augmented assignment

It's mostly tuple construction that I needed to implement gradients, but I thought while I'm at it I'd add support for a few extra features (all pretty straightforward).

### DIFF
--- a/myia/parser.py
+++ b/myia/parser.py
@@ -41,6 +41,7 @@ Block
 import ast
 import inspect
 import textwrap
+import operator
 from types import FunctionType
 from typing import \
     overload, Any, Dict, List, Optional, Tuple, Type, NamedTuple
@@ -164,7 +165,9 @@ class Parser:
                             node.lineno + self.line_offset - 1,  # type: ignore
                             node.col_offset)  # type: ignore
         else:
-            raise TypeError(f'No location for {node}')  # pragma: no cover
+            # Some nodes like Index carry no location information, but
+            # we basically just pass through them.
+            return None  # pragma: no cover
 
     def parse(self) -> Graph:
         """Parse the function into a Myia graph."""
@@ -229,6 +232,10 @@ class Parser:
         pass
 
     @overload  # noqa
+    def process_node(self, block: 'Block', node: ast.slice) -> ANFNode:
+        pass
+
+    @overload  # noqa
     def process_node(self, block: 'Block', node: ast.stmt) -> 'Block':
         pass
 
@@ -284,6 +291,31 @@ class Parser:
         """Process special constants: `True`, `False`, `None`, etc."""
         return self.environment.map(node.value)
 
+    def process_Tuple(self, block: 'Block', node: ast.Tuple) -> ANFNode:
+        """Process tuple literals."""
+        op = self.environment.ast_map[ast.Tuple]
+        elts = [self.process_node(block, e) for e in node.elts]
+        return Apply([op, *elts], block.graph)
+
+    def process_Subscript(self, block: 'Block',
+                          node: ast.Subscript) -> ANFNode:
+        """Process subscripts: `x[y]`."""
+        op = self.environment.ast_map[ast.Subscript]
+        value = self.process_node(block, node.value)
+        slice = self.process_node(block, node.slice)
+        return Apply([op, value, slice], block.graph)
+
+    def process_Index(self, block: 'Block', node: ast.Index) -> ANFNode:
+        """Process subscripts with simple index: `x[y]`."""
+        return self.process_node(block, node.value)
+
+    def process_Attribute(self, block: 'Block',
+                          node: ast.Attribute) -> ANFNode:
+        """Process attributes: `x.y`."""
+        op = self.environment.ast_map[ast.Attribute]
+        value = self.process_node(block, node.value)
+        return Apply([op, value, Constant(node.attr)], block.graph)
+
     # Statement implementations
 
     def process_statements(self, block: 'Block',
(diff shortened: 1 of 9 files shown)


myia/parser.py
```diff
@@ -72,34 +105,30 @@ class Environment:
...
-    Lastly, the environment has a mapping from AST nodes to Myia nodes that the
-    parser uses e.g. for mapping binary operators to the correct primitives.
-
     Functions parsed in different environments will have no nodes in common.
     Functions parsed in the same environment will use the same Myia objects for
     the same Python objects.
 
     """
 
-    def __init__(self, object_map: Dict[int, ANFNode],
```
bartvm: I remember that the problem with `ast_map` was that that there are combinations of AST nodes that map to a single operator (e.g. assign to a tuple means unpacking). That said, I do think there is place for two mappings: One of them maps Python syntax to nodes, the other one maps Python objects to nodes. One is immutable during parsing, the other one is not. Folding them together makes the code harder to understand IMO.

breuleux: They are not folded together. I made a static `ast_map` at the top of the file that maps AST node types to `operator.<op>`. To resolve `ast.Add`, the parser first looks up what operator it is in `ast_map` then it looks up the operator in `object_map`. I think that's sensible, since you're probably never going to want `a + b` to differ from `operator.add(a, b)`.

myia/parser.py
```diff
@@ -311,13 +381,73 @@ def process_Return(self, block: 'Block', node: ast.Return) -> 'Block':
...
+                    op = self.environment.map(operator.setitem)
+                    obj = self.process_node(block, targ.value)
+                    idx = self.process_node(block, targ.slice)
+                    new_node = Apply([op, obj, idx, anf_node], block.graph)
+                    write(targ.value, new_node)
+                else:
+                    # UNSUPPORTED: f()[x] = value
+                    raise NotImplementedError(
+                        "You can only set a slice on a variable."
+                    )  # pragma: no cover
```
abergeron: I don't think we should support subscript assignment at all since it changes the semantic from an inplace modification to a copy-and-modify.

In the face on aliasing, it will break code.  We attempt to discover a simple aliasing case when the value is not a local variable, but we cannot be certain that it is safe with only the information that the parser has here.  This would need flow analysis to make sure it's ok.

At that point it would be majorly confusing for the user to have this syntax work sometimes and not others for seemingly arbitrary reasons.

myia/parser.py
```diff
@@ -284,6 +322,38 @@ def process_NameConstant(self, block: 'Block',
...
+    def process_Subscript(self, block: 'Block',
+                          node: ast.Subscript) -> ANFNode:
+        """Process subscripts: `x[y]`."""
+        op = self.environment.map(operator.getitem)
+        value = self.process_node(block, node.value)
+        slice = self.process_node(block, node.slice)
+        return Apply([op, value, slice], block.graph)
+
+    def process_Index(self, block: 'Block', node: ast.Index) -> ANFNode:
+        """Process subscripts with simple index: `x[y]`."""
```
abergeron: This docstring is really confusing for someone not familiar with the python AST.  Maybe something like: `Process subscript indexes`

myia/parser.py
```diff
@@ -311,13 +381,73 @@ def process_Return(self, block: 'Block', node: ast.Return) -> 'Block':
     def process_Assign(self, block: 'Block', node: ast.Assign) -> 'Block':
         """Process an assignment."""
         anf_node = self.process_node(block, node.value)
-        if len(node.targets) == 1 and isinstance(node.targets[0], ast.Name):
-            anf_node.debug.name = node.targets[0].id
-            block.write(node.targets[0].id, anf_node)
-        else:
-            raise NotImplementedError(node.targets)
+        targ, = node.targets
```
abergeron: We should probably support `a = b = 1` at this point.

myia/parser.py
```diff
@@ -311,13 +381,73 @@ def process_Return(self, block: 'Block', node: ast.Return) -> 'Block':
...
+                    op = self.environment.map(setattr)
+                    obj = self.process_node(block, targ.value)
+                    idx = Constant(targ.attr)
+                    new_node = Apply([op, obj, idx, anf_node], block.graph)
+                    write(targ.value, new_node)
+                else:
+                    # UNSUPPORTED: f().x = value
+                    raise NotImplementedError(
+                        "You can only set an attribute on a variable."
+                    )  # pragma: no cover
```
abergeron: I'm against attribute assignment for the same reasons as subscript assignment above.

myia/parser.py
```diff
@@ -311,13 +381,73 @@ def process_Return(self, block: 'Block', node: ast.Return) -> 'Block':
...
+        # create two multiplication nodes, but if we integrate CSE later
+        # that problem will go away on its own.
+        app = ast.BinOp()
+        app.op = node.op
+        app.left = node.target
+        app.right = node.value
+        ass = ast.Assign()
+        ass.targets = [node.target]
+        ass.value = app
+        return self.process_Assign(block, ass)
```
abergeron: I think we should disallow augmented assignment since that too implies inplace modification.

myia/parser.py
bartvm: We've discussed this in the past. I think there are pros and cons for each approach, let me rehash them:

* `x[i] = v` is much more user-friendly than `x = tensor.set_subtensor, x, i , v)`
* It will make it a lot easier to compile existing functions with minimal changes; otherwise I need to go through all of my NumPy code changing subscript assignments to their functional equivalent. This is problematic if one of our goals is for `@myia` being all that is needed to get speedups.
* This really only matters if `x` is used afterwards in a different scope
* If we want to support e.g. `print` statements or `rand` then we will need functions with side-effects. Whatever solution we use in those cases might be used here as well.
* We might be able to do some clever things and rewrite this:
```python
def f(x):
    x[i] = v
    return sum(x)

s = f(x)
print(x)
```
do this
```python
def f(x):
    x[i] = v
    return x, sum(x)

x, s = f(x)
print(x)
```

On the other hand:

* The rewriting approach I mentioned above might get pretty tricky in the presence of HOFs and closures.
* If we don't rewrite, then diverging from Python behavior is risky, and no matter how clearly we put it in the documentation that `x[i] = v` should be read as `x is x with x[i] = v` some people will be surprised by this behavior.
* Using a function call is more explicit, less magic, which is good

I'm a bit on the fence. If we can come up with a "functional update" syntax that isn't too bad (like OCaml's), then I could be more easily swayed to the explicit approach. Maybe something like `x = x.set(i, v)`?

breuleux: I think it would be premature to try to settle this issue now, so what I'm going to do is pull out the support here and leave it in a new branch so that we can apply it later if we decide that this is something we want.

Alternatively, we could leave it and have a flag in the syntax checker to turn it on or off, but that's not in clean yet.

lamblin: Just my two cents here:
- I think supporting `x += 1` should not be any trouble, since it can legitimately be understood by Python as a reassignment, `x = x + 1`
- No matter which syntax we come up with, numerous users will expect `x.set(i, v)` or `set_subtensor(x[i], v)` to perform the assignment in itself, and expect `x` to hold the new value subsequently (even if they do not assign the return value, or assign it to `new_x`).
- On the other hand, users would also expect that to actually produce side effects. For instance, they would expect:
```python
y = x
x[i] = v
assert y[i] == v
```
So unless we can correctly detect that case and forbid pseudo-inplace updates when the original value is aliased, maybe we're better off forbidding it altogether.

breuleux: `x += y` amounts to `x = x.__iadd__(y)`, which is a destructive operation for several types, so we'd still have to forbid it for lists, ndarrays, and so on.

abergeron: My position is not that we should not try to support the `x[i] = v` syntax.  It's rather that we should make sure to preserve the semantics if we do.  Until we can assure that, we should not claim to support it.

In the long term I think it would be possible to do a transform like @bartvm showed to propagate the changes and at that point we should introduce the support.

PR #33, opened 2018-02-06
Toposort

abergeron:
This is a toposort function along with some fixes for flake8 in the tests.

### DIFF
--- a/myia/anf_ir_utils.py
+++ b/myia/anf_ir_utils.py
@@ -1,9 +1,8 @@
 """Utilities for manipulating and inspecting the IR."""
-
 from typing import Iterable
 
 from myia.anf_ir import ANFNode, Constant, Graph
-from myia.graph_utils import dfs as _dfs
+from myia.graph_utils import dfs as _dfs, toposort as _toposort
 
 
 #######################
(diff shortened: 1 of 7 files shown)


myia/anf_ir_utils.py
```diff
@@ -17,3 +17,25 @@ def dfs(root: ANFNode, follow_graph: bool = False) -> Iterable[ANFNode]:
                 to_visit.append(in_)
         if isinstance(node.value, Graph) and follow_graph:
             to_visit.append(node.value.return_)
+
+
+def toposort(graph: Graph) -> Iterator[ANFNode]:
```
breuleux: I would need a more generic `toposort` which takes a `succ` function, i.e.

```python
T = TypeVar('T')
def toposort(root: T, succ: Callable[[T], Iterable[T]]) -> Iterable[T]:
```

I think it would also be useful to raise an exception if there is a cycle, instead of looping infinitely. That would help debug optimizations that accidentally introduce cycles, for example.

tests/test_anf_ir_utils.py
```diff
@@ -18,3 +18,26 @@ def test_dfs_graphs():
...
+
+
+def test_toposort():
+    g0 = Graph()
+    g0.output = Constant(1)
+    g1 = Graph()
+    in0 = Constant(g0)
+    value = Apply([in0], g1)
+    g1.output = value
+    assert list(toposort(g1)) == [in0, value, g1.return_.inputs[0], g1.return_]
```
breuleux: Suggestion: in addition to checking the list explicitly, you can automatically check that the sort is correct with this function:

```python
def _check_toposort(order, root):
    nodes = set(dfs(root))
    assert len(order) == len(nodes)
    for node in nodes:
        for i in node.inputs:
            assert order.index(i) < order.index(node)
```

abergeron:
This should be ok now.

PR #35, opened 2018-02-08
Add node construction methods to Graph

breuleux:
This is a draft proposal to make explicit graph construction and reusing Constants easier.

* Add an `environment` attribute to `Graph`. That way, graph transforms and optimizer can easily get ahold of it and use it to build constants.
* Add `Environment.apply` to build an `Apply` node. That method could eventually implement common subexpression elimination.
* Rename `Environment.map` to `Environment.constant`, because I find it clearer, and I don't think it's meant to return non-constants.
* Add `Graph.apply`. `g.apply(primops.add, node, 1)` is the same as `Apply([g.environment.constant(primops.add), node, g.environment.constant(1)], g)`, but a lot easier to write and read.
* Add `Graph.add_parameter`, which creates a parameter for that graph and also adds it to the parameter list. The new parameter is returned.
* Add `Graph.constant` as a shortcut to `Graph.environment.constant`, for convenience.

I also hide the implementation detail that we use `id(obj)` as the key in environments, which adds the flexibility of not doing that in the future. I'll note that there is some danger with that approach, because an object's id can be reused after the GC reclaims it. In the case of functions, where the constant contains a Graph and might not contain a pointer to the original function, it is possible (although unlikely) that we'd get a collision.

Another point to discuss is whether `Environment` should use a `WeakValueDictionary`. That would allow the GC to reclaim unused constants.

### DIFF
--- a/myia/anf_ir.py
+++ b/myia/anf_ir.py
@@ -74,7 +74,7 @@ class Graph:
         """Create a constant for the given object."""
         return Constant(obj)
 
-    def apply(self, inputs, graph):
+    def apply(self, *inputs):
         """Create an Apply node with given inputs, bound to this graph."""
         inputs = [i if isinstance(i, ANFNode) else self.constant(i)
                   for i in inputs]
--- a/myia/anf_ir.py
+++ b/myia/anf_ir.py
@@ -64,6 +64,22 @@ class Graph:
         else:
             self.return_ = Apply([Constant(primops.return_), value], self)
 
+    def add_parameter(self):
+        """Add a new parameter to this graph (appended to the end)."""
+        p = Parameter(self)
+        self.parameters.append(p)
+        return p
+
+    def constant(self, obj):
+        """Create a constant for the given object."""
+        return Constant(obj)
+
+    def apply(self, inputs, graph):
+        """Create an Apply node with given inputs, bound to this graph."""
+        inputs = [i if isinstance(i, ANFNode) else self.constant(i)
+                  for i in inputs]
+        return Apply(inputs, self)
+
     def __str__(self) -> str:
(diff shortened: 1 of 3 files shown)


breuleux:
> I'll note that there is some danger with that approach, because an object's id can be reused after the GC reclaims it. In the case of functions, where the constant contains a Graph and might not contain a pointer to the original function, it is possible (although unlikely) that we'd get a collision.

Turns out it's not that unlikely. It happened in a test in the `type_inference` branch when I tried to run them on my machine: two different functions were created with the same id, and one test failed because of it. @abergeron suggested using [weakref.finalize](https://docs.python.org/3/library/weakref.html#weakref.finalize) to remove an object's id from the map when the GC collects it. I think it's either that or use the actual objects as keys.

breuleux:
I changed the PR to simply add a few methods to `Graph`. If that looks good, I'll change the rest of the code to use the new methods.

myia/anf_ir.py
```diff
@@ -64,6 +64,22 @@ def output(self, value: 'ANFNode') -> None:
...
+        """Add a new parameter to this graph (appended to the end)."""
+        p = Parameter(self)
+        self.parameters.append(p)
+        return p
+
+    def constant(self, obj):
+        """Create a constant for the given object."""
+        return Constant(obj)
+
+    def apply(self, inputs, graph):
```
abergeron: Why does this take `graph`?

breuleux: That's a mistake from copy pasting the version of that function I had implemented on `Environment`. Should be `def apply(self, *inputs)`.

abergeron:
This has been here long enough, time to merge

PR #38, opened 2018-02-12
Add finalizers to purge expired objects from the environment

abergeron:
Fixes the bug where sometimes functions get mapped to the graph of a previous function that was reclaimed by the GC.

### DIFF
--- a/myia/api.py
+++ b/myia/api.py
@@ -10,7 +10,7 @@ from myia.vm import VM as VM_
 from myia.py_implementations import implementations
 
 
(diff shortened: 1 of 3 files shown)


abergeron:
To pass the tests this will need to be merged after #39

abergeron:
I've discovered and fixed a bug in function mapping with this PR.  Basically inner functions would overwrite the mapping of the top-level function in the environment and this would break in cases involving mutual recursion of global functions.

PR #39, opened 2018-02-12
Fix the version of code checkers to avoid new errors on clean

abergeron:
I've also fixed all the errors with the selected versions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ setup(
     packages=find_packages(exclude=['docs', 'tests']),
     install_requires=[],
     extras_require={
-        'test': ['flake8', 'pytest', 'codecov', 'pytest-cov', 'mypy',
-                 'pydocstyle'],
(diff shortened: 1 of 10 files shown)


setup.py
```diff
@@ -24,7 +24,7 @@
     packages=find_packages(exclude=['docs', 'tests']),
     install_requires=[],
     extras_require={
-        'test': ['flake8', 'pytest', 'codecov', 'pytest-cov', 'mypy',
-                 'pydocstyle'],
+        'test': ['flake8==3.5.0', 'pytest', 'codecov',
+                 'pytest-cov', 'mypy==0.560', 'pydocstyle==2.1.1'],
```
bartvm: Versions ideally shouldn't be pinned in `setup.py` (only lower bounded). If you want to recreate a specific environment, creating a `requirements.txt` is better.

PR #41, opened 2018-02-19
Generic unity

abergeron:
This takes the unification engine that was made for the type inference and makes it reusable for other things (like optimizations).

This also contains the commits in #27 so we should merge it after that one.

For review, only the last commit `Make the unification engine generic and reusable.` is new.

### DIFF
--- a/myia/unity.py
+++ b/myia/unity.py
@@ -1,0 +1,503 @@
+"""Unification module."""
+from typing import (Dict, Tuple, List, Iterable, Union, Callable, cast,
+                    TypeVar, Generic, Any, Type)
+from functools import reduce
+
+
+T = TypeVar('T')
+U = TypeVar('U')
+
+Value = Union['Var[U]', U]
+FnFiltT = Callable[[Value[U]], bool]
+FilterT = Union[Iterable[Value[U]], FnFiltT[U]]
+EquivT = Dict['Var[U]', Value[U]]
+
+
+class UnificationError(Exception):
+    """Exception raised for errors in unification."""
+
+
+class Var(Generic[U]):
+    """Basic universal variable type."""
+
+    __slots__ = ('tag',)
+
+    def __init__(self, tag: str) -> None:
+        """Create a Var."""
+        self.tag = tag
+
+    @classmethod
+    def _parse_args(self, args: Tuple[Any, ...], kwargs: Dict[Any, Any]):
+        assert len(kwargs) == 0
+        assert len(args) == 1
+        assert isinstance(args[0], str)
+        return args
+
+    def matches(self, value: Value[U]) -> bool:
+        """Return True if the variable matches the value given.
+
+        Note that this relation is transitive, but not associative.
+        """
+        return True
+
+    def __str__(self) -> str:
+        return self.tag
+
+    def __repr__(self) -> str:
+        return f"Var({self.tag})"
+
+
+class Seq(tuple):
+    """Class to mark sequence of values matched by an SVar."""
+
+    def __repr__(self) -> str:
+        return "Seq" + super().__repr__()
+
+
+class SVar(Var[U]):
+    """Variable to represent a variable length of values."""
+
+    __slots__ = ()
+
+    def matches(self, value: Value[U]) -> bool:
+        """Only does special matching."""
+        return isinstance(value, (Seq, SVar))
+
+    def __str__(self) -> str:
+        return f"*{self.tag}"
+
+    def __repr__(self) -> str:
+        return f"SVar({self.tag})"
+
+
+class UnionVar(Var[U]):
+    """Variable for a possible set of values."""
+
+    __slots__ = ('values',)
+
+    def __init__(self, tag: str, values: Iterable[Value[U]]) -> None:
+        """Create a UnionVar."""
+        super().__init__(tag)
+        self.values = set(values)
+
+    @classmethod
+    def _parse_args(self, args: Tuple[Any, ...], kwargs: Dict[Any, Any]):
+        assert len(kwargs) == 0
+        assert len(args) == 2
+        assert isinstance(args[0], str)
+        return (args[0], tuple(sorted(set(args[1]), key=id)))
+
+    def matches(self, value) -> bool:
+        """Engine bypasses this."""
+        raise UnificationError("matches called on a UnionVar")
+
+    def __repr__(self) -> str:
+        return f"UnionVar({self.tag}, {self.values})"
+
+
+class RestrictedVar(Var[U]):
+    """Variable restricted to a set of values."""
+
+    __slots__ = ('legal_values',)
+
+    def __init__(self, tag: str,
+                 legal_values: Tuple[Value[U], ...]) -> None:
+        """Create a RestrictedVar."""
+        super().__init__(tag)
+        self.legal_values = legal_values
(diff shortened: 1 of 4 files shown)


abergeron:
This branch works now, but it's a bit of a mess with multiple unrelated changes mixed together.  I'll do some cleanup to separate the issues later.

abergeron:
Spoke too soon apparently.  I'll look into the issue.

abergeron:
I did a major reorganization so this PR is now only the generic unification engine, with no unrelated commits.

This should be ready for review and/or merging.

myia/unity.py
```diff
@@ -0,0 +1,503 @@
...
+
+    def __init__(self) -> None:
+        """Create a unification engine."""
+        self._seq = 1
+        self._vars: Dict[Any, Var[T]] = dict()
+
+    def _get_var(self, cls: Type[Var[T]], args: Tuple[Any, ...]) -> Var[T]:
+        pargs = cls._parse_args(args, dict())
+        if pargs not in self._vars:
+            self._vars[pargs] = cls(*pargs)
```
breuleux: I wasn't sure what the utility of caching `Var` instances is, and after talking with Arnaud we determined that it wasn't truly useful and a leftover of earlier design decisions. So I think we should just drop the cache and create `Var` instances directly.

myia/unity.py
```diff
@@ -0,0 +1,503 @@
...
+        Arguments:
+            tag: An identifier for the variable. Two variables with the
+                same filter and identifier will return the same object.
+            filter: A predicate, or a set of values the variable is
+                allowed to take.
+        """
+        tag = self._get_tag(tag)
+        if callable(filter):
+            return self._get_var(FilterVar, (tag, filter))
+        elif filter:
```
breuleux: This will fall through if `filter == ()` and create a `Var` which will match anything, instead of the correct behavior in this case which is to match nothing.

myia/unity.py
```diff
@@ -0,0 +1,503 @@
...
+        """Create a UnionVar (represents multiple possibilities)."""
+        return self._get_var(UnionVar,  # type: ignore
+                             (self._get_tag(tag), values))
+
+    class VisitError(Exception):
+        """Report unvisitable object."""
+
+    def visit(self, fn: Callable[[Value[T]], Value[T]],
+              value: Value[T]) -> Value[T]:
+        """Apply `fn`to each element of `value` and return the result."""
```
breuleux: Missing space between `fn` and `to`.

myia/unity.py
```diff
@@ -0,0 +1,503 @@
...
+        Type are passed through without modification, variables are
+        duplicated with a new id.
+
+        This preserves relationships between variables like this::
+
+            clone(Tuple(v1, v1)) -> Tuple(v2, v2)
+
+        Arguments:
+            v: expression
+            copy_map: Dictionnary of variable mapping
```
breuleux: Dictionary*, mappings*

myia/unity.py
```diff
@@ -0,0 +1,503 @@
...
+                elif isinstance(v, RestrictedVar):
+                    filter = v.legal_values
+                copy_map[v] = self.var(filter=filter)
+
+        elif isinstance(v, Seq):
+            copy_map[v] = Seq(self._clone(val, copy_map) for val in v)
+
+        else:
+            try:
+                return self.visit(lambda v: self._clone(v, copy_map), v)
```
breuleux: Might as well cache the result in `copy_map` as well, if it's hashable.

myia/unity.py
```diff
@@ -0,0 +1,503 @@
...
+            except self.VisitError:
+                return v
```
breuleux: Do we want to ignore this error?

myia/unity.py
```diff
@@ -0,0 +1,503 @@
...
+            # if there is a single match, we record it as the union value
+            equiv.update(ok.popitem()[1])
+            return equiv
+        else:
+            # if there were multiple matches, we try to find the
+            # differences between them and see if we can find
+            # common ground.
+
+            # Essentially we are looking for match differences and
+            # those differences should at most be for a single variable.
```
breuleux: This is to avoid backtracking, right? Might be worth explaining this design decision a bit more.

myia/unity.py
abergeron: ok

myia/unity.py
abergeron: ok

myia/unity.py
```diff
@@ -0,0 +1,503 @@
+"""Unification module."""
```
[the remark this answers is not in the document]
abergeron: I went with "unify.py".

myia/unity.py
```diff
@@ -0,0 +1,503 @@
...
+def noseq(fn: Callable[[Value[U]], Value[U]], u: Value[U]) -> Value[U]:
+    """Make sure that there are no Seq in the value."""
+    um = fn(u)
+    if isinstance(um, Seq):
+        raise TypeError("Multiple values in single-value position")
+    return um
+
+
+class Unification(Generic[T]):
+    """Unification engine."""
```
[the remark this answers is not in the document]
abergeron: I just moved the var(), svar() and uvar() methods out of the class.

myia/unity.py
abergeron: The cache is gone.

myia/unity.py
abergeron: ok

myia/unity.py
abergeron: Yes, we assume it's a non-container value (like Bool() for types).

myia/unity.py
abergeron: This is because I'm not sure what to do if we have more than one difference.  Backtracking will not help if you have two legitimate matches like `(var(),)` and `(var(), svar())` for something like `(1,)`.  I think we can just rule this set of options as not supported.

Also, I have never seen this happen unless I was trying to force it for a test, so I don't think it will be a problem.

myia/unity.py
abergeron: Values still have to be hashable so we just always add them to the map now.

breuleux:
A few issues I've run into adapting this for the optimizer:

* It would be nice to be able to name variables.

* It seems that `Var`'s constructor can take parameters, even though `Var` does not define a constructor. It must be `Generic[U]` that does something weird.

* `reify` does not deal with `SVar` properly. `reify((1, 2, S), {S: (3, 4)})` where `S` is an `SVar` should produce `(1, 2, 3, 4)`, not `(1, 2, (3, 4))`.

abergeron:
> It would be nice to be able to name variables.

done

> It seems that `Var`'s constructor can take parameters, even though `Var` does not define a constructor. It must be `Generic[U]` that does something weird.

Indeed `Generic[U]` does weird things.  I fixed it by implementing __init__ for the point above.

> `reify` does not deal with `SVar` properly. `reify((1, 2, S), {S: (3, 4)})` where `S` is an `SVar` should produce `(1, 2, 3, 4)`, not `(1, 2, (3, 4))`.

Yes and no.  An `SVar` should be matched to a `Seq`, not just a tuple.  Also dealing with it is the responsibility of your visit method (through the `expandlist` and/or `noseq` helper methods if you want).

PR #46, opened 2018-03-15
Graph cloning

breuleux:
Cloning graphs is necessary in several situations, like inlining a graph in another, or creating a graph that can be modified without changing the original. This implements graph cloning with the `GraphCloner` class, which has the following features:

* Multiple graphs can be cloned together. If a graph is cloned, all references to the original, in all cloned graphs, will be replaced by references to the clone.
* Cloning a graph also clones every graph that's nested in it. This required a new method in `NestingAnalyzer` called `scopes`.
* For inlining, a host graph and a list of new parameters can be specified for the clone of a graph.
* The optional `total` option will trigger cloning of every encountered graph, guaranteeing that the clone and the original are completely disjoint.

I believe this corresponds to the "lambda mangling" algorithm described in the Thorin paper.

I also put some utilities in `myia.debug.label` that are useful for writing tests. For example, `myia.debug.label.Index(graph)` creates a mapping from node/graph names to nodes/graphs, so that you can parse a Python function and easily check the node corresponding to an inner function or a variable by name. This is used extensively in the tests for `GraphCloner`.

Note: this PR includes #45, which should be merged before this one.

### DIFF
--- a/myia/clone.py
+++ b/myia/clone.py
@@ -47,7 +47,13 @@ class GraphCloner:
                  relation: str = 'copy') -> None:
         """Initialize a GraphCloner."""
         self.todo: Set[Graph] = set()
-        self.graph_mappings: Dict[Graph, Union[Graph, bool]] = {}
+        # graph_clones maps a graph to either:
+        # * The clone of this graph
+        # * The string 'inline', if the graph was inlined
+        self.graph_clones: Dict[Graph, Union[Graph, str]] = {}
+        # repl maps:
+        # * A node to its clone
+        # * A graph to either its clone or the graph in which to inline it
         self.repl: Dict[Union[Graph, ANFNode],
                         Union[Graph, ANFNode]] = {None: None}
         self.total = total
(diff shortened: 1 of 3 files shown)


myia/clone.py
```diff
@@ -0,0 +1,227 @@
...
+            self.graph_mappings[graph] = target_graph
+        else:
+            self.graph_mappings[graph] = True
+        self.todo.add(graph)
+
+        return target_graph
+
+    def add_clone(self,
+                  graph: Graph,
+                  target_graph: Graph = None,
```
abergeron: I don't see any point where this parameter has a non-default value.  I question its usefulness.

myia/clone.py
```diff
@@ -0,0 +1,227 @@
...
+    is_parameter, is_constant
+from myia.utils import smap
+
+
+#################
+# Graph cloning #
+#################
+
+
+class GraphCloner:
```
abergeron: I'm having trouble wrapping my head around how this all fits together.

Also, I question the usefulness of having this as a class rather than a function.

myia/clone.py
breuleux: It will have a non-default value whenever this is used for inlining, which is actually most of the time. If you wanted to inline an application of `f` in `g`, you would call `add_clone(f, g, apply.inputs)`. I just didn't add the inlining optimizations in yet.

myia/clone.py
breuleux: I think it's easier that way. Cloning a graph entails cloning every graph that's nested in it, and if you want to clone a set of graphs, it is best to do it all at once, because you need to change all calls to these graphs to calls to the cloned versions. A function version would have a lot of parameters and nested helpers, which isn't really any better in my opinion.

PR #50, opened 2018-03-26
Graph analyzer

abergeron:
This contains the type inference and the value inference.

Some of the commits from #41 are also there, but those should go away once that is merged.  The first actual commit of this PR is "Analysis framework."


abergeron:
Since the generic unification was merged, I rebased this.  The only "external" commit left here is from #27 for the type definitions.

myia/analyze/value.py
```diff
@@ -0,0 +1,131 @@
...
+from typing import Dict, Callable
+from .graph import Plugin
+
+from myia import primops as P
+from myia.primops import Primitive
+from myia.utils import Named
+from myia.anf_ir import Constant, Apply, Parameter, Graph
+from myia.py_implementations import implementations as py_implementations
+
+ESTIMATORS: Dict[Primitive, Callable] = dict()
```
breuleux: Use `myia.utils.Registry`

```python
ESTIMATORS: Registry[Primitive, Callable] = Registry()
register_estimator = ESTIMATORS.register
```

myia/analyze/value.py
```diff
@@ -0,0 +1,131 @@
...
+    This has some trouble with loops and such since those are
+    implemented using recursive calls and we don't handle that.
+
+    It does deal with conditionals ok.
+    """
+
+    NAME = "value"
+
+    def __init__(self, *, implementations: Dict[Primitive, Callable] = None,
+                 estimators: Dict[Primitive, Callable] = None) -> None:
```
breuleux: Couldn't we set `py_implementations` and `ESTIMATORS` as default values to the respective parameters, instead of `None`? Saves four lines of code.

myia/analyze/value.py
```diff
@@ -0,0 +1,131 @@
...
+            estimators = ESTIMATORS
+        self._estimators = estimators
+
+    def visit(self, fn: Callable, value):
+        """Nothing to visit."""
+        raise self.analyzer.DU.VisitError  # pragma: no cover
+
+    def on_attach(self):
+        """Add shortcuts."""
+        # Don't explicitely depend on TypePlugin since that will create a cycle
```
breuleux: *Does* it require a `TypePlugin` to be present at all, though?

myia/analyze/value.py
```diff
@@ -0,0 +1,131 @@
...
+from myia.primops import Primitive
+from myia.utils import Named
+from myia.anf_ir import Constant, Apply, Parameter, Graph
+from myia.py_implementations import implementations as py_implementations
+
+ESTIMATORS: Dict[Primitive, Callable] = dict()
+
+
+NO_VALUE = Named('NO_VALUE')
+NOT_CONSTANT = Named('NOT_CONSTANT')
```
breuleux: In what circumstances would a node be associated to the `NOT_CONSTANT` value? From a quick grep of the source I don't see it instantiated anywhere, and it's not tested either.

myia/unify.py
```diff
@@ -481,3 +482,153 @@ def reify(self, v: Value[T], equiv: EquivT[T]) -> Value[T]:
...
+def ddict(**kwargs: Dict[str, Any]) -> DomainDict:
+    """Shorthand to create DomainDicts faster."""
+    return DomainDict(**kwargs)
+
+
+class DomainUnification(Unification[Any]):
+    """Unification that can work across different domains.
+
+    This means that you can run unification on two different sets of
+    values and sometimes reach acorss the sets for a value that you
```
breuleux: *across

myia/unify.py
```diff
@@ -481,3 +482,153 @@ def reify(self, v: Value[T], equiv: EquivT[T]) -> Value[T]:
...
+    need.
+
+    The domains are identified by strings and all the user-facing
+    functions require DomainDicts as input to indetify the staring
```
breuleux: *identify, *starting

myia/unify.py
```diff
@@ -481,3 +482,153 @@ def reify(self, v: Value[T], equiv: EquivT[T]) -> Value[T]:
...
+       ddict(dom1=(val1, val2))
+
+    will assume domain 'dom1' for val1 and val2.  If you want to
+    switch Domains you can do::
+
+      ddict(dom1=(val1, ddict(dom2=val2)))
+
+    This will use 'dom1' for val1 (and the tuple) and 'dom2' for val2.
+
+    This class also changes slightlt the contract of the unify and
```
breuleux: *slightly

myia/unify.py
```diff
@@ -481,3 +482,153 @@ def reify(self, v: Value[T], equiv: EquivT[T]) -> Value[T]:
...
+    unify_raw methods in the presence of DomainDicts.  They require
+    that all the domains in the first dictionary are present in the
+    second, but not the inverse.
+
+    """
+
+    def __init__(self, domains: Dict[str, Callable]) -> None:
+        """Supply intial domains."""
```
breuleux: *initial

myia/unify.py
```diff
@@ -481,3 +482,153 @@ def reify(self, v: Value[T], equiv: EquivT[T]) -> Value[T]:
...
+
+    def _dompush(self, dom):
+        self._domstack.append(dom)
+
+    def _dompop(self):
+        self._domstack.pop(-1)
+
+    def visit(self, fn, value):
+        """Visit according to current domain."""
+        return self._curdom()(fn, value)
```
breuleux: I think I brought it up in conversation a couple weeks ago but I'm going to raise this issue again: wouldn't it be simpler to have one universal `visit` method? Support for new types could easily be added using some `register_visitor` decorator, or a `__visit__` magic method. Different domains are going to visit different types of data, but they are not going to visit the same types differently, as far as I can tell.

tests/analyze/test_graph.py
```diff
@@ -0,0 +1,156 @@
...
+    assert not is_return(v3)
+
+
+def test_Plugin():
+    # This is for coverage only since it is a dumb test
+    p = Plugin()
+    p.on_attach()
+
+
+def test_PLuginManager():
```
breuleux: *test_PluginManager

myia/analyze/prims.py
```diff
@@ -0,0 +1,59 @@
...
+from myia.primops import Primitive
+from myia.dtype import Function, Number, Bool, Tuple
+from myia.unify import var, svar, Var, SVar
+
+
+SIGNATURES: DictT[Primitive, Function] = dict()
+
+
+def register_signature(prim: Primitive, s: Function):
+    """Register the signature for a primive.
```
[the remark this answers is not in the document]
abergeron: ok

myia/analyze/type.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    tt = type(value)
+    if tt in TYPE_MAP:
+        return TYPE_MAP[tt]
+    if tt == tuple:
+        etts = tuple(typeof(e) for e in value)
+        return Tuple(etts)
+    if tt is Primitive:
+        return SIGNATURES[value]
+    raise TypeError("Cannot assign type to: {}".format(value))
```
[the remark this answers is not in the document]
abergeron: ok

myia/analyze/type.py
```diff
@@ -0,0 +1,213 @@
...
+        # Register dependencies
+        analyzer.add_plugin(ValuePlugin())
+
+        analyzer.add_shortcut('infer_type', self.infer_type)
+        analyzer.add_shortcut('infer_args', self.infer_args)
+
+    def infer_type(self, graph: Graph, *args: Type):
+        """Get the return type of a call of `graph`.
+
+        args here should be a the type of the arguments passed-in.
```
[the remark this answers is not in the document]
abergeron: yes

myia/analyze/type.py
```diff
@@ -0,0 +1,213 @@
...
+        """
+        if graph not in self.analyzer.graphs:
+            raise ValueError("Unknow graph")
```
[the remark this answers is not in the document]
abergeron: ok

myia/analyze/type.py
```diff
@@ -0,0 +1,213 @@
...
+                graph_t = Function((self.get_type(p)
+                                    for p in node.graph.parameters), node_t)
+                self.analyzer.graphs[node.graph][self.NAME] = graph_t
+            return node_t
+
+        else:
+            raise AssertionError("Unknown node type")
+
+    def unify_apply(self, node: Apply, equiv):
+        """Unifiy an Apply with its function type."""
```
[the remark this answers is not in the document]
abergeron: ok

myia/analyze/graph.py
```diff
@@ -0,0 +1,221 @@
...
+        return len(self._plugins)
+
+
+class GraphAnalyzer:
+    """Process a graph from inputs to return to perform some analyses."""
+
+    def __init__(self, plugins: Iterable[Plugin]) -> None:
+        """Create the basic attributes and register the plugins."""
+        self.graphs: Dict[Graph, Dict[str, Any]] = dict()
+        self._info_map: Dict[ANFNode, Dict[str, Any]] = dict()
```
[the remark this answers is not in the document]
abergeron: Some plugins manipulate it, yes, but you could argue that plugins are not "external" clients.

It is supposed to be private with regards to users of GraphAnalyzer.

myia/analyze/value.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def register_estimator(prim: Primitive):
+    """Register an estimator for a primitive.
+
+    This serves to estimate the value of a primitive in the present of
```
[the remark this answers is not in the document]
abergeron: ok

myia/analyze/value.py
abergeron: done

myia/analyze/value.py
abergeron: I have this ingrained aversion to mutable containers as default values.  This is because python will return the same object each time which links the calls.

myia/analyze/value.py
abergeron: Not at the moment, but probably a bit later it will.  I could make it a soft dependency though.

myia/analyze/value.py
abergeron: Currently none, you are right.  I should probably just remove it and we can add it back later if we improve the value inference.

myia/unify.py
abergeron: ok

myia/unify.py
abergeron: ok

myia/unify.py
abergeron: ok

myia/unify.py
abergeron: ok

myia/unify.py
abergeron: Different domains have different terminators.  This is not possible with a universal visit method.

For example, when you are matching values and you encounter `Tuple(Float(64), Float(64))`, you don't need to reach inside the tuple.  If Tuple had a visitor for it, you wouldn't be able to make that distinction.

tests/analyze/test_graph.py
abergeron: ok

myia/unify.py
abergeron: We discussed in person and we will move to a model with global visit.

myia/unify.py
```diff
@@ -313,8 +327,12 @@ def unify_union(self, w: UnionVar[T], v: Value[T],
             equiv[diff] = UnionVar(set(okv[diff] for okv in ok_values))
             return equiv
 
-    def unify_raw(self, w: Value[T], v: Value[T],
-                  equiv: EquivT[T]) -> EquivT[T]:
+    def _getvar(self, v):
+        if hasattr(v, '__var__'):
+            return v.__var__()
```
breuleux: I know I had made it a method originally in #52, but I think it may be better to use a `__var__` attribute than a method.

abergeron: ok

PR #51, opened 2018-03-27
Update README

breuleux:
I want to update the README so that people who land on Myia's GitHub page can have an idea what's going on. I took the liberty of saying a beta version might be ready by July (although I don't know if I should).

I'm planning to merge this tomorrow before my GTC talk.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -40,19 +40,20 @@ Roadmap
 Current
 ~~~~~~~
 
-* **Parser**: Supports `def`, `if`, `while`, operators, function calls.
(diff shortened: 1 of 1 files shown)


abergeron:
I'm all for this, but it won't change the main page since that displays the master branch and this PR is on clean.

abergeron:
Maybe we should switch the branches, while keeping a backup of the current master somewhere.

breuleux:
Yes, my plan was to rename `master` to `old` and make `clean` the default branch.

PR #52, opened 2018-04-06
Pattern-based optimization

breuleux:
This adds a simple mechanism to define and apply pattern-based optimizations. A pattern and its replacement can be defined as nested tuples (s-expressions of sorts) in which primitives, constants and variables (Var instances) can figure. For example, to define `x+0 -> x`, you can define `X = Var('X'); PatternSubstitutionOptimization((primops.add, X, 0), X)`

`PatternOptimizerSinglePass` and `PatternOptimizerEquilibrium` can be used to respectively apply a set of patterns to every node in a graph, and to apply the patterns on a set of graphs repeatedly until there are no changes. The implementation is straightforward for now: it toposorts the graph, applies patterns, rinse, repeat. If a smarter process is needed we can deal with that another time.

<s>This includes PR #46, which is required for inlining, although that's pointless right now because inlining is not included or tested in this branch at the moment (it'll come soon enough).</s>


myia/opt.py
```diff
@@ -0,0 +1,205 @@
...
+            return Apply(new_inputs, g)  # type: ignore
+        elif is_parameter(node):
+            g = fn(node.graph)
+            if not isinstance(g, Graph) or g is not node.graph:
+                # Note: this condition will be triggered if e.g. there is a
+                # Parameter in a pattern to reify. It's not clear what that's
+                # supposed to mean unless the Parameter already exists in a
+                # concrete graph, so we raise an Exception just in case.
+                raise Exception('Unification cannot create new Parameters.') \
+                    # pragma: no cover
```
abergeron: Why are you doing this check?

In any case, if you need to keep it, I would make that an assert.

breuleux: Because when I was testing things, I carelessly made a pattern that had a parameter, and it took me a while to realize why matching didn't work. I think I prefer to keep an Exception (maybe TypeError would be more appropriate), in case this (constructing patterns) gets exposed to more developers in the future.

PR #56, opened 2018-04-12
Node replacement and graph isomorphism

breuleux:
* Add `myia.anf_ir_utils.replace(node1, node2)` to replace a node by another in a graph.
* Add `myia.debug.utils.isomorphic(graph1, graph2)` to check whether two graphs are structurally equivalent. This is very handy to test `replace` and optimizations.

The PR also includes the following minor changes:

* Add `cls` argument to `is_constant`, so that e.g. one can write `is_constant(node, int)`.
* Create `myia.debug.utils` to hold some debugging utilities that may be generally useful.
* Move `myia.debug.label.Index` to `myia.debug.utils.GraphIndex`.
* The tests for `replace` use `GraphIndex` and `isomorphic`, and these are likely to be useful for many other tests, so I added tests for both in `tests/debug/test_utils.py`.


myia/debug/utils.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    if len(g1.parameters) != len(g2.parameters):
+        return False
+
+    prev_equiv = equiv
+    equiv = dict(zip(g1.parameters, g2.parameters))
+    if prev_equiv:
+        equiv.update(prev_equiv)
+
+    def same(n1, n2):
```
abergeron: This should be non-recursive.  Also, I would probably move it to anf_ir_utils since it is something that would be generally useful.

breuleux:
I made `isomorphic` non-recursive. I'm not sure if I should move it to `anf_ir_utils` just yet, since it's only used in tests. It could also be moved if/when part of the code requires it.

abergeron:
I would prefer to move it to utils right now so that it is where someone would expect it when they need it.

PR #57, opened 2018-04-19
Move ir-related code into myia.ir submodule

breuleux:
This moves `ir.py`, `anf_ir.py`, `anf_ir_utils.py` and `clone.py` into the `myia.ir` submodule. Most symbols are exported by `ir/__init__.py` so that they can be imported directly from `myia.ir` instead of having to figure out which file it is in.

This also modifies all local imports to use the explicit relative import syntax.


myia/api.py
```diff
@@ -3,10 +3,10 @@
 from types import FunctionType
 from typing import Any, Callable, Dict, List
 
-from myia import parser
-from myia.anf_ir import Graph, Constant, ANFNode
-from myia.vm import VM as VM_
-from myia.prim import implementations, ops as P
+from . import parser
```
lamblin: Did we decide something regarding absolute vs. relative imports?
If so, that would be the opportunity to standardize.

myia/cconv.py
```diff
@@ -8,11 +8,11 @@
 from typing import Dict, Iterable, Optional, Set
 from collections import defaultdict
 
-from myia.anf_ir import ANFNode, Constant, Graph
-from myia.anf_ir_utils import \
+from .ir.anf import ANFNode, Constant, Graph
+from .ir.utils import \
```
lamblin: I vaguely remember pep8 not liking escaping the end of lines, and favoring parentheses.

myia/ir/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+from .clone import GraphCloner  # noqa
+from .utils import (  # noqa
+    succ_deep, succ_deeper, succ_incoming, succ_bidirectional,
+    exclude_from_set, freevars_boundary,
+    dfs, toposort, accessible_graphs,
+    destroy_disconnected_nodes,
+    isomorphic,
+    replace,
+    is_apply, is_parameter, is_constant, is_constant_graph, is_special
+)
```
lamblin: Rather than `# noqa`-ing all the things, it's cleaner to define `__all__`.

myia/debug/inject.py
```diff
@@ -2,7 +2,7 @@
 
 import sys
 import pdb
-from myia.debug.buche import buche
+from .buche import buche
 from buche.debug import BucheDb
 from buche import reader
```
lamblin: To maximize legibility, can we order imports with builtins first, then external dependencies, then current package?
I can't remember if that was in pep8 or another list of good practices.

myia/api.py
breuleux: No. I'm voting for relative with this PR, and if anyone disagrees, we can debate it.

myia/cconv.py
breuleux: I have a slight preference for the backslashes and `flake8` didn't complain. I'm open to changing it, though, if that's what others prefer.

myia/ir/__init__.py
breuleux: I suppose, although I find it annoyingly redundant.

myia/debug/inject.py
breuleux: Yes. That's what I do in the other files, but I suppose I forgot to do it there.

abergeron: We can use isort to do it automatically.  There is even a flake8 plugin to check this.

https://pypi.org/project/isort/

myia/ir/__init__.py
abergeron: I don't like `__all__` since we have to define everything twice.  The solution in Theano was to whitelist all the `__init__.py` files, but `#noqa` for the imports is better I think.

It would be nice if we could just silence that warning, but it's the same as the one for regular files.

myia/api.py
abergeron: I would stay with absolute for top-level imports (like here), but use relative within a sub-package.

The reason is that it reduces the changes required when we change files to packages or move things around.

breuleux: @abergeron Do you mean using `from myia.ir import ...` from sub-packages instead of `from ..ir import ...`? Because I've done the latter.

myia/debug/inject.py
breuleux: Well, that's useful. `isort -ca -m 2 myia/**.py` seems good.

myia/cconv.py
breuleux: `isort` offers seven different ways to handle long imports (scroll down on https://pypi.org/project/isort/). I'd be fine with either 2 or 5. Thoughts?

PR #61, opened 2018-04-30
Graph analyzer

abergeron:
This is take two.

- Now uses abstract evaluation and is much better at inferring values.
- We fix the type of graphs (so they aren't polymorphic anymore).
- MetaVars are back to type some special cases.


abergeron:
Also, as the test results suggest, I need to write more doc.

abergeron:
@breuleux has a better solution to this.

PR #64, opened 2018-05-03
Remove mypy testing

### LINKED ISSUE #48: Get rid of mypy
I want to discuss this, but I feel like mypy is causing more problems that it solves.

Its type inference is minimal at best, it gets confused the moment you do something that is slightly complex (like a decorator) and is generally unhelpful.  Moreover, I realized that I spend about 10%-20% of my time figuring out how to organize the code to make it happy.  This is not productive and is starting to be a drain on my time.

If we have a consensus on this, I would submit a PR to remove mypy from travis.  I don't know if we want to keep the annotations in place or not.  They can be useful for documentation at least.

abergeron:
Since there are basically no objections to #48, here is a PR to implement it.

The motivator for me is that I'll need numpy for array types and mypy chokes on it.

Fixes #48

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,6 +99,3 @@ venv.bak/
 
 # mkdocs documentation
 /site
-
-# mypy
-.mypy_cache/
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ env:
   - TEST_SUITE=static
   - TEST_SUITE=unit
 script:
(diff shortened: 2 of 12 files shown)


breuleux:
For good measure, could we also get rid of every `# type: ignore` comment and every call to `cast(...)`?

abergeron:
Done.

PR #67, opened 2018-05-04
Type inference

breuleux:
This implements type and value inference (shape inference can easily be added in the future) using a method based on `async` and `await`. This algorithm does not involve variables or unification. Here's what works:

* Type inference works through conditionals, loops, recursion.
* Type inference can rely on value inference and vice versa, so e.g. `getitem` on a tuple can infer the return type based on the value of the index.
* Graphs can be polymorphic (closures, too).
* Higher order primitives such as `map` can be typed, even if their argument is a graph or a closure.
* Higher order graphs can also be typed. The inferrer can deal with stuff like returning a different function depending on a condition (see `test_infer::test_hof_2`), although that may need to be tested a bit more extensively.

I think the algorithm is fairly solid, the only part that I trust a little less is the `InferrerEquivalenceClass` class that I use to determine whether two functions have the same type. It essentially collects every input type signature each function has seen, and checks that every function in the equivalence class has the same return type for each tuple of input types (keep in mind the relationship between input type and return type is an arbitrary function). It's powerful as hell, but I wouldn't be surprised if there was a bug or two.

I think this is probably already good enough, but there are some improvements left:

* The value inferrer currently infers everything it possibly can and there needs to be a way to prevent it from completely unrolling all loops when it knows the number of iterations. Basically a cap on the depth of the search tree.
* Some graphs can lead to deadlocked coroutines, which is currently solved using a very coarse timeout. I believe there are better ways to handle this. (Note: these graphs don't have valid types, I mean stuff like `def f(): return f()`)
* I want to add limited support for `isinstance`. This should allow graphs that dispatch on the type of their arguments. This shouldn't require any modifications to the inferrer itself. This will allow us to define e.g. `a + b` as the application of a graph that dispatches to lower-level primitives like `add_f32`, `add_i64`, etc. depending on input types.

After this, I plan on combining the results of the inferrer with existing optimization code in order to produce a type-specialized graph with the property that every node in every graph is associated to a single type. In a nutshell, a new graph would be built for every possible signature of a polymorphic graph. Then, at last, codegen should be a breeze.

### DIFF
--- a/myia/infer.py
+++ b/myia/infer.py
@@ -11,7 +11,13 @@ from .cconv import NestingAnalyzer
 ANYTHING = Named('ANYTHING')
 
 
-class MyiaTypeError(Exception):
+class InferenceError(Exception):
+    """Inference error in a Myia program."""
+
+    pass
+
+
+class MyiaTypeError(InferenceError):
     """Type error in a Myia program."""
 
     pass
@@ -557,7 +563,7 @@ class InferenceEngine:
             argrefs = [self.ref(node, ctx) for node in n_args]
             try:
                 return await inf(*argrefs)
-            except MyiaTypeError as e:
+            except InferenceError as e:
                 self.log_error([ref], e)
                 raise
             except RuntimeError as e:
@@ -633,13 +639,13 @@ class InferenceEngine:
                 e.g. References.
             err: Can be one of:
                 * An exception.
-                * A string to wrap as a MyiaTypeError.
+                * A string to wrap as an InferenceError.
                 * A future. If the future failed due to an exception,
                   the exception is logged, otherwise the method returns
                   without doing anything.
         """
         if isinstance(err, str):
-            err = MyiaTypeError(err)
+            err = InferenceError(err)
         elif isinstance(err, asyncio.Future):
             err = err.exception()
             if err is None:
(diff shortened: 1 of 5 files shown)


myia/infer.py
```diff
@@ -0,0 +1,618 @@
...
+        return self.inferrer_fn(self.engine, *args)
+
+    def provably_equivalent(self, other):
+        """Whether this inferrer is provably equivalent to the other.
+
+        Two PrimitiveInferrers are equivalent if they use the same
+        inferrer function.
+        """
+        return isinstance(other, PrimitiveInferrer) \
+            and other.inferrer_fn == self.inferrer_fn
```
abergeron: Maybe compare nargs too.

myia/infer.py
```diff
@@ -0,0 +1,618 @@
...
+    def __hash__(self):
+        return hash((self.engine, self.parts))
+
+    def __eq__(self, other):
+        return type(other) is Context \
+            and self.engine == other.engine \
+            and self.parts == other.parts
+
+    def __or__(self, other):
+        return Context(self.engine, self.parts | other)
```
abergeron: Might check that both engines are equal here.

breuleux:
I addressed the review and also tweaked the API a bit, because in its current state the same `InferenceEngine` should not be run for multiple graphs and args (I suppose that wouldn't be too difficult to fix). The graph and args have to be passed to the constructor now.

PR #75, opened 2018-05-23
Array types and shape inference

abergeron:
I've added a minimal set of primitive working on arrays, but they shouldn't be considered final.  They were mostly needed to test the shape inference and interaction with value/type inference.

### DIFF
--- a/myia/prim/py_implementations.py
+++ b/myia/prim/py_implementations.py
@@ -230,7 +230,7 @@ def map_array(fn, array):
 @vm_register(primops.map_array)
 def _map_array_vm(vm, fn, array):
     def fn_(x):
-        return vm.call(fn, x)
+        return vm.call(fn, (x,))
     return map_array(fn_, array)
 
 
--- a/myia/prim/py_implementations.py
+++ b/myia/prim/py_implementations.py
@@ -216,7 +216,7 @@ def shape(array):
     return array.shape
 
 
-@register(primops.map_array)
+@py_register(primops.map_array)
 def map_array(fn, array):
     """Implement `map_array`."""
     def f(ary):
@@ -227,7 +227,14 @@ def map_array(fn, array):
     return np.apply_along_axis(f, 0, array)
 
 
-@register(primops.scan_array)
+@vm_register(primops.map_array)
+def _map_array_vm(vm, fn, array):
+    def fn_(x):
+        return vm.call(fn, x)
+    return map_array(fn_, array)
+
+
+@py_register(primops.scan_array)
 def scan_array(fn, init, array, axis):
     """Implement `scan_array`."""
     # This is inclusive scan because it's easier to implement
@@ -242,7 +249,14 @@ def scan_array(fn, init, array, axis):
     return np.apply_along_axis(f, axis, array)
 
 
-@register(primops.reduce_array)
+@vm_register(primops.scan_array)
+def _scan_array_vm(vm, fn, init, array, axis):
+    def fn_(a, b):
+        return vm.call(fn, [a, b])
+    return scan_array(fn_, init, array, axis)
+
+
+@py_register(primops.reduce_array)
 def reduce_array(fn, init, array, axis):
     """Implement `reduce_array`."""
     def f(ary):
@@ -252,6 +266,13 @@ def reduce_array(fn, init, array, axis):
             val = fn(val, x)
         return val
     return np.apply_along_axis(f, axis, array)
+
+
+@vm_register(primops.reduce_array)
+def _reduce_array_vm(vm, fn, init, array, axis):
+    def fn_(a, b):
+        return vm.call(fn, [a, b])
+    return reduce_array(fn_, init, array, axis)
 
 
 @register(primops.distribute)
--- a/myia/prim/py_implementations.py
+++ b/myia/prim/py_implementations.py
@@ -216,13 +216,13 @@ def shape(array):
 
 
 @register(primops.map_array)
-def map_array(fn, array, axis):
+def map_array(fn, array):
     def f(ary):
         it = np.nditer([ary, None])
         for x, y in it:
             y[...] = fn(x)
         return it.operands[1]
-    return np.apply_along_axis(f, axis, array)
+    return np.apply_along_axis(f, 0, array)
 
 
 @register(primops.scan_array)
@@ -233,7 +233,8 @@ def scan_array(fn, init, array, axis):
         val = init
         it = np.nditer([ary, None])
         for x, y in it:
-            y[...] = fn(val, x)
+            val = fn(val, x)
+            y[...] = val
         return it.operands[1]
     return np.apply_along_axis(f, axis, array)
 
@@ -252,6 +253,11 @@ def reduce_array(fn, init, array, axis):
 @register(primops.distribute)
 def distribute(v, shape):
     return np.broadcast_to(v, shape)
+
+
+@register(primops.reshape)
+def reshape(v, shape):
+    return np.reshape(v, shape)
 
 
 @register(primops.dot)
--- a/myia/prim/py_implementations.py
+++ b/myia/prim/py_implementations.py
@@ -212,11 +212,13 @@ def setattr(data, attr, value):
 
 @register(primops.shape)
 def shape(array):
+    """Implement `shape`."""
     return array.shape
 
 
 @register(primops.map_array)
 def map_array(fn, array):
+    """Implement `map_array`."""
     def f(ary):
         it = np.nditer([ary, None])
         for x, y in it:
@@ -227,6 +229,7 @@ def map_array(fn, array):
 
 @register(primops.scan_array)
 def scan_array(fn, init, array, axis):
+    """Implement `scan_array`."""
     # This is inclusive scan because it's easier to implement
     # We will have to discuss what semantics we want later
     def f(ary):
@@ -241,6 +244,7 @@ def scan_array(fn, init, array, axis):
 
 @register(primops.reduce_array)
 def reduce_array(fn, init, array, axis):
+    """Implement `reduce_array`."""
     def f(ary):
         val = init
         it = np.nditer([ary])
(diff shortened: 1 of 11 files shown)


myia/prim/shape_inferrers.py
```diff
@@ -0,0 +1,157 @@
...
+    if v is True:
+        # We only visit the first branch if the condition is provably true
+        return await tb_inf()
+    elif v is False:
+        # We only visit the second branch if the condition is provably false
+        return await fb_inf()
+    elif v is ANYTHING:
+        # The first branch to finish will return immediately. When the other
+        # branch finishes, its result will be checked against the other.
+        return await engine.assert_same('shape', tb_inf(), fb_inf())
```
breuleux: For safety's sake, I think there should be an else clause that raises an exception if v is somehow not True, False or ANYTHING (e.g. I'm pondering adding an ABSENT keyword for tracks that are not being inferred at all). Same goes for `infer_type_if` which has the same oversight.

myia/prim/shape_inferrers.py
```diff
@@ -0,0 +1,157 @@
...
+    return await engine.get('shape', ary)
+
+
+@shape_inferrer(P.reduce_array, nargs=4)
+async def infer_shape_reduce_array(engine, fn, init, ary, ax):
+    """Infer the shape of reduce_array."""
+    shp = await engine.get('shape', ary)
+    ax_v = await engine.get('value', ax)
+    if ax_v == ANYTHING:
+        return (None,) * (len(shp) - 1)
```
breuleux: Could we use ANYTHING or some special object instead of None?

myia/prim/type_inferrers.py
```diff
@@ -199,6 +199,75 @@ def default(self):
     return t
 
 
+@type_inferrer(P.shape, nargs=1)
+async def infer_type_shape(engine, ary):
+    """Infer the return type of shape."""
+    shp = await engine.get('shape', ary)
+    return Tuple([Int(64)]*len(shp))
```
breuleux: Is `shp` guaranteed to be a tuple? Just making sure.

myia/prim/type_inferrers.py
```diff
@@ -199,6 +199,75 @@ def default(self):
...
+    ary_t = await engine.get('type', ary)
+    if not isinstance(ary_t, Array):
+        raise MyiaTypeError('Expected array')
+    xref = engine.vref({'type': ary_t.elements})
+    return Array(await fn_t(xref))
+
+
+@type_inferrer(P.scan_array, P.reduce_array, nargs=4)
+async def infer_type_across_array(engine, fn, init, ary, ax):
+    """Infer the return type of map_array."""
```
breuleux: Wrong doc.

myia/prim/type_inferrers.py
```diff
@@ -199,6 +199,75 @@ def default(self):
...
+    ary_t = await engine.get('type', ary)
+    init_t = await engine.get('type', init)
+    ax_t = await engine.get('type', ax)
+    if not isinstance(ary_t, Array):
+        raise MyiaTypeError('Expected array')
+    if not ary_t.elements == init_t:
+        raise MyiaTypeError("Initial value must have the same type "
+                            "as array elements")
+    if not ax_t == UInt(64):
+        raise MyiaTypeError("Axis must be u64")
```
breuleux: Do we even have any way to create a `UInt(64)` at the moment?

myia/prim/type_inferrers.py
```diff
@@ -199,6 +199,75 @@ def default(self):
...
+    """Infer the return type of distribute."""
+    v_t = await engine.get('type', v)
+    if isinstance(v_t, Array):
+        v_t = v_t.elements
+    elif not isinstance(v_t, (Number, Bool)):
+        raise MyiaTypeError("Array elements must be numbers or bool")
+    shp_t = await engine.get('type', shp)
+    if (not isinstance(shp_t, Tuple) or
+            not all(e == UInt(64) for e in shp_t.elements)):
+        raise MyiaTypeError("Shape must be (i64, ...)")
```
breuleux: `u64`, not `i64`

myia/prim/type_inferrers.py
```diff
@@ -199,6 +199,75 @@ def default(self):
...
+    return Array(v_t)
+
+
+@type_inferrer(P.reshape, nargs=2)
+async def infer_type_reshape(engine, v, shape):
+    """Infer the return type of reshape."""
+    shp_t = await engine.get('type', shape)
+    if (not isinstance(shp_t, Tuple) or
+            not all(e == UInt(64) for e in shp_t.elements)):
+        raise MyiaTypeError("Shape must be (i64, ...)")
```
breuleux: See above.

myia/prim/shape_inferrers.py
```diff
@@ -0,0 +1,157 @@
...
+
+def prod(iterable):
+    """Return the product of the elements of the iterator."""
+    return reduce(operator.mul, iterable, 1)
+
+
+shape_inferrer_constructors = {}
+
+
+class DefaultShapeInferrer(Inferrer):
```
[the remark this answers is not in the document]
abergeron: ok

myia/prim/py_implementations.py
```diff
@@ -195,6 +195,68 @@ def setattr(data, attr, value):
...
+
+@register(primops.map_array)
+def map_array(fn, array):
+    """Implement `map_array`."""
+    def f(ary):
+        it = np.nditer([ary, None])
+        for x, y in it:
+            y[...] = fn(x)
+        return it.operands[1]
+    return np.apply_along_axis(f, 0, array)
```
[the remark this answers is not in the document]
abergeron: No, because vectorize wants to know the signature and I don't want to guess it.

myia/prim/shape_inferrers.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+shape_inferrer = partial(register_inferrer,
+                         constructors=shape_inferrer_constructors)
+
+
+@shape_inferrer(P.return_, nargs=1)
+async def infer_shape_return(engine, v):
+    """Infer the shape of return."""
+    return await engine.get('shape', v)
```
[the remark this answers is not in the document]
abergeron: For consistency's sake, `engine.get()` (32 instances) is more prevalent than `v[]` (27 instances), but if we decide the latter is what we want, I'll make the change.

myia/prim/shape_inferrers.py
abergeron: With the current code, if we aren't inferring values, you would get an exception.

Also, if it's not True, False, or ANYTHING, you'll get an error in the type checker (since the type will not be `Bool()`).

myia/prim/shape_inferrers.py
abergeron: Probably

myia/prim/type_inferrers.py
abergeron: Yes

myia/prim/type_inferrers.py
abergeron: fixed

myia/prim/type_inferrers.py
abergeron: Not with the parser.

myia/prim/type_inferrers.py
abergeron: fixed (both)

myia/prim/shape_inferrers.py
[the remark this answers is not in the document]
breuleux: If you count them on this branch, yes. `v[]` is more common in clean.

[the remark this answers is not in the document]
abergeron: All right, I'll make the change.

myia/prim/shape_inferrers.py
abergeron: I've added a safety branch.

myia/prim/shape_inferrers.py
```diff
@@ -0,0 +1,161 @@
...
+def prod(iterable):
+    """Return the product of the elements of the iterator."""
+    return reduce(operator.mul, iterable, 1)
+
+
+shape_inferrer_constructors = {}
+
+
+class ScalarShapeInferrer(Inferrer):
+    """Shape inferrer for all primitves that don't take arrays."""
```
lamblin: "primitives"

myia/prim/shape_inferrers.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def __init__(self, engine):
+        """Initialize the ScalarShapeInferrer."""
+        super().__init__(engine, 'scalar_shape_inferrer')
+
+    async def __call__(self, *args):
+        """Since no arrays are involved, the shape is always ()."""
+        return ()
```
lamblin: I would find it slightly confusing that the shape of a list or tuple would be `()`.
Could we have a way to distinguish arithmetic scalars from other non-arrays?

myia/prim/shape_inferrers.py
```diff
@@ -0,0 +1,161 @@
...
+        """This is always equal to itself."""
+        return type(self) == type(other)
+
+
+class ShapeTrack(Track):
+    """Infer the shape of a constant."""
+
+    def __init__(self, engine, name, *,
+                 constructors=shape_inferrer_constructors):
+        """Intialize a ShapeTrack."""
```
lamblin: "Initialize"

myia/prim/shape_inferrers.py
```diff
@@ -0,0 +1,161 @@
...
+from . import ops as P
+from .ops import Primitive
+
+
+def prod(iterable):
+    """Return the product of the elements of the iterator."""
+    return reduce(operator.mul, iterable, 1)
+
+
+shape_inferrer_constructors = {}
```
lamblin: I think I missed when this gets populated.

myia/prim/shape_inferrers.py
abergeron: Originally I wanted there to be no shape for all non-arrays, but that confuses the inferrer in numerous spots, so I settled on ().

I don't think that distinguishing scalars from lists and tuple is relevant since we don't support the numpy-style automatic cast to array from those.  I could make it clearer that the shape is only relevant for things that are typed `Array()`

myia/prim/shape_inferrers.py
abergeron: From the `@shape_inferrer` decorator, which is a partial on `register_inferrer` and this dictionary.

abergeron:
Rebased

PR #78, opened 2018-05-24
Make it easier for ops to call one of their arguments.

abergeron:


myia/vm.py
```diff
@@ -165,6 +165,24 @@ def _succ_vm(self, node: ANFNode) -> Iterable[ANFNode]:
         if is_constant_graph(node):
             yield from self._vars[node.value]
 
+    def call(self, fn, args):
+        """Call the `fn` object.
+
+        `fn` can be anthing that would be valid as the first element
```
breuleux: *anything

PR #84, opened 2018-06-12
Pipeline

breuleux:
This defines a pipeline system for Myia. First you build a `PipelineDefinition`, then you call `make()` to create a `Pipeline` object, then you call that object with the required inputs, e.g. `pipeline_def.make()(input=fn, argspec={"type": Int(64)})`.

A `PipelineDefinition` can be configured. Given `myia.api.standard_pipeline`, which defines all stages, you can customize it in several ways:

```python
# Only run these steps: parse, closure convert, and export
pip1 = std.select('parse', 'cconv', 'export')

# Turn off optimization
pip2 = std.configure(opt=False)

# Run optimizations A and B (all three are equivalent)
# Merge means we concatenate [A, B] to previous opts (otherwise we'd be replacing them)
from myia.pipeline import Merge
pip3 = std.configure({'opt.opts': Merge([A, B])})
pip3 = std.configure(opt={'opts': Merge([A, B])})
pip3 = std.configure(opt=Merge(opts=[A, B]))

# Add a new optimization step
# Optimizer.partial creates a partial application of the Optimizer step
pip4 = std.insert_after('opt', opt2=Optimizer.partial(opts=[A, B]))

# Change how deep the value inferrer goes
pip5 = std.configure({'infer.tracks.value.max_depth': 10})
```

`select`, `configure`, `insert_before`, `insert_after` can all be chained.

A pipeline also contains resources which are shared between all stages. Each stage has access to the pipeline object, and to the resources in their `resources` field. `standard_pipeline` defines the resources `py_implementations` that maps primitives to their implementations, and `convert` which can be used to parse functions or convert data structures to Myia's format.

### DIFF
--- a/myia/api.py
+++ b/myia/api.py
@@ -5,6 +5,7 @@ from types import FunctionType
 from typing import Any, List
 
 from . import parser
+from .cconv import closure_convert
 from .infer import InferenceEngine
 from .ir import Graph, clone
 from .opt import PatternEquilibriumOptimizer, lib as optlib
@@ -207,6 +208,26 @@ class Specializer(PipelineStep):
         return {'graph': result}
 
 
+class ClosureConverter(PipelineStep):
+    """Pipeline step to closure convert a graph.
+
+    Inputs:
+        graph: The graph to closure convert.
+
+    Outputs:
+        graph: The closure converted graph.
+    """
+
+    def __init__(self, pipeline_init):
+        """Initialize a ClosureConverter."""
+        super().__init__(pipeline_init)
+
+    def step(self, graph):
+        """Closure convert the graph."""
+        closure_convert(graph)
+        return {'graph': graph}
+
+
 class Exporter(PipelineStep):
     """Pipeline step to export a callable.
 
@@ -263,6 +284,7 @@ standard_pipeline = PipelineDefinition(
                 optlib.inline_unique_uses,
             ]
         ),
+        cconv=ClosureConverter.partial(),
         export=Exporter.partial(
             implementations=vm_implementations
         )
--- a/myia/api.py
+++ b/myia/api.py
@@ -197,10 +197,6 @@ class Specializer(PipelineStep):
         graph: The specialized graph.
     """
 
-    def __init__(self, pipeline_init):
-        """Initialize a Specializer."""
-        super().__init__(pipeline_init)
-
     def step(self, graph, inferrer):
         """Specialize the graph according to argument types."""
         spc = TypeSpecializer(inferrer)
@@ -217,10 +213,6 @@ class ClosureConverter(PipelineStep):
     Outputs:
         graph: The closure converted graph.
     """
-
-    def __init__(self, pipeline_init):
-        """Initialize a ClosureConverter."""
-        super().__init__(pipeline_init)
 
     def step(self, graph):
         """Closure convert the graph."""
--- a/myia/api.py
+++ b/myia/api.py
@@ -2,60 +2,43 @@
 
 import operator
 from types import FunctionType
-from typing import Any, Dict, List
+from typing import Any, List
 
 from . import parser
-from .ir import ANFNode, Graph, clone
+from .infer import InferenceEngine
+from .ir import Graph, clone
 from .opt import PatternEquilibriumOptimizer, lib as optlib
+from .pipeline import PipelineStep, PipelineDefinition
 from .prim import py_implementations, vm_implementations, ops as P
+from .prim.value_inferrers import ValueTrack, value_inferrer_constructors
+from .prim.type_inferrers import TypeTrack, type_inferrer_constructors
+from .prim.shape_inferrers import ShapeTrack, shape_inferrer_constructors
+from .specialize import TypeSpecializer
 from .utils import TypeMap
 from .vm import VM
 
 
-def default_object_map() -> Dict[Any, ANFNode]:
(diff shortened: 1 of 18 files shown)


myia/api.py
```diff
@@ -72,44 +56,250 @@ def _convert_function(env, fn):
...
+    )
+    return pdef.make()(input=obj)['output']
+
+
+############
+# Pipeline #
+############
+
+
+class Converter(PipelineStep):
```
abergeron: This is apparently no longer a PipelineStep

myia/api.py
```diff
@@ -72,44 +56,250 @@ def _convert_function(env, fn):
...
+        graph: The closure converted graph.
+    """
+
+    def step(self, graph):
+        """Closure convert the graph."""
+        closure_convert(graph)
+        return {'graph': graph}
+
+
+class Exporter(PipelineStep):
```
abergeron: Might want to rename that to indicate that it runs with the DebugVM

myia/pipeline.py
```diff
@@ -0,0 +1,516 @@
...
+        return kw, {}
+    valid = spec.args + spec.kwonlyargs
+
+    good = {k: v for k, v in kw.items() if k in valid}
+    bad = {k: v for k, v in kw.items() if k not in valid}
+
+    return good, bad
+
+
+class NS:
```
abergeron: This too.

myia/pipeline.py
```diff
@@ -0,0 +1,516 @@
...
+
+    def __setitem__(self, item, value):
+        self.__dict__[item] = value
+
+    def __repr__(self):
+        args = [f'{k}={v}' for k, v in self.__dict__.items()]
+        return f'NS({", ".join(args)})'
+
+
+class Partial:
```
abergeron: And this.

myia/pipeline.py
```diff
@@ -0,0 +1,516 @@
...
+            'reset': b is returned, or takes primacy over the data in a.
+    """
+    if isinstance(b, MergeMode):
+        mode = b.mode
+        b = b.value
+    assert not isinstance(a, MergeMode)
+    return _merge_map[type(a)](a, b, mode)
+
+
+def _filter_keywords(f, kw):
```
abergeron: This might go better in utils.

myia/pipeline.py
```diff
@@ -0,0 +1,516 @@
...
+        if invalid:
+            keys = ", ".join(f"'{k}'" for k in invalid.keys())
+            raise TypeError(f"{f} has no argument(s) named {keys}")
+
+    def partial(self, **keywords):
+        """Refine this Partial with additional keywords."""
+        return merge(self, keywords)
+
+    def __call__(self, **kwargs):
+        """Run the pipeline on the given arguments, from start to finish."""
```
abergeron: This docstring seems off.

myia/pipeline.py
```diff
@@ -0,0 +1,516 @@
...
+        """
+        if isinstance(step, str):
+            if step.startswith('!'):
+                return self.step_names.index(step[1:])
+            else:
+                plus = 1 if upper_bound else 0
+                return self.step_names.index(step) + plus
+        elif isinstance(step, int) or step is None:
+            return step
+        else:  # pragma: no cover
```
abergeron: This is actually super easy to test.

myia/pipeline.py
```diff
@@ -0,0 +1,516 @@
...
+    def step(self, **kwargs):
+        """Execute this step only.
+
+        The received arguments come from the previous step.
+        """
+        raise NotImplementedError()
+
+    def __call__(self, **args):
+        """Execute the pipeline all the way up to this step."""
+        return self.pipeline[:self.name](**args)
```
abergeron: I'm not sure how would someone call this.

tests/ir/test_clone.py
```diff
@@ -223,6 +223,7 @@ def f(x, y):
 
 
 def test_clone_without_forcing_manager():
+    @clone
```
abergeron: So we clone three times now?

myia/api.py
breuleux: Indeed it isn't.

myia/api.py
breuleux: OK.

myia/pipeline.py
breuleux: If the steps are `parse,resolve,export`, `pipeline.steps.resolve(input=xyz)` would run `parse,resolve` and stop there. That interface was used at some point to call the `convert` step, when it was a step, but maybe it's not relevant anymore.

tests/ir/test_clone.py
breuleux: There is a manager on the result of `@parse`. The `@clone` is basically doing `f = clone(f)`, which gives us a graph without a manager. The subsequent clones are cloning that clone.

myia/pipeline.py
breuleux: Yeah, the first half of `pipeline.py` should be in utils, at which point I think utils should probably be split in a few files. I wanted to do that as a separate PR afterwards.

tests/ir/test_clone.py
abergeron: ok

myia/pipeline.py
abergeron: ok

PR #85, opened 2018-06-13
Optimized VM for runtime

abergeron:
This is my current code for a runtime VM, still in python because I'm still prototyping.  The code is a bit messy so it will have to be reorganized before we can merge.  I've shared it to get early feedback on the design.

It is able to run all myia control primitives (if, partial, return, calling graphs/partials/variables).  It doesn't do tail call yet.

It has support for limited expressions (add, sub, mul, div, uadd, usub) since those were easy to map to nnvm, the other operators would require a bit of work, but hopefully not too much.


abergeron:
Now this should work correctly, with tailcalls and tailifs but it's still missing tests.  I'll get on that.

myia/compile/transform.py
```diff
@@ -0,0 +1,322 @@
...
+        self.max_height = max(self.max_height, self._height)
+
+    def add_instr(self, instr, *args):
+        """Append instruction to the list."""
+        self.instrs.append((instr,) + args)
+
+    def push(self, node):
+        """Simulate pushing the value for node on the stack.
+
+        This records the postion so that other nodes can refer to this
```
breuleux: position*

myia/compile/transform.py
```diff
@@ -0,0 +1,322 @@
...
+        """
+        assert node not in self.slots
+        self.slots[node] = self.height
+        self.height += 1
+
+    def ref(self, node):
+        """Get the stack reference for the value of a node.
+
+        This can actually cause a push if the node is a constant that
+        wasn't referreded to before.
```
breuleux: *referred

myia/compile/vm.py
```diff
@@ -0,0 +1,267 @@
...
+class FinalVM:
+    """Run a sequence of instructions.
+
+    These instructions can represent multiple graphs with arbitrary
+    recursion between them.
+    """
+
+    def __init__(self, code):
+        """Create a VM with the specified instructions."""
+        self.code = tuple(code)
```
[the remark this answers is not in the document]
abergeron: They are documented at the beginning of eval, but I can also do that here.

myia/compile/transform.py
```diff
@@ -0,0 +1,322 @@
...
+    """Pipeline step to cut the graph into linear portions and control flow.
+
+    Inputs:
+        graph: A graph
+
+    Outputs:
+        splits: list of graph portions
+
+    """
+
```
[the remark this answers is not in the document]
abergeron: Yes and no.  The set of primitives here depends on what the converter and the vm supports, so I don't think it makes sense to have it be configurable if setting anything other than the default doesn't work.

abergeron:
This is now working and tested.  There might still be warts in the design though.

myia/compile/utils.py
```diff
@@ -0,0 +1,19 @@
+"""Utility functions for graph compilation and code generation."""
+
+from ..ir import is_apply
+
+
+def get_outputs(lst, uses, seen):
+    """Return the list of node whose values are required beyond this segment.
```
breuleux: *nodes

myia/compile/debug_lin.py
```diff
@@ -0,0 +1,70 @@
...
+from ..prim.ops import cons_tuple
+from ..vm import VM
+
+
+def make_tuple(vals, g):
+    """Make a graph that builds a tuple from the specified values."""
+    tup = g.constant(())
+    for v in reversed(list(vals)):
+        tup = g.apply(cons_tuple, v, tup)
+    return tup
```
breuleux: I had to implement the same helper in the grad branch, so this could be useful for other code. Could you move it to ir.utils?

myia/compile/transform.py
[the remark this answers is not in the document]
breuleux: Different backends may support different sets of primitives.

myia/compile/utils.py
```diff
@@ -0,0 +1,19 @@
...
+
+    Arguements:
```
breuleux: *Arguments

myia/compile/transform.py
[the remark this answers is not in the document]
abergeron: Then they will write their own tools for splitting graphs.

myia/compile/debug_lin.py
abergeron: I agree it should be shared, but since it depends on prim.ops.cons_tuple, I'm not sure it should be under ir/.

Maybe graph_utils?

abergeron:
I'll merge this in a couple of days if there are no objections.

PR #91, opened 2018-07-09
Support for more Python language features

breuleux:
This adds support for the following language features:

* Conditional expressions: `a if cond else b`
* Boolean expressions: `a and b or c`
* Lambda expressions: `lambda x, y: x + y`
* For loops: `for x in xs: a = a + x`

`for` loops are implemented using three new primitives, `iter`, `hasnext` and `next` (although they have a different interface from Python's `iter` and `next`, so it might be appropriate to rename them). Roughly speaking:

```python
for x in xs:
    a = a + x
```

Becomes:

```python
it = iter(xs)
while hasnext(it):
    x, it = next(it)
    a = a + x
```

Currently, `iter(xs)` just returns the tuple `(0, xs)` as a sort of ad hoc iterator object, and `next` increments the counter in the first element.

### DIFF
--- a/myia/parser.py
+++ b/myia/parser.py
@@ -165,7 +165,6 @@ class Parser:
 
     def make_condition_blocks(self, block):
         """Make two blocks for an if statement or expression."""
-
         with About(block.graph.debug, 'if_true'):
             true_block = Block(self)
         true_block.preds.append(block)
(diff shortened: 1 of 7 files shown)


myia/parser.py
```diff
@@ -262,6 +265,71 @@ def process_Compare(self, block: 'Block', node: ast.Compare) -> ANFNode:
         right = self.process_node(block, node.comparators[0])
         return Apply([ops[0], left, right], block.graph)
 
+    def process_BoolOp(self, block: 'Block', node: ast.BinOp) -> ANFNode:
+        """Process boolean operators: `a and b`, `a or b`."""
+        def fold(block, values, mode='and'):
```
abergeron: This shouldn't have a default value for mode.

breuleux: True.

PR #94, opened 2018-07-19
New primitives: identity, bool_and, bool_or, switch

breuleux:
This adds the following primitives:

* `identity`, the identity function.
* `bool_and`, which implements `a & b` on booleans
* `bool_or`, which implements `a | b` on booleans
* `switch`, which is an eager version of `if` (`switch(a, b, c) == b if a else c` when `b` and `c` are atoms, or alternatively, it is equivalent to `[c, b][a]`)
* The `elim_identity` optimization. self-explanatory.

These primitives allow us to write some conditional/boolean code without creating unnecessary graphs, which will be less taxing on the inferrer and optimizer. They will be used by an upcoming PR.

Note: this is based on the commits in #93.

### DIFF
--- a/myia/prim/py_implementations.py
+++ b/myia/prim/py_implementations.py
@@ -346,6 +346,12 @@ def _list_map_vm(vm, f, xs):
     return list(map(f_, xs))
 
 
+@register(primops.identity)
+def identity(x):
+    """Implement `identity`."""
+    return x
+
+
 @vm_register(primops.resolve)
 def _resolve_vm(vm, data, item):
     """Implement `resolve` for the VM."""
--- a/myia/prim/py_implementations.py
+++ b/myia/prim/py_implementations.py
@@ -146,6 +146,22 @@ def bool_not(x):
     return not x
 
 
+@register(primops.bool_and)
+def bool_and(x, y):
+    """Implement `bool_and`."""
+    assert x is True or x is False
+    assert y is True or y is False
+    return x and y
+
+
+@register(primops.bool_or)
+def bool_or(x, y):
+    """Implement `bool_or`."""
+    assert x is True or x is False
+    assert y is True or y is False
+    return x or y
+
+
 @register(primops.typeof)
 def typeof(x):
     """Implement typeof."""
--- a/myia/prim/py_implementations.py
+++ b/myia/prim/py_implementations.py
@@ -402,3 +402,9 @@ def _next(it):
     """Implement `next`."""
     n, data = it
     return (data[n], (n + 1, data))
+
+
+@register(primops.switch)
+def switch(c, x, y):
+    """Implement `switch`."""
+    return x if c else y
--- a/tests/test_infer.py
+++ b/tests/test_infer.py
@@ -15,7 +15,8 @@ from myia.prim import Primitive
 from myia.prim.py_implementations import \
     scalar_add, scalar_mul, scalar_lt, head, tail, list_map, hastype, \
     typeof, scalar_usub, dot, distribute, shape, array_map, array_scan, \
-    array_reduce, reshape, partial as myia_partial, identity, bool_and, bool_or
+    array_reduce, reshape, partial as myia_partial, identity, \
+    bool_and, bool_or, switch
 
 
 B = Bool()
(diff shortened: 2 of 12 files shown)


tests/test_infer.py
```diff
@@ -1064,3 +1065,62 @@ def g(c):
...
+    return bool_or(x, y)
+
+
+@infer(
+    type=[
+        (B, i64, i64, i64),
+        (i64, i64, i64, InferenceError),
+        (B, i64, f64, InferenceError),
+        ({'value': True}, i64, f64, i64),
+        ({'value': False}, i64, f64, f64),
```
abergeron: I'm not convinced that those two last cases should work.  We should spec switch so that types on both branches must match in all cases since they are both evaluated.

tests/test_infer.py
```diff
@@ -1064,3 +1065,62 @@ def g(c):
...
+
+        ({'type': B, 'value': True},
+         {'type': ai64, 'shape': (6, 13)},
+         {'type': ai64, 'shape': (6, 14)},
+         (6, 13)),
+
+        ({'type': B, 'value': False},
+         {'type': ai64, 'shape': (6, 13)},
+         {'type': ai64, 'shape': (6, 14)},
+         (6, 14)),
```
abergeron: I'm not sure if we want to allow for shape mismatch.  I think it might be a better idea if both sides are required to have the same shape.

Also, there are no tests where one of the branches has `ANYTHING` as a shape element.

myia/prim/py_implementations.py
```diff
@@ -146,6 +146,22 @@ def bool_not(x):
     return not x
 
 
+@register(primops.bool_and)
+def bool_and(x, y):
+    """Implement `bool_and`."""
+    assert x is True or x is False
+    assert y is True or y is False
```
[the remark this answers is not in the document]
breuleux: Why not? `bool_and` is only meant to work on boolean values.

tests/test_infer.py
breuleux: Tricky. Consider a statement like `x = switch(hastype(x, Array), identity, to_array)(x)` which ensures that `x` is an `Array`. If `x` is `i64`, `to_array` has type signature `i64 -> Array(i64)`, but `identity` does not. If we force the evaluation of both branches, that statement will fail. Do we want that statement to fail? Possibly. But that removes the one use case I had for `switch` in the core.

tests/test_infer.py
breuleux: I suppose. I'd tend to use the same behavior as for type inference, though, whichever it is.

myia/prim/py_implementations.py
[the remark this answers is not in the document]
abergeron: Yes, but we don't do checks like that for any other primitive.

tests/test_infer.py
abergeron: I guess that's fine if switch is only used as the backend for if (and maybe a few special uses in the core).

tests/test_infer.py
abergeron: Yes, I agree, we should keep the same behaviour as type inference.

myia/prim/py_implementations.py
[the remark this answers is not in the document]
breuleux: We do now. The previous PR added a call to `_assert_scalar(x, y)` to every scalar primitive (which didn't break anything, because they were never called on non-scalars to begin with), and `assert x is True or x is False` in also present in `bool_not`.

tests/test_infer.py
breuleux: I think it's fine from the user's perspective as well. Even if the condition is known to be True, the specializer will ask for the type of both branches nevertheless, and if evaluating the False branch would cause a type error, there will be an InferenceError. Basically, code that would cause an error in Python will also cause an error here.

abergeron: Ok then.

myia/prim/py_implementations.py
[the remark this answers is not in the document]
abergeron: I missed that.  So in that case, this is good.

PR #96, opened 2018-07-23
Misc improvements

breuleux:
This adds a set of small, self-contained improvements, as well as one bug fix.

### DIFF
--- a/myia/api.py
+++ b/myia/api.py
@@ -40,6 +40,7 @@ scalar_object_map = {
     operator.or_: P.bool_or,
     operator.getitem: P.getitem,
     operator.setitem: P.setitem,
+    bool: P.identity,
     getattr: P.getattr,
     setattr: P.setattr,
 }
--- a/myia/api.py
+++ b/myia/api.py
@@ -1,6 +1,7 @@
 """User-friendly interfaces to Myia machinery."""
 
 import operator
+import numpy as np
 from types import FunctionType
 
 from . import dtype, parser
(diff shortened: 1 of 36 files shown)


tests/test_specialize.py
```diff
@@ -249,21 +252,40 @@ def f2():
...
             return f1
         else:
             return f2
 
-    return choose(c)(x, 2), choose(not c)(2, y)
+    return choose(c)(x, y), choose(not c)(x, y)
+
+
+@mark.xfail(reason="Distinct contexts are created for 2 and 3, "
+                   "leading to Problem(POLY).")
```
abergeron: Is this hard to fix?

tests/test_lang.py
```diff
@@ -1,4 +1,5 @@
 
+# from builtins import getattr
```
abergeron: Perhaps remove this.

myia/opt/lib.py
```diff
@@ -289,13 +289,17 @@ def replace_applicator(optimizer, node, equiv):
...
+    in the exact same order, and it must be either a Primitive or a global
+    function.
     """
     g = equiv[G].value
     out = g.output
     if is_apply(out) and out.inputs[1:] == g.parameters:
-        return out.inputs[0]
+        inner = out.inputs[0]
+        if is_constant(inner, Primitive) \
+                or is_constant_graph(inner) and inner.value.parent is None:
```
abergeron: Leave a comment to note that if parent is not `g` this should be ok, but with `None` we are 100% sure it is ok.

myia/prim/py_implementations.py
```diff
@@ -245,7 +245,22 @@ def setitem(data, item, value):
 @vm_register(primops.getattr)
 def _vm_getattr(vm, data, attr):
     """Implement `getattr`."""
-    return vm.convert(getattr(data, attr))
+    from types import MethodType
+    from ..vm import Partial
+    # I don't know how else to get a reference to this type
+    method_wrapper_type = type((0).__add__)
```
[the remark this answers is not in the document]
breuleux: Yeah, let's wait for a more compelling reason to switch to 3.7.

myia/api.py
```diff
@@ -376,8 +364,47 @@ def step(self, graph):
...
+######################
+
+
+scalar_parse = scalar_pipeline \
+    .select('parse', 'resolve') \
+    .make_transformer('input', 'graph')
+
+
+scalar_debug_compile = scalar_debug_pipeline \
+    .select('parse', 'resolve', 'export') \
```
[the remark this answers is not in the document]
breuleux: The step is named `'export'` in both pipelines, even though they use different VM implementations.

myia/compile/debug_lin.py
```diff
@@ -52,6 +52,9 @@ def ref(n):
     outputs = get_outputs(lst, lst[0].graph.manager.uses, set(eqv.keys()))
     g.output = make_tuple((eqv[o] for o in outputs), g)
 
+    # Clone in case g contains subgraphs that have a different manager
+    g = clone(g)
```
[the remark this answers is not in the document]
breuleux: It was when I tried to make an example with `switch`, but then there was another problem with it returning a Graph instead of a numerical reference the VM could jump to.

myia/debug/label.py
```diff
@@ -96,15 +96,20 @@ def label(self, node, force=None, fn_label=None):
                              True if force is None else force)
         elif is_constant(node):
             v = node.value
-            if isinstance(v, (int, float, str, Named, Namespace)) or v == ():
+            if v is None or v == ():
+                return repr(v)
+            elif isinstance(v, (int, float, str, Named, Namespace)):
                 return repr(v)
```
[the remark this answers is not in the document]
breuleux: There is no difference, I just thought it looked a little tidier.

tests/test_specialize.py
breuleux: Not necessarily, broadly speaking I can see two solutions. One is to have a `max_depth` of 0 for non-boolean types, but that may cut several important optimization opportunities. The other is to make a specialized graph for each possible context, and then checking whether they are all isomorphic, including the inferred types for all nodes. If they are all isomorphic, then we can simply return one of them.

tests/test_lang.py
breuleux: Right.

myia/opt/lib.py
breuleux: I'll do that.

myia/api.py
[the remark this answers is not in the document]
abergeron: ok

tests/test_specialize.py
abergeron: If we ever fix this, I think we should use solution 2.

breuleux: I agree.

PR #97, opened 2018-07-23
Add composite core operations to standard_object_map

breuleux:
This defines operators in `standard_object_map` that have semantics closer to Python. Operations like `x + y` can now add scalars to arrays and vice versa. Binary operations on arrays are also broadcasted as in numpy through the composition of the operations `broadcast_shape`, `distribute` and `array_map2`.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -9,6 +9,7 @@ exclude_lines =
     @overload
     @abstractmethod
     @compile
+    @core
     pragma: no cover
     AssertionError
     NotImplementedError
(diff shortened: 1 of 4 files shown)


myia/prim/composite.py
```diff
@@ -0,0 +1,292 @@
+"""Implementations of primitives as graphs."""
```
abergeron: This should probably not be under `prim/`.

breuleux: Moved to `myia/`

PR #99, opened 2018-07-31
Misc improvements (bis)

breuleux:
Most of these changes are meant to improve debuggability and support a future PR implementing neat tracebacks for type errors.

### DIFF
--- a/myia/prim/type_inferrers.py
+++ b/myia/prim/type_inferrers.py
@@ -3,7 +3,7 @@
 
 from functools import partial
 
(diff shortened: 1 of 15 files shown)


myia/prim/type_inferrers.py
```diff
@@ -18,6 +18,12 @@
 type_inferrer_constructors = {}
 
 
+def _shape_type(t):  # noqa: D400
```
abergeron: This should be disabled globally (dot at the end of the first line in docstrings).

PR #103, opened 2018-08-14
Add shape inference for tuple elements.

abergeron:


abergeron:
This is missing tests.

myia/prim/shape_inferrers.py
```diff
@@ -22,6 +23,29 @@ def prod(iterable):
 shape_inferrer_constructors = {}
 
 
+class TupleShape:
+    """Class to distinguish the shape of tuples from the shape of arrays."""
+
+    __slots__ = ['tup']
```
breuleux: Could we use a different name for this field, like `shapes`?

abergeron:
Now it has tests.

PR #105, opened 2018-08-15
Classes

breuleux:
This adds support for Python 3.7's dataclasses through the `Class` type and the `make_record` primitive. Myia will recognize the attributes and methods of any class decorated with `@dataclass(frozen=True)`. `make_record(cls, *args)` is the Myia op used to instantiate a dataclass. Future optimization passes will transform these records into tuples, which should be straightforward to do with all the information collected by the inferrer.

This is built on top of PR #101.

### DIFF
--- a/myia/prim/shape_inferrers.py
+++ b/myia/prim/shape_inferrers.py
@@ -7,7 +7,7 @@ from ..infer import ANYTHING, GraphInferrer, register_inferrer, \
     PartialInferrer, Track, MyiaShapeError, Inferrer,  MetaGraphInferrer
 from ..ir import Graph, MetaGraph
 from ..dtype import Array
-from ..utils import is_dataclass
+from ..utils import is_dataclass_type
 
 from . import ops as P
 from .inferrer_utils import static_getter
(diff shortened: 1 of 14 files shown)


myia/utils/misc.py
```diff
@@ -410,3 +410,8 @@ def __getitem__(self, name):
 def eprint(*things):
     """Print to stderr."""
     print(*things, file=stderr)
+
+
+def is_dataclass(cls):
+    """Returns whether cls is a dataclass."""
+    return isinstance(cls, type) and hasattr(cls, '__dataclass_fields__')
```
abergeron: This is super weird because the name doesn't imply that it checks for a type.  Perhaps improve the name?

myia/prim/shape_inferrers.py
```diff
@@ -65,6 +66,8 @@ def from_value(self, v, context):
             return GraphInferrer(self, v, context)
         elif isinstance(v, MetaGraph):
             return MetaGraphInferrer(self, v)
+        elif is_dataclass(v):
+            raise NotImplementedError('Dataclasses are not supported yet')
```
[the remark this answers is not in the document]
breuleux: No, it's like `Tuple`, it should associate a shape to each field. I figured I'd wait for your PR on supporting shape inference through tuples.

myia/utils/misc.py
breuleux: It does imply that though. A dataclass *is* a type.

myia/prim/shape_inferrers.py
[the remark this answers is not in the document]
abergeron: Ok

myia/utils/misc.py
abergeron: I expect that it checks for instances of dataclasses with this name.

PR #106, opened 2018-08-20
Rewrite dtype

breuleux:
* Every Myia type is now an instance of `type`.
* `ismyiatype(t)` checks if `t` is a Myia type, and `ismyiatype(t, model)` additionally checks that `t` is a subclass of `model`.
* Types are parameterized with `[]`, e.g. `Int[64]` instead of `Int(64)`.
* Use `Bool` instead of `Bool()`

### DIFF
--- a/myia/utils/misc.py
+++ b/myia/utils/misc.py
@@ -415,3 +415,13 @@ def eprint(*things):
 def is_dataclass_type(cls):
     """Returns whether cls is a dataclass."""
     return isinstance(cls, type) and hasattr(cls, '__dataclass_fields__')
+
+
+def as_frozen(x):
+    """Return an immutable representation for x."""
+    if isinstance(x, dict):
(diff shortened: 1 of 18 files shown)


myia/utils/misc.py
```diff
@@ -415,3 +415,13 @@ def eprint(*things):
 def is_dataclass_type(cls):
     """Returns whether cls is a dataclass."""
     return isinstance(cls, type) and hasattr(cls, '__dataclass_fields__')
+
+
+def as_frozen(x):
+    """Return an immutable representation for x."""
+    if isinstance(x, dict):
+        return tuple((k, as_frozen(v)) for k, v in x.items())
```
abergeron: This should also sort the returned items otherwise two calls with `{'x': 1, 'y': 2}`  and `{'y': 2, 'x': 1}` would be unequal.

breuleux: True that.

PR #107, opened 2018-08-22
Infer shape for classes and lists

abergeron:


myia/prim/shape_inferrers.py
```diff
@@ -47,6 +52,46 @@ def __hash__(self):
         return hash((type(self), self.shape))
 
 
+class ListShape:
```
breuleux: Maybe use `@dataclass` for this. It'll generate `__init__`, `__eq__` and `__hash__` for you.

PR #108, opened 2018-08-23
Reorganizing operations

breuleux:
* Add the `operations` module as a sort of repository for the operations the parser inserts in the graph. These are mapped to either primitives or graphs in `api`. The only primitives the parser inserts directly are `resolve` and `return`. This might not be the best way to proceed, but it works.
* `iter`, `next` and `hasnext` now refer to the methods `__myia_iter__`, `__myia_next__` and `__myia_hasnext__`. These are defined on tuples, list and arrays and it is possible to define iterators over custom datatypes.
* `getitem` now refers to the `__getitem__` method. `tuple_getitem`, `list_getitem` and `array_getitem` are the primitives used for tuples, lists and arrays.
* Idem for `setitem` and `len`.
* `cons_tuple` is removed and replaced by the variadic primitive `make_tuple`.
* `head` is removed, because it is redundant with `xs[0]`. `tail` is kept because no other operation does what it does.
* `array_map` is now variadic, and `array_map2` is therefore removed.


myia/prim/shape_inferrers.py
```diff
@@ -214,18 +207,15 @@ def from_value(self, v, context):
...
     seq_t = await seq['type']
-
-    if ismyiatype(seq_t, Tuple):
-        seq_sh = await seq['shape']
-        idx_v = await idx['value']
-        assert idx_v is not ANYTHING
-        return seq_sh.shape[idx_v]
-    # For any other type
-    raise InferenceError("Unknown type")  # pragma: no cover
+    assert ismyiatype(seq_t, Tuple)
```
abergeron: Maybe you can remove that part since the type inferrer will check that.

abergeron: Maybe you can remove that part since the type inferrer checks that.

PR #112, opened 2018-08-27
Make is_constant etc. into methods

breuleux:
`is_constant`, `is_apply`, `is_parameter`, `is_constant_graph` are now methods on `ANFNode`. They are used a lot, and it was annoying to import them from utils.

### DIFF
--- a/myia/ir/utils.py
+++ b/myia/ir/utils.py
@@ -1,10 +1,10 @@
 """Utilities for manipulating and inspecting the IR."""
 
-from typing import Any, Iterable, Set
+from typing import Iterable, Set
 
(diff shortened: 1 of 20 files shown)


myia/ir/utils.py
```diff
@@ -191,31 +191,31 @@ def isomorphic(g1, g2, equiv=None):
     return rval
 
 
-##################
-# Misc utilities #
-##################
+# ##################
+# # Misc utilities #
+# ##################
```
abergeron: Perhaps you should delete those.

breuleux: Oops, right

PR #114, opened 2018-08-29
Add operations: log, exp, sin, cos, tan, cast, hyper_add, zeros_like

breuleux:
* Add arithmetic operations `log`, `exp`, `sin`, `cos`, `tan`. They all require a floating point input at the moment, since I'm not sure what to do on integer inputs.
* Add `scalar_cast(x, t)` to convert one numerical type to another.
* Add `hyper_add`, which adds tuples, lists, arrays, records, element-wise (will be used for gradient).
* Add `zeros_like`, which generates a "zero" with the same type and shape as the input. It works on scalars, arrays, tuples, lists and records. The implementation uses `HyperMap` and `scalar_cast`.


myia/composite.py
```diff
@@ -102,6 +105,43 @@ def usub(x):
...
 
+exp = MultitypeGraph('exp')
+log = MultitypeGraph('log')
+sin = MultitypeGraph('sin')
+cos = MultitypeGraph('cos')
+tan = MultitypeGraph('tan')
+
+
+@core
+@exp.register(Number)
```
abergeron: This is in the wrong order

PR #121, opened 2018-09-12
Maintain inferred properties through graph transforms

breuleux:
This adds all that's necessary to keep type and shape information throughout transformations and optimizations. Only the types of parameters must be maintained manually.

### DIFF
--- a/myia/infer/graph_infer.py
+++ b/myia/infer/graph_infer.py
@@ -476,11 +476,11 @@ class ExplicitInferrer(Inferrer):
         self.retval = retval
         refs = []
         for v in argvals:
-            # Work around having to find defaults for the other tracks
-            # TODO: check if this causes problems
-            d = {name: ANYTHING for name in self.engine.tracks.keys()}
-            d[track.name] = v
-            refs.append(VirtualReference(d))
+            # NOTE: This VirtualReference lacks the other tracks, so in
+            # certain situations such as declaring an ExplicitInferrer
+            # to be equivalent to a GraphInferrer, an error might happen
+            # when the GraphInferrer is called on these references.
+            refs.append(VirtualReference({track.name: v}))
         self.cache[tuple(refs)] = retval
 
     async def infer(self, *args):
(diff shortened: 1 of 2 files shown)


myia/api.py
```diff
@@ -417,6 +431,70 @@ def step(self, graph, inferrer, inference_context):
...
+        pass
+
+    def _on_add_edge(self, event, src, key, dest):
+        src.expect_inferred.update(src.inferred)
+        src.inferred.clear()
+        self.todo.add(src)
+        # self.run()
+
+    def _on_drop_edge(self, event, *args):
+        pass
```
abergeron: Maybe clean up the unneeded event handlers.

myia/infer/graph_infer.py
```diff
@@ -397,6 +466,46 @@ def provably_equivalent(self, other):
...
+    def __init__(self, track, argvals, retval):
+        """Initialize ExplicitInferrer."""
+        super().__init__(track, None)
+        self.argvals = argvals
+        self.retval = retval
+        refs = []
+        for v in argvals:
+            # Work around having to find defaults for the other tracks
+            # TODO: check if this causes problems
+            d = {name: ANYTHING for name in self.engine.tracks.keys()}
```
abergeron: I'm pretty sure this is going to bite us in the ass later.

myia/api.py
breuleux: OK

myia/infer/graph_infer.py
breuleux: After some investigation, if only the relevant track is set, an error will occur when declaring an ExplicitInferrer to be equivalent to a GraphInferrer, because the latter will be executed on the refs created by the former. This only happens in a somewhat contrived test in `test_infer`, so I decided to break that test and mark it xfail for now.

abergeron: Apparently this is not as rare as we think, since two more high-order tests fail in my vm branch.  This may be due to me enabling more inlining by default.

PR #122, opened 2018-09-14
make_list and list_reduce

breuleux:
This adds support in the parser for list literals, as well as the operations `make_list` and `list_reduce`. The empty list is typed as `List[T]` where `T` is a variable, which means it may be forced to a concrete type if necessary, e.g. `[1, 2] + []` would make `[]` into an empty list of `Int[64]`, if we had such a thing as list concatenation. If the empty list is not combined with anything else it will be given the type `List[Problem[VOID]]`.

This also adds `list_reduce(fn, lst, dflt)` which has the same signature as `functools.reduce`, except that `dflt` is always required.

This also generalizes `find_matching_shape` so that it works on any kind of shape, e.g. `find_matching_shape(ListShape((4, 5)), ListShape((7, 5))) == ListShape((ANYTHING, 5))`.


myia/prim/shape_inferrers.py
```diff
@@ -108,27 +108,54 @@ def provably_equivalent(self, other):
         return type(self) == type(other)
 
 
-def find_matching_shape(shps):
-    """Returns a shape that matches all shapes in `shps`."""
-    shp = shps[0]
-    shps = shps[1:]
+@overload
+def _generalize_shape(s1: ListShape, s2):
+    return ListShape(_generalize_shape(s1.shape, s2.shape))
```
abergeron: All of these don't check that s2 is the right type.

myia/prim/shape_inferrers.py
```diff
@@ -439,3 +466,22 @@ def to_element(self, sh):
...
     n = max(len(tx.elements), len(ty.elements))
     return TupleShape([NOSHAPE] * n)
+
+
+@shape_inferrer(P.make_list, nargs=None)
+async def infer_shape_make_list(track, *elems):
+    """Infer the return shape of make_list."""
+    shps = [await e['shape'] for e in elems]
+    if len(shps) == 0:
+        return ListShape(NOSHAPE)  # pragma: no cover
```
abergeron: This is very easy to test.

myia/prim/shape_inferrers.py
breuleux: It's tested in `find_matching_shape` but it won't work recursively, that's true. Fixed.

myia/prim/shape_inferrers.py
breuleux: Yeah, I took it out, though, because the shape of `[]` is kind of undefined, not `ListShape(NOSHAPE)`, in the sense that if you did `[array1] + []` the result would have shape `ListShape(shape(array1))` -- the shape of [] is just ignored. Better just forbid it for now, I think.

PR #123, opened 2018-09-20
Improvements to inference engine

breuleux:
This fixes a few issues:

* When two `PartialInferrers` are declared equivalent, their caches will be merged properly.
* `ScalarShapeInferrer` will populate its cache.
* Removed `required_tracks`.
* Added `tied_tracks` parameter, which allows us to tell the inferrer to infer the shape whenever the type is inferred.


breuleux:
Added a commit to improve `TransformedReference`, so you can specify any primitive or graph to act as the transformer, and it will reuse the existing inferrers on the various tracks. `track.to_element` is removed and replaced by the `getelement` primitive, which is semantically equivalent and only exists within the inferrer.

PR #126, opened 2018-09-26
Parse truediv/floordiv correctly

abergeron:
This also comes with some fixes for closures in the Debug VM that were uncovered by the new tests.


myia/composite.py
```diff
@@ -219,12 +244,40 @@ def int_bool(x):
...
+ui16 = UInt[16]
+i8 = Int[8]
+i16 = Int[16]
+f32 = Float[32]
+f64 = Float[64]
+
+
+@core
+def int_truediv(x, y):
+    """Implementation of `int_truediv_f64`."""
```
breuleux: `int_truediv_f64`?

PR #130, opened 2018-10-16
Remove if primitive

abergeron:


myia/compile/transform.py
```diff
@@ -190,6 +190,10 @@ def step(self, graph, splits):
             if isinstance(split, list):
                 run, inputs, outputs = \
                     self.pipeline.resources.lin_convert(split)
+                # prime the arguemnts because self.ref() can invalidate
```
breuleux:
```suggestion
                # prime the arguments because self.ref() can invalidate
```

myia/opt/lib.py
```diff
@@ -333,22 +318,3 @@ def drop_into_call(optimizer, node, equiv):
...
-    z = equiv[Z]
-
-    y2 = Graph()
-    y2.output = sexp_to_node(((y,), *equiv[Xs]), y2)
-
-    z2 = Graph()
-    z2.output = sexp_to_node(((z,), *equiv[Xs]), z2)
-
-    new = (P.if_, equiv[X], y2, z2)
-    return sexp_to_node(new, node.graph)
```
breuleux: This optimization should still work, with the pattern `(((P.switch, X, Y, Z),), xs)`

abergeron: I've added it back.

PR #132, opened 2018-10-25
Add EnvType and related operations

breuleux:
This adds `EnvType`, which is basically a dictionary indexed by nodes. It is used like this:

```python
e = newenv
e = env_setitem(e, embed(x), value1)  # value1 must have same type as x
e = env_setitem(e, embed(y), value2)  # value2 must have same type as y
env_getitem(e, embed(x), default)     # value1
env_getitem(e, embed(y), default)     # value2
env_getitem(e, embed(z), default)     # default, must have same type as z
```

The `embed` operation is only recognized by the inferrer and produces a `SymbolicKeyInstance` on the value track, with type `SymbolicKeyType`. That key contains a reference to the node and its inferred properties. These can be reduced to some unique integer key after inference.

The `env_add` operation adds two environments together. Ideally we will want to eliminate these calls using the optimizer.

For the gradient, the gradient of a function's closure will be an Env and will have type EnvType, which is only parameterized implicitly through the keys used to get and set items. This makes it easier to make it compatible with the type system.


myia/prim/py_implementations.py
```diff
@@ -585,3 +585,27 @@ def _list_reduce_vm(vm, fn, lst, dflt):
     def fn_(a, b):
         return vm.call(fn, [a, b])
     return list_reduce(fn_, lst, dflt)
+
+
+@register(primops.embed)
+def embed(node):
+    """Implement `embed`."""
+    raise NotImplementedError()
```
abergeron: This should have an implementation in the debug vm if it can't have one here.

myia/prim/type_inferrers.py
```diff
@@ -532,3 +533,35 @@ def chk(data_v, item_v):
...
+    await track.assert_same(track.engine.vref(key_v.inferred), x)
+    return EnvType
+
+
+@type_inferrer(P.env_getitem, nargs=3)
+async def infer_type_env_getitem(track, env, key, default):
+    """Infer the return type of env_getitem."""
+    await track.check(EnvType, env)
+    await track.check(SymbolicKeyType, key)
+    key_v = await key['value']
```
abergeron: What happens if you get ANYTHING here?

myia/prim/shape_inferrers.py
```diff
@@ -365,3 +370,28 @@ def from_value(self, v, context):
...
+@shape_inferrer(P.embed, nargs=1)
+async def infer_shape_embed(track, x):
+    """Infer the return shape of embed."""
+    return NOSHAPE
+
+
+@shape_inferrer(P.env_setitem, nargs=3)
+async def infer_shape_env_setitem(track, env, key, x):
+    """Infer the return shape of env_setitem."""
+    return EnvType
```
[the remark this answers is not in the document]
breuleux: Not really, no, I'll put `NOSHAPE` instead.

myia/prim/py_implementations.py
breuleux: OK, I'll do that.

myia/prim/type_inferrers.py
breuleux: I suppose that could happen if you retrieved it from a data structure. It shouldn't happen under normal circumstances, though, so I'll just throw in an assert.

PR #133, opened 2018-10-25
Add a pass to remove tuple manipulation from the graph

abergeron:
This will leave tuples that are returned from functions for now, but will remove them anywhere else.


tests/test_specialize.py
```diff
@@ -58,6 +59,8 @@ def run_test(args):
                         print(f'   {nlbl} ({type(n).__name__}) :: {n.type}')
                         print(f'      {err.args[0]}')
                     raise verr
+                except:  # noqa: E722
+                    raise
```
breuleux: Why?

PR #134, opened 2018-10-25
Support J operator in the type system

breuleux:
This adds the operators `J` and `Jinv`, the type `JTagged`, the MetaGraph `grad`, and type and shape inference support for all of them.

This is only the type system part, working implementation will be in the next PR.


myia/composite.py
```diff
@@ -595,3 +597,79 @@ def _array_zero(xs):
...
+
+
+class GradOperation(MetaGraph):
+    """Implements the grad(f) operation.
+
+    grad(f)(x, ...) returns df(x, ...)/dx. Derivatives of other inputs are
+    thrown out.
+
+    TODO: This currently will not work on primitives, but it is an easy fix.
+    We just need to know how many parameters f takes.
```
abergeron: I feel that it's easy enough to wrap a primitive in a graph if you really need it that we don't need to implement that.

myia/infer/jinf.py
```diff
@@ -0,0 +1,39 @@
+"""Implements JInferrer."""
+
+from ..prim import ops as P
+from ..debug.label import short_relation_symbols as syms
```
abergeron: This is unecessary.

myia/composite.py
breuleux: We'll see.

myia/infer/jinf.py
breuleux: Helps a bit for debugging.

PR #135, opened 2018-10-26
Gradient generation

breuleux:
This adds gradients for various primitives, and gradient generation. The tests pass and there is 100% coverage, but it's worth mentioning the whole thing is slightly rushed. A few comments:

* There are two xfails which should be fixed at some point.
* Support for array operations has not been added yet.
* The default pipeline doesn't support gradients yet. The one that does is in test_grad, but it's a bit wonky (likely responsible for one of the xfails).
* The grad transform isn't very well documented.
* J/Jinv won't work with the debug VM right now.

### DIFF
--- a/myia/debug/finite_diff.py
+++ b/myia/debug/finite_diff.py
@@ -1,0 +1,270 @@
+"""Estimate gradients with finite differences."""
+
+
+from typing import Callable, Dict, List, Any
+import numpy
+import itertools
+
+from ..utils import smap, overload
+
+
+@smap.variant
+def _zeros_like(self, x: object):
+    return 0
+
+
+class NoTestGrad:
+    """Class to signify that a value's gradient shouldn't be tested.
+
+    I don't fully remember how this is supposed to work.
+    """
+
+    def __init__(self, value):
+        """Initialize NoTestGrad."""
+        self.value = value
+
+
+# The variation applied on an input in either direction to estimate
+# the gradient.
+eps = 1e-10
+
+# The tolerance for the difference between the estimation and the
+# computed gradient.
+rel_error = 1e-03
+
+
+def clean_args(args):
+    """Remove instances of NoTestGrad in the given arguments."""
+    return tuple(a.value if isinstance(a, NoTestGrad) else a for a in args)
+
+
+@overload(bootstrap=True)
+def gen_paths(self, obj: NoTestGrad, path):
+    """Generate all paths to a scalar through an object.
+
+    For example, ((a, b), {'x': c}) would generate the paths
+    (0, 0), (0, 1) and (1, 'x') for a, b and c.
+    """
+    yield from []
+
+
+@overload  # noqa: F811
+def gen_paths(self, obj: (list, tuple), path):
+    for i, x in enumerate(obj):
+        yield from gen_paths(x, path + (i,))
+
+
+@overload  # noqa: F811
+def gen_paths(self, obj: numpy.ndarray, path):
+    for coord in itertools.product(*[range(d) for d in obj.shape]):
+        yield path + (coord,)
+
+
+@overload  # noqa: F811
+def gen_paths(self, obj: object, path):
+    yield path
+
+
+def resolve_path(obj, path):
+    """Follow the given path on the given object."""
+    for p in path:
+        obj = obj[p]
+    return obj
+
+
+@overload(bootstrap=True)
+def gen_variants(self, obj: NoTestGrad, gen, path):
+    """
+    Generate perturbated variants of the given object.
+
+    For each scalar element in obj, generate a list of copies obj where that
+    element has been modified by gen, and the path to that element.
+    Basically:
+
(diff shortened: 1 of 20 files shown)


myia/prim/grad_implementations.py
```diff
@@ -0,0 +1,245 @@
...
+parse = standard_pipeline \
+    .select('parse') \
+    .make_transformer('input', 'graph')
+
+
+def bprop_to_augm(prim, fn):
+    """Given a function for the bprop, make the augmented function."""
+    info = NamedDebugInfo(prim=prim, name=prim.name)
+
+    bprop = parse(fn)
```
abergeron: Add a clone here, otherwise you'll get manager errors.  Also because you modify the returned graph.

tests/test_grad.py
```diff
@@ -0,0 +1,403 @@
...
+        return g(x) * g(x + 10.0)
+
+    def g(x):
+        return x * b
+
+    return f(g, a) + f(g, b)
+
+
+@pytest.mark.xfail(
+    reason="I think the opts to eliminate j/jinv are messing this one up."
```
abergeron: If this is true then the opts should be fixed or disabled.

myia/debug/finite_diff.py
```diff
@@ -0,0 +1,270 @@
...
+
+class NoTestGrad:
+    """Class to signify that a value's gradient shouldn't be tested.
+
+    I don't fully remember how this is supposed to work.
+    """
+
+    def __init__(self, value):
+        """Initialize NoTestGrad."""
+        self.value = value
```
abergeron: Support for NoTestGrad seems to make this ridiculously more complex than it should be.

myia/prim/grad_implementations.py
breuleux: The `clone(outer)` at the end of the function will clone `bprop` along with it, but it's true that it's better to clone a graph before modifying it, so I'll fix that.

tests/test_grad.py
breuleux: I disabled the opts and added support for J/Jinv in py_implementations.

breuleux:
I fixed one of the xfails and added support for J/Jinv in py_implementations. It seems to work.

breuleux:
That would be a good idea, yeah. There is some documentation in another branch which may still be helpful, but it needs to be updated for the new code: https://github.com/breuleux/myia/blob/gradient_transform/docs/gradient.ipynb

I'll try to add more comments and explanations.

breuleux:
I have added more comments and explanations, and I added a jupyter doc (GitHub renders it nicely):

https://github.com/breuleux/myia/blob/gradgen/docs/gradient.ipynb

myia/grad.py
```diff
@@ -0,0 +1,657 @@
...
+        return clone(g)
+    else:
+        return g
+
+
+@overload  # noqa: F811
+def J(graph: Graph, manager):
+    """Implement J on a Graph."""
+    if graph.transforms.get('grad', None):
+        return graph.transforms['grad']
```
lamblin: Would that risk to cause perturbation confusion if `J` is called more than once?

tests/test_grad.py
```diff
@@ -0,0 +1,415 @@
...
+    # if statement, so I prefer to name the test 'test_if'
+    if a > b:
+        return a
+    else:
+        return b
+
+
+@grad_test((4.0, 5.0), (6.4, -7.8))
+def test_if2(a, b):
+    # This is max, but what this is really testing is the most basic
```
lamblin: This is not really `max` any more, I think :)

myia/grad.py
breuleux: No, this is just to ensure a graph is only transformed once. Different calls of the result of `J(f)` will compute their perturbations separately.

tests/test_grad.py
breuleux: Indeed, it is not.

breuleux:
@lamblin I'm not sure what you mean by "add a test with a grad() in the original function".

lamblin:
I mean, a test decorated by `@grad_test`, like the other ones, but where the decorated function calls `grad`  (or something equivalent) already. That would be a test for a second-order derivative.
I understand if we are not there yet, though.

breuleux:
Oh, I see. It can't work with `@grad_test` since the Python version of the function is run to estimate derivatives, and it is not able to deal with `grad` right now. I tried to compile the graph for `grad(grad(f))` on some simple function, but it doesn't work. I roughly know what the problem is, but I'd rather work on that a bit later.

breuleux:
@lamblin @bartvm Is it OK to merge this?

myia/prim/grad_implementations.py
lamblin: OK.

myia/grad.py
lamblin: OK.

PR #136, opened 2018-11-05
Add optimizations

breuleux:
This adds several optimization-related improvements:

* Remove distribute/array_reduce when the target shape is the same as the input's shape.
* Simplify combinations of `env_getitem`, `env_setitem`, and empty envs.
* Fix `is_trivial_graph` (an xfail now passes).
* Add a way to create conditional patterns more simply.

Also a few misc improvements:

* Improved type inference for shape tuples so that we correctly assign the type `UInt[64]` to literal elements of a literal shape.
* The numbered labels for specialized graphs are now more readable.

### DIFF
--- a/tests/test_infer.py
+++ b/tests/test_infer.py
@@ -1276,6 +1276,14 @@ def test_distribute(v, shp):
     return distribute(v, shp)
 
 
+@infer(type=[(ai32_of(3, 7), ai32)],
+       shape=[(ai32_of(3, 7), (3, 7)),
+              (ai32_of(7), (3, 7)),
+              (ai32_of(3), InferenceError)])
+def test_distribute2(v):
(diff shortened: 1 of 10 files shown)


tests/test_infer.py
```diff
@@ -1276,6 +1276,14 @@ def test_distribute(v, shp):
     return distribute(v, shp)
 
 
+@infer(type=[(ai32_of(3, 7), ai32)],
+       shape=[(ai32_of(3, 7), (3, 7)),
+              (ai32_of(7), (3, 7)),
+              (ai32_of(3), InferenceError)])
```
abergeron: Maybe add a check for `ai32_of(1, 7)` or `ai32_of(1)`.

PR #139, opened 2018-11-07
Make the nnvm target configurable

abergeron:


abergeron:
"cuda" has been tested and it works (when you have a cuda-enabled tvm-libs installed, which is not publicly available).

PR #141, opened 2018-11-10
Advanced optimization

breuleux:
This adds several advanced optimizations that are critical in optimizing the graphs created by gradients.

* "Incorporator" optimizations take a graph of the form `f((lambda x: y)(a), b)` and maps it to `(lambda x: f(y, b))(a)`. For example, if you have `(lambda x: (a, b))(c)[0]`, we transform that into `(lambda x: (a, b)[0])(c)`, which opens up an opportunity to optimize the getitem further and eliminate `b`.
  * Such optimizations are triggered for `tuple_getitem` (as above), `env_getitem` and calls (basically `f(x)(y)` => `f'(x, y)`)
* Many more optimizations apply on `switch`:
  * `switch(cond, x, x) => x`
  * `switch(cond, switch(cond, a, b), c) => switch(cond, a, c)`
  * `switch(cond, x, y)[0] => switch(cond, x[0], y[0])`
  * Incorporators apply through switch. For example, `switch(cond, f, g)(x)[0]` becomes `switch(cond, f0, g0)(x)`, where `f0(x)` is equivalent to `f(x)[0]`.
* The following simplification is applied to eliminate `env_add`:
  * `env_getitem(env_add(e1, e2), key, dflt) => hyper_add(env_getitem(e1, key, dflt), env_getitem(e2, key, dflt))`
* A global optimizer eliminates `J` and `Jinv` if no application of these primitives remains on functions.

Taken together, these optimizations appear to work: when applied to the test cases in `test_grad`, they successfully eradicate all occurrences of the `Env` and `JTagged` types. There are most probably situations not covered by these optimizations, but this is a promising start.

The one issue is that the optimizer simply doesn't scale anymore. I was not able to optimize `test_pow10` because it was still running after five minutes, and I'm not sure it's because there's an infinite loop. The fact is that the time to run it balloons as optimizations are added. From `debug_opt` to `opt`, running `test_while_2` goes from 1.5 to 8.5 seconds, for example.

I don't think that the optimization strategy is intrinsically costly. I think it's actually pretty good. However, the naive equilibrium optimizer is terrible.

### DIFF
--- a/myia/api.py
+++ b/myia/api.py
@@ -824,6 +824,31 @@ step_prepare = Preparator.partial(
 )
 
 
+step_debug_opt = Optimizer.partial(
+    phases=dict(
+        main=[
+            # Branch culling
+            optlib.simplify_always_true,
+            optlib.simplify_always_false,
+
+            # Safe inlining
+            optlib.inline_core,
+            optlib.simplify_partial,
+
+            # Miscellaneous
+            optlib.elim_j_jinv,
+            optlib.elim_jinv_j,
+        ],
+        grad=[
+            optlib.expand_J,
+        ],
+        renormalize='renormalize',
+        cse=CSE.partial(report_changes=False),
+        jelim=optlib.JElim.partial(),
+    )
+)
+
+
 step_opt = Optimizer.partial(
     phases=dict(
         main=[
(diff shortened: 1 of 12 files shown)


nouiz:
It was proven in Theano that the equilibrum optimization can't scale!
Make it "smart" by iterating only on the opt that have a chance to do
something isn't enough to make it fast at scale.
But it help in the small/medium scale.

breuleux:
@nouiz For the most part, these are not the same optimizations as Theano had. Most of them monotonically decrease graph size, except for the gradient transform and a few others that involve copying graphs. I tried just playing with the ordering a little, and it greatly improved the optimization time, so I think it's just a matter of prioritizing the graph reductions and then applying the costlier operations only when needed.

PR #144, opened 2018-11-14
Working grad in the VM for the MLP model

abergeron:


myia/api.py
```diff
@@ -895,6 +896,9 @@ def wrapped(*args):
             optlib.getitem_env_add,
             optlib.simplify_array_map,
         ],
+        unfuse=[
+            optlib.unfuse_composite,
+        ],
```
breuleux: I think I would put this as a separate opt step right before validate, because this transform is more a hindrance than an asset to the rest of the optimizer.

myia/opt/lib.py
```diff
@@ -189,6 +189,38 @@ def bubble_op_tuple_binary(optimizer, node, equiv):
...
+    """Tranform array_map on a graph to a graph of array_maps.
+
+    This must be applied to scalar-only graphs.
+    """
+    from ..grad import GraphRemapper
+
+    # This has to be defined inline because of circular imports
+    class UnfuseRemapper(GraphRemapper):
+        def __init__(self, g):
+            super().__init__('unfused', g.graphs_used.keys() | {g})
```
breuleux: You need to implement `gen_constant`, otherwise you'll generate code like `array_map(scalar_sub, xs, 1)`, but what you need to generate is something along the lines of `array_map(scalar_sub, xs, distribute(scalar_to_array(1), shape(xs)))`. This issue will happen with the gradient of `pow`.

Also, you can use the same trick on free variables (implement `gen_fv`). This would allow some closures to be passed to `array_map`.

tests/test_compile.py
```diff
@@ -2,15 +2,18 @@
...
     typeof, scalar_add, partial
 
 compile_pipeline = standard_pipeline
 
+debug_fn = standard_debug_pipeline \
+    .select('parse', 'resolve', 'infer', 'specialize', 'export')
 
-def parse_compare(*tests, optimize=True, array=False):
+
+def parse_compare(*tests, optimize=True, array=False, python=True):
```
breuleux: The `python` option sort of defeats the point of `parse_compare`, which is to compare the result of the call in Python to the result from Myia. If we are testing the gradient, we should be using `grad_test`.

tests/test_model.py
```diff
@@ -99,6 +112,45 @@ def test_forward_infer(model, x):
...
+         {'value': rand(3, 8)},
+         InferenceError),
+    ]
+)
+def test_backward_infer(model, x, y):
+    return grad(cost)(model, x, y)
+
+
+@parse_compare((make_model(), rand(3, 10), rand(3, 8)),
+               array=True, python=False)
```
breuleux: That tests that the model runs, but it doesn't test that its output is sensible, does it? We should use `grad_test`, although it appears we can't use it directly, it'll have to be adapted.

myia/api.py
abergeron: I'm not so sure about that.  It is easier to reason about separate operations than a composite graph.

tests/test_compile.py
abergeron: I've added this option mostly for the grad test on the model.  If we end making grad_test work on dataclasses, then this can be removed.

tests/test_model.py
abergeron: Yes, I agree with everything.

myia/api.py
breuleux: The gradient doesn't generate a very nice graph as an argument to `array_map`, it generates something with calls to `J`, `make_tuple`, nested graphs, and so on, and then relies on the optimizer to clean it all up. It seems it gets cleaned up properly prior to `unfuse`, but I wouldn't rely on this.

myia/opt/lib.py
abergeron: Ok.  I should also add some tests for that.

myia/api.py
abergeron: Then perhaps we should unfuse before the gradient happens.

That way the gradient should be relatively simple.

myia/opt/lib.py
abergeron: implementing `gen_constant` turns out to be more complex than necessary since I have the primitives to worry about and I don't know the proper shape for the constant at that point.

breuleux: The shape can be found in `xs[0].shape` in the remapper's outer scope, but sure.

myia/api.py
breuleux: Yeah, I noticed that issue too. `unfuse_composite` + `getitem_tuple` should be sufficient.

PR #163, opened 2019-02-07
Rewrite of the inference system

breuleux:
This PR changes the inference system to be overall simpler and easier to work with.

Changes to the API:

* The inferrer now lives in `myia.abstract`
* Instead of having value, type, shape tracks, the inferrer works with instances of `AbstractValue` and its subclasses, e.g. `AbstractTuple`. An abstract value can contain elements and it can also have a value, a type or a shape, in its `values` dictionary. At the moment, only `AbstractArray` has a shape. Not all abstract values have a value or a type.
* The argspec for the pipeline must now be a tuple of AbstractValues. A Python object can be converted to an AbstractValue with `myia.abstract.from_value(x, broaden=True/False)`.
* `node.inferred` is gone, replaced by `node.abstract`.
* AbstractValues don't keep a reference to the engine, unlike Inferrers in the previous system.
* There are still Inferrer objects but they are kept in the engine.
* The engine now tracks the set of functions that can be called at any given point in the program.
* NOSHAPE, TupleShape, ListShape, etc. are all gone.
* The old equivalence checker is replaced by `myia.abstract.utils.amerge`.
* I had to move pipeline creation back where it was in the `@myia` decorator because it didn't work otherwise.

Changes to tests:

* The tests in `test_infer` now tests all tracks at once. About 50 tests were cut overall.
* The tests run about twice as fast (47 seconds vs 92 seconds for the current master on my machine). I am not certain where all of the difference comes from, though I know one particular improvement accounts for about 15 seconds, and maybe 4-5 seconds worth of cut tests. Either way, it's a welcome improvement.


myia/abstract/infer.py
```diff
@@ -0,0 +1,679 @@
...
+            }
+        )
+
+    async def infer_constant(self, ctref):
+        """Infer the type of a ref of a Constant node."""
+        v = self.pipeline.resources.convert(ctref.node.value)
+        res = to_abstract(v, ctref.context, ref=ctref)
+        t = build_type(res)
+        if dtype.ismyiatype(t, dtype.Number):
+            prio = 1 if dtype.ismyiatype(t, dtype.Float) else 0
```
abergeron: Do you have tests of 32 vs 64 bits for this?

Should we extend the priorities into an "upcast map"?

This should be done in a separate PR if we decide to do it.

myia/abstract/prim.py
```diff
@@ -0,0 +1,1022 @@
...
+            dataref=r_data,
+        )
+        self.cache[(data, item)] = rval
+        return rval
+
+
+abstract_inferrer_constructors[P.getattr] = _GetAttrInferrer.partial()
+
+
+# setattr = Primitive('setattr')
```
abergeron: Is this a leftover?

abergeron:
This results in a speedup of about 50% on the model specialize_backwards test (from ~6s to ~4s).

However the measurement is hard since the compile phase is having mood swings ranging from 1s to 2.5s.  I'll check what's up with that later.

myia/abstract/data.py
```diff
@@ -0,0 +1,538 @@
...
+    def __init__(self, *poss, value=None):
+        """Initialize an AbstractFunction.
+
+        Provide either *poss or value, not both.
+
+        Arguments:
+            poss: Possible Functions that could be called here.
+            value: Either Possibilities or Pending.
+        """
+        v = Possibilities(poss) if value is None else value
```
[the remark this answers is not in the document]
breuleux: I'll add an assertion.

myia/abstract/infer.py
breuleux: Realistically this is only going to be triggered for Python int/float (64 bits), because that's what a literal in the code is going to be, but I suppose it may be useful to cover other int/float types just to be safe.

I don't know about upcasting, we can discuss that separately if you wish.

myia/abstract/prim.py
breuleux: Yes. Good catch.

breuleux: Although, there's no inferrer for `setattr` yet, so I might leave a comment instead to mark the spot.

abergeron: I'm ok with that.

PR #164, opened 2019-02-11
Elemwise Metagraph

breuleux:
This replaces the previous method of handling elemwise operations on scalars, arrays, scalar+array, by the Elemwise metagraph. The metagraph generates smaller, simpler graphs, which allows the pipeline to go faster. The tests on my machine go from 43s to 38s, although there is a good deal of variability from one run to another.


myia/prim/py_implementations.py
```diff
@@ -571,6 +572,10 @@ def broadcast_shape(shpx, shpy):
             shp.append(b)
         elif b == 1:
             shp.append(a)
+        elif a == ANYTHING:
+            shp.append(b)
+        elif b == ANYTHING:
+            shp.append(a)
```
abergeron: I'm not sure about this.  ANYTHING should never be encountered during actual execution so it feels out of place here.

breuleux: Yeah, that's a valid point. It's a code reuse thing: it is useful for the inferrer to perform this operation on ANYTHING, and during execution, well, these paths won't be entered anyway. Do you think it would be better to duplicate this functionality?

abergeron: On one hand I don't like duplicating and on the other hand this also feels wrong.  So I'm not sure what the perfect solution here.

We might as well go with this for now, it shouldn't be super hard to split it up later.

PR #168, opened 2019-02-26
Gradient on lists, improvements to manager and optimizations

breuleux:
This adds two tests checking that gradients work on lists, as well as a few missing optimizations and a rewrite of several of the manager's statistics to be lazier.


abergeron:
Looks ok.

PR #169, opened 2019-03-04
Misc improvements (nonies)

breuleux:
This improves tracebacks and adds the tanh primitive.

### DIFF
--- a/myia/abstract/prim.py
+++ b/myia/abstract/prim.py
@@ -430,6 +430,7 @@ uniform_prim(P.scalar_log)(py.scalar_log)
 uniform_prim(P.scalar_sin)(py.scalar_sin)
 uniform_prim(P.scalar_cos)(py.scalar_cos)
 uniform_prim(P.scalar_tan)(py.scalar_tan)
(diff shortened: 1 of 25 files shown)


myia/abstract/prim.py
```diff
@@ -86,7 +89,7 @@ def __init__(self, infer):
 def standard_prim(*prims):
     """Decorator to define and register a StandardInferrer."""
     def deco(fn):
-        xinf = StandardInferrer.partial(infer=fn)
+        xinf = StandardInferrer.partial(prim=prims[0], infer=fn)
```
abergeron: As far as I can see the only spot that uses multiple primitives is `P.identity` and `P.return_`.  Maybe that could just be some special case and we should enforce single-argument here.

All in all this doesn't make a great difference since it's almost always use for a single primitive, but the code here just feels wrong at first glance.

breuleux: Yeah, fair point, I'll enforce single argument.

breuleux: I made the change.

PR #175, opened 2019-03-25
Structure for backends

abergeron:
This provides a structure for backends with a uniform interface.  To make it work I had to remove the debug backend and so there is only the NNVM backend for now, but there should be more to come soon.


myia/api.py
```diff
@@ -25,10 +25,16 @@ class MyiaFunction:
...
+    def __init__(self, fn, specialize_values=[],
+                 backend=None, backend_options=None):
         """Initialize a MyiaFunction."""
         self.fn = fn
         self.specialize_values = set(specialize_values)
+        pp = standard_pipeline.configure({
+            'compile.backend': backend,
+            'compile.backend_options': backend_options
+        })
+        self.pip = pp.make()
```
breuleux: It may cause issues if `make` is called here, because that means all recompiles of the function will reuse the same manager, the same inferrer, etc. and I'm not confident that'll always work.

myia/compile/backends/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+        arguments of the same type and number as the root graph.
+        """
+        raise NotImplementedError('compile')
+
+    def from_numpy(self, a):
+        """Convert a numpy ndarray to a backend value."""
+        raise NotImplementedError("from_numpy")
+
+    def to_numpy(self, v):
+        """Convert a backlend value to a numpy.ndarray."""
```
breuleux:
```suggestion
        """Convert a backend value to a numpy.ndarray."""
```

PR #177, opened 2019-03-27
Overload wrapper

breuleux:
Adds the possibility to wrap calls to an overloaded function with a wrapper function, which receives the method to dispatch to as its first argument (or None, if there is no method to dispatch to).

Additionally, if the `initial_state` function is provided, `bootstrap=True` is implied and top level invocations will initialize a state. The state will be kept through recursive invocations through `self` but each top level invocation creates a new state. This allows things like keeping track of seen objects.

Overload also now automatically receives a name based on the first function or wrapper to be registered to it.


abergeron:
It doesn't appear to forward the docstring from what I can see.

PR #178, opened 2019-04-01
PyTorch backend

abergeron:
This is a backend using pytorch for execution.

Some things are not supported (like unsigned ints) and other are less optimal than they could be (since we use the python runtime interface directly).

If there is a API released for torchscript, it might be interesting to build sub-graphs out of that in the future.  The other option might be to synthesize python function out of the graph structure and parse those using torchscript, but that seems sub-optimal.

### DIFF
--- a/myia/compile/backends/pytorch.py
+++ b/myia/compile/backends/pytorch.py
@@ -76,6 +76,7 @@ def pytorch_array_map(op):
         impl = simple_mapping[fn]
     else:
         raise NotImplementedError(f'array_map of {fn}')
+
     def _impl(*args):
         return (impl(*args),)
     return _impl, op.inputs[2:]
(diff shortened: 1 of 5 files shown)


myia/compile/backends/pytorch.py
```diff
@@ -0,0 +1,191 @@
...
+
+def type_to_pytorch_type(t):
+    """Map myia types to pytorch types."""
+    if t not in _type_map:  # pragma: no cover
+        raise TypeError(f"Unsupported type: {t}")
+    return _type_map[t]
+
+
+simple_mapping = {
+    P.scalar_add: lambda a, b: a + b,
```
breuleux: Could we just use `torch.add` instead? Ditto for the rest of the mappings.

abergeron: `torch.add` doesn't actually add two tensors, it just add a scalar to a tensor.  Same for the rest.

PR #179, opened 2019-04-02
Add debug module

breuleux:
A `debug` module now sits in the top level alongside `myia` and `tests`, formerly in a different repository called `myia_debug`, but that was causing some problems. `debug` mostly contains code pertaining to displaying graphs and data using buche. Running `buche pytest` instead of `pytest` will use buche to display progress and an interactive interpreter when breakpoints are hit.

There is also a utility that can be called with `python -m debug`, see `debug/README.md` for more information.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,6 @@ env:
   - TEST_SUITE=unit
 script:
   - if [[ $TEST_SUITE == "static" ]]; then flake8 && pydocstyle myia; fi
-  - if [[ $TEST_SUITE == "unit" ]]; then pytest --cov=myia/ --cov=tests/ tests; fi
+  - if [[ $TEST_SUITE == "unit" ]]; then pytest --cov=./ tests; fi
 after_success:
   - if [[ $TEST_SUITE == "unit" ]]; then codecov; fi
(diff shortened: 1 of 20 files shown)


myia/compile/backends/__init__.py
```diff
@@ -36,7 +36,7 @@ def loader():
 }
 
 # This is used as the default backend and options if None is provided
-default_name = 'nnvm'
+default_name = 'pytorch'
```
abergeron: Is this intentional?  Not that I'm against it necessarily.

.travis.yml
```diff
@@ -15,6 +15,6 @@ env:
   - TEST_SUITE=unit
 script:
   - if [[ $TEST_SUITE == "static" ]]; then flake8 && pydocstyle myia; fi
-  - if [[ $TEST_SUITE == "unit" ]]; then pytest --cov=./ tests; fi
+  - if [[ $TEST_SUITE == "unit" ]]; then pytest --cov=myia/ --cov=tests/ tests; fi
```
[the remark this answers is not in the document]
breuleux: Ohh, I forgot about that file. I'll change that instead.

myia/compile/backends/__init__.py
breuleux: It was to get rid of the annoying warnings due to the `pytest.ini` change, but they pop up anyway, so that's probably pointless.

abergeron:
You also have to update codecov.yml so that codecov is not unhappy.

breuleux:
The tests seem to pass anyway?

abergeron:
Well then there is no problem.

PR #180, opened 2019-04-03
Relay backend

abergeron:
This add a backend implementation using Relay.  It may not be in as good a shape as the other two (consider it beta), but having it in tree means it will not degrade to the point of irrelevance.

Also, I've added a bit of `pragma: no cover` for things I couldn't trigger, but added exceptions to those branches so that if future code triggers those, we can remove the exception and add a test.


abergeron:
Coverage was bad because it skipped the relay backend due to the nnvm version that was too old.  I bumped it.

myia/compile/backends/relay.py
```diff
@@ -0,0 +1,370 @@
...
+def to_relay_type(self, a: object):  # pragma: no cover
+    raise ValueError("Unknown type:", build_type(a))
+
+
+def ashape(node):
+    """Make sure shape isn't None, that makes relay crash later."""
+    shp = node.shape
+    if shp is None:  # pragma: no cover
+        raise RuntimeError("You found a way to trigger this, please report it")
+        shp = ()
```
breuleux: Is `shp = ()` a legitimate thing to do if this is triggered? If shape is None, this is not an array.

myia/compile/backends/relay.py
```diff
@@ -0,0 +1,370 @@
...
+    P.scalar_div: relay.op.divide,
+    P.scalar_mod: relay.op.mod,
+    P.scalar_pow: relay.op.power,
+    P.scalar_floor: relay.op.floor,
+    P.scalar_uadd: lambda x: x,
+    P.scalar_usub: relay.op.negative,
+    P.scalar_exp: relay.op.exp,
+    P.scalar_log: relay.op.log,
+    # This is not right tangent vs hyperbolic tangent
+    # P.scalar_tan: relay.op.tanh,
```
breuleux:
```suggestion
```

myia/compile/backends/relay_helpers.py
```diff
@@ -0,0 +1,94 @@
...
+        mod[gv] = funcs[gv]
+
+    return mod
+
+
+def optimize(mod):
+    """Optimize all the functions in a module.
+
+    Modules are the only mutable piece of Relay.  We write an
+    optimization pass over the module which destructably updates each
```
breuleux: destructively?

myia/abstract/utils.py
```diff
@@ -145,6 +150,42 @@ def build_type(self, x: AbstractType):
     return dtype.TypeType
 
 
+# A variant of built_dtype with precise function types.
+# Use for the .type attributes of nodes
+
+
+@build_type.variant
+def build_type_fn(self, x: AbstractFunction):
```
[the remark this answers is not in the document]
abergeron: Yes it had some issues because `build_type` is used in a number of places and most of those don't have the correct information for complete function types.

myia/compile/backends/relay.py
abergeron: Is shape is None, then it is a scalar.  But scalars are defined as 0-dim arrays for Relay.

We may want to do something differently eventually once there is better support for actual scalars.

myia/compile/backends/relay_helpers.py
abergeron: This means that the functions are modified in-place, just like our optimizations passes.

breuleux: Oh, haha, sorry, I meant to point out the typo, but I suppose that was a really confusing way to do that.

abergeron: fixed

PR #181, opened 2019-04-05
Misc improvements (undecies)

breuleux:
Just a few things that I might as well push now.


myia/opt/lib.py
```diff
@@ -621,8 +624,8 @@ def inline(optimizer, node, equiv):
             if g.recursive:
                 return node
 
-        clone = GraphCloner(total=False)
-        clone.add_clone(g, node.graph, args)
+        clone = GraphCloner(inline=(g, node.graph, args), total=False)
+        # clone.add_clone(g, node.graph, args)
```
abergeron: This seems like a leftover.

breuleux: It certainly is!

PR #182, opened 2019-04-05
Disable incorporate optimizations

abergeron:
We will probably re-enable them after a rewrite to make them safer, but for now, they interfere with the implementation of more complex models with reasonable performance.


abergeron:
This confounds me.  I've never had a local failure with this branch, on multiple machines and multiple systems/versions.  I've tried reproducing the travis environment and it still passes.  I also tried to have the tests run on a loop overnight and still no failures.

I really have no idea why it fails, and no way to debug this.

myia/compile/backends/__init__.py
```diff
@@ -109,13 +109,25 @@ def to_dlpack(self, v):
...
+
     def convert_value(self, v, t):
         """Convert a value to the appropriate backend representation."""
         if ismyiatype(t, dtype.Number):
             return self.from_scalar(v, t)
         elif ismyiatype(t, dtype.Tuple):
             return tuple(self.convert_value(v, t)
                          for v, t in zip(v, t.elements))
+        # elif ismyiatype(t, dtype.Array):
+        #    return self.from_numpy(v)
```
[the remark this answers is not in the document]
abergeron: Because it is now unused.  Perhaps it should be uncommented with an exception though.

myia/compile/backends/relay.py
```diff
@@ -42,7 +42,10 @@ def to_relay_type(self, a: AbstractArray):
 
 @overload  # noqa: F811
 def to_relay_type(self, a: AbstractFunction):
-    return self(a.get_unique())
+    sings = list(self(sing) for sing in a.get_sync())
+    for sing in sings[1:]:
+        assert sing == sings[0]
+    return sings[0]
```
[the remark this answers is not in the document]
abergeron: get_unique's purpose is not only to build signatures out of types.  I'm not sure if that would be safe to do.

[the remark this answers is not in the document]
breuleux: Yeah I misread, forget about that.

abergeron:
As a hailmary I'm trying to re-run the tests on a new master.

abergeron:
I found the problem and it comes from the merge with master.  I'll update this later with a fix.

abergeron:
This still doesn't work, but at least we know why now.

The problem comes from the fact that partial are recursively typed and the type machinery is not yet ready to deal with that.  In the confusion it thinks a call has multiple types and errors out that way.

abergeron:
I've rebased this on the current master and fix the small issues that popped up.  It should now pass all tests and be ready.

.travis.yml
```diff
@@ -16,5 +16,7 @@ env:
 script:
   - if [[ $TEST_SUITE == "static" ]]; then flake8 && pydocstyle myia; fi
   - if [[ $TEST_SUITE == "unit" ]]; then pytest --cov=./ tests; fi
+  - if [[ $TEST_SUITE == "unit" ]]; then pytest tests; fi
+  - if [[ $TEST_SUITE == "unit" ]]; then pytest tests/test_grad.py::test_list_for; fi
```
breuleux: What are these for?

myia/opt/lib.py
```diff
@@ -702,6 +702,15 @@ def replace_applicator(optimizer, node, equiv):
 ##################
 
 
+def check_uses(g):
+    """Returns True if a graph a only one usage."""
```
breuleux: * The function's name could be clearer, e.g. `check_used_once(g)`
* The docstring is not worded correctly.
* I believe this function is equivalent to `sum(mng.graph_users[g].values()) == 1`.

.travis.yml
abergeron: Oh shit, those are leftovers from when I tried to figure out the problem.  I'll remove them.

myia/opt/lib.py
abergeron: * Yes
* Yes
* Probably

abergeron: It is equivalent so I swapped the implementation.

PR #183, opened 2019-04-08
Union types

breuleux:
This implements part of the union types feature, namely `AbstractUnion` and support for it in the `switch` statement. There are still issues, namely supporting `AbstractUnion`, `hastype` and `unsafe_static_cast` in the backend, as well as recursive ADTs which will eventually come in a different PR.

### DIFF
--- a/myia/abstract/utils.py
+++ b/myia/abstract/utils.py
@@ -26,6 +26,7 @@ from .data import (
     AbstractClass,
     AbstractJTagged,
     AbstractUnion,
+    abstract_union,
     TrackDict,
     VirtualFunction,
     GraphFunction,
@@ -244,6 +245,11 @@ def abstract_clone(self, x: AbstractClass, *args):
 
 
 @overload  # noqa: F811
+def abstract_clone(self, x: AbstractUnion, *args):
+    return abstract_union([self(y, *args) for y in x.options])
+
+
+@overload  # noqa: F811
 def abstract_clone(self, x: AbstractJTagged, *args):
     return AbstractJTagged(self(x.element, *args))
 
@@ -309,6 +315,11 @@ async def abstract_clone_async(self, x: AbstractClass):
         x.methods,
         await self(x.values)
     )
+
+
+@overload  # noqa: F811
+async def abstract_clone_async(self, x: AbstractUnion):
+    return abstract_union([await self(y) for y in x.options])
 
 
 @overload  # noqa: F811
--- a/myia/abstract/utils.py
+++ b/myia/abstract/utils.py
@@ -387,11 +387,15 @@ async def concretize_abstract(self, x: Pending):
 
 @overload  # noqa: F811
 async def concretize_abstract(self, r: Reference):
-    return Reference(
-        r.engine,
-        r.node,
-        await self(r.context)
-    )
+    # There's a bit of a messup here where we can't build Reference using the
+    # constructor because the node in the reference is sometimes unmanaged.
+    # That's likely because of redirections made by the union code in switch.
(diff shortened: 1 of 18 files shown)


myia/abstract/utils.py
```diff
@@ -321,11 +327,15 @@ def abstract_clone(self, x: object, *args):
...
 @overload  # noqa: F811
 async def concretize_abstract(self, r: Reference):
-    return Reference(
-        r.engine,
-        r.node,
-        await self(r.context)
-    )
+    # There's a bit of a messup here where we can't build Reference using the
+    # constructor because the node in the reference is sometimes unmanaged.
+    # That's likely because of redirections made by the union code in switch.
```
abergeron: That doesn't sound pleasant for the rest of the code either.

Is there really no way to fix this?

myia/compile/transform.py
```diff
@@ -9,6 +9,7 @@
 def set_types(graph, argspec, outspec, pipeline):
     """Re-infer and set all the types for the nodes."""
     pipeline.resources.inferrer.infer(graph, argspec, outspec, clear=True)
+    graph.manager.keep_roots()
```
abergeron: Shouldn't this be done before inference?

myia/pipeline/steps.py
```diff
@@ -549,6 +549,15 @@ def convert_arg(self, arg, orig_t: AbstractArray, backend):
...
 
 
+@overload  # noqa: F811
+def convert_arg(self, arg, orig_t: AbstractUnion, backend):
+    for opt in orig_t.options:
+        try:
+            return self(arg, opt, backend)
+        except TypeError:
+            continue
+
```
abergeron: Two things:
- I don't think that having Union types for top-level arguments is something we want to support.
- If we do keep it, there should be an error when none of the options match.

myia/abstract/utils.py
breuleux: The issue is that the constructor of Reference does some extra things such as making sure that the context matches the node's graph, which requires the manager, but in some cases some unmanaged graphs may linger in various caches, and we'll try and fail to copy them through this. There are various solutions:

* Try to eliminate these references since they are probably unnecessary, but I believe that would probably be best addressed by a rewrite of the specializer.
* Simplify the constructor of Reference to not do that and use a helper function to do the context matching when necessary. As it turns out, most of the code already uses `engine.ref(...)` instead of `Reference(engine, ...)`, so I could simply move some code from the constructor of `Reference` to `engine.ref`.

myia/compile/transform.py
breuleux: The extra graph pops up during the inference process. And come to think of it, I figured out why this happens: the new version of `switch` explicitly calls `engine.pipeline.resources.convert(bool)`, because the condition to `if` is usually wrapped with `bool`, and `switch` wants to skip over it to find the real operation. I will check if I can find a better way to do this, then that call won't be necessary anymore.

myia/pipeline/steps.py
breuleux: We do want to support this: if the function operates on trees, one of the arguments will be a tree, and a tree (let's say a binary tree) would have the type `Tree :: Union[Nil, Cons[left :: Tree, right :: Tree]]`.

Good catch about the error, though, that's definitely something I forgot.

myia/compile/transform.py
abergeron: Ok

myia/pipeline/steps.py
abergeron: Ok about support.

PR #195, opened 2019-05-02
Support cloning recursive data

breuleux:
This ameliorates `broaden` and `concretize_abstract` so that they can create and handle certain recursive data structures, in particular a `PartialApplication` that receives itself as one of its arguments.

When rebased on top of these changes, #182 should pass the tests.

### DIFF
--- a/myia/abstract/utils.py
+++ b/myia/abstract/utils.py
@@ -3,6 +3,7 @@
 from dataclasses import dataclass
 from functools import reduce
 from itertools import chain
+from types import GeneratorType, AsyncGeneratorType
 
 from .. import dtype
 from ..utils import overload
@@ -212,15 +213,30 @@ class CheckState:
 @overload.wrapper(initial_state=lambda: None)
 def abstract_check(__call__, self, x, *args):
     """Check that the given object has a certain property."""
+    def proceed():
+        if hasattr(x, prop):
+            return getattr(x, prop) is x
+        elif __call__(self, x, *args):
+            if isinstance(x, (AbstractValue, Context, Reference)):
+                setattr(x, prop, x)
+            return True
+        else:
+            return False
+
     prop = self.state.prop
-    if hasattr(x, prop):
-        return getattr(x, prop) is x
-    elif __call__(self, x, *args):
-        if isinstance(x, (AbstractValue, Context, Reference)):
-            setattr(x, prop, x)
-        return True
+    cache = self.state.cache
+
+    try:
+        rval = cache.get(x, None)
+    except TypeError:
+        return proceed()
+
+    if rval is None:
+        cache[x] = True
+        cache[x] = proceed()
+        return cache[x]
     else:
-        return False
+        return rval
 
 
 @overload  # noqa: F811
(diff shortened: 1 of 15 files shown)


myia/abstract/utils.py
```diff
@@ -195,20 +197,169 @@ def build_type_fn(self, x: Function):
...
+def abstract_clone(__call__, self, x, *args):
     """Clone an abstract value."""
+    def proceed():
+        if isinstance(x, AbstractValue) and x in cache:
+            return cache[x]
+        result = __call__(self, x, *args)
+        if not isinstance(result, GeneratorType):
+            return result
+        constructor = None
+        while True:
```
abergeron: Why does this need to loop while the async version doesn't?

myia/abstract/utils.py
```diff
@@ -195,20 +197,169 @@ def build_type_fn(self, x: Function):
...
+    cache: dict
+    prop: str
+
+
+@overload.wrapper(initial_state=lambda: None)
+def abstract_check(__call__, self, x, *args):
+    """Check that the given object has a certain property."""
+    def proceed():
+        if hasattr(x, prop):
+            return getattr(x, prop) is x
```
abergeron: This doesn't only check that it has the property but also that the value is x.

I feel like this is too specific to `is_concrete`, but I could be wrong.

myia/abstract/utils.py
```diff
@@ -195,20 +197,169 @@ def build_type_fn(self, x: Function):
...
+        elif __call__(self, x, *args):
+            if isinstance(x, (AbstractValue, Context, Reference)):
+                setattr(x, prop, x)
```
abergeron: This also feels specific to `is_concrete`.

myia/abstract/utils.py
```diff
@@ -272,15 +436,54 @@ def abstract_clone(self, x: object, *args):
...
+
+@overload  # noqa: F811
+async def abstract_clone_async(self, x: AbstractScalar):
     return AbstractScalar(await self(x.values))
 
 
 @overload  # noqa: F811
 async def abstract_clone_async(self, x: AbstractFunction):
-    return AbstractFunction(*(await self(x.get_sync())))
+    yield (yield AbstractFunction)(*(await self(x.get_sync())))
```
abergeron: I don't get why we need two yields here.  I'm not saying it is wrong, I just don't get it.

myia/abstract/utils.py
breuleux: Neither really "needs" to loop, the issue is that async generators can't return a value, so I have to treat the second yield as the return value, whereas normal generators raise `StopIteration` with the function's return value. I should probably remove the loop anyway, since it's not clear what entering it twice is supposed to mean.

myia/abstract/utils.py
breuleux: `is_broad` does the same thing.

myia/abstract/utils.py
breuleux: Yeah, maybe, but I'm not sure what other criterion to put. I guess try/catch would work, but some classes might raise some exotic exception here.

myia/abstract/utils.py
breuleux: See my other comment, async generators cannot return values.

abergeron: Ok.

myia/abstract/utils.py
abergeron: I see that you skip over the problem elsewhere by resetting the wrapper to `None`.

I guess that's ok but maybe it should be documented better (especially since the docs do not match the code currently).

myia/abstract/utils.py
abergeron: You should remove the loop from the non-async variant since it should only ever loop once and it would be clearer with an explicit second invocation of `send()`, possibly followed by an assert  that we don't continue past that.

myia/abstract/utils.py
breuleux: Oh, that's a mistake, actually. `wrapper` should not be set to `None` because that would turn off the cache and might cause an infinite loop. I'll fix that.

abergeron: All that's left is to fix the docstring for abstract_check.

breuleux: I think the docstring is fine, for a wide sense of "property". `validate` does not check for an attribute on an object, but it does check the "property" of being valid. Do you have a better wording in mind?

abergeron: I get what you mean, but property evokes the python concept for me.  I'm not sure what would be a better wording.

breuleux: Check that a certain predicate applies?

abergeron: That sounds good, yes.

PR #199, opened 2019-05-13
Add RNN and LSTM examples

breuleux:
This adds `examples/rnn.py` and `examples/lstm.py`, along with a few fixes that make them run properly. Only `rnn.py` is added to the test suite for this PR, since the LSTM is currently a bit too costly.

### DIFF
--- a/examples/lstm.py
+++ b/examples/lstm.py
@@ -20,6 +20,7 @@ from myia.debug import traceback  # noqa
 ###########
 
 
+dtype = 'float32'
 device_type = 'cpu'
 # device_type = 'cuda'  # Uncomment to run on the gpu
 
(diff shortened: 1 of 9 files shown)


examples/lstm.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+# def test_run():
+#     """Run the model.
+
+#     This function is run automatically in the test suite to check against
+#     regressions, so keep a low number of narrow layers to make sure it runs
+#     quickly.
+#     """
+#     run_helper(1, 1, 5, (10, 3))
```
abergeron: Maybe a comment to explain that we don't test LSTM yet.  Or better, make it fail with xfail.

breuleux: OK

PR #201, opened 2019-05-30
Misc improvements (terdecies)

breuleux:
This mainly removes set support from intern and modifies Possibilities and Union not to use sets. `abstract_clone` does not handle `Context` or `Reference` anymore.

### DIFF
--- a/myia/abstract/data.py
+++ b/myia/abstract/data.py
@@ -438,7 +438,7 @@ class AbstractUnion(AbstractStructure):
     def __init__(self, options):
         """Initialize an AbstractUnion."""
         super().__init__({})
-        self.options = frozenset(options)
+        self.options = tuple(options)
 
     def children(self):
         """Return the set of options."""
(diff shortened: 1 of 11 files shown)


myia/abstract/data.py
```diff
@@ -438,7 +439,7 @@ class AbstractUnion(AbstractStructure):
     def __init__(self, options):
         """Initialize an AbstractUnion."""
         super().__init__({})
-        self.options = frozenset(options)
+        self.options = tuple(options)
```
abergeron: Maybe a comment about how this can't be hashed until the type is complete.

myia/abstract/data.py
```diff
@@ -459,7 +460,7 @@ def abstract_union(options):
             opts += option.options
         else:
             opts.append(option)
-    opts = frozenset(opts)
+    opts = tuple(opts)
```
abergeron: Here too.

myia/abstract/utils.py
```diff
@@ -810,14 +757,14 @@ def chk(a):
...
-    if x1.issuperset(x2):
+    if set(x1).issuperset(set(x2)):
         return x1
     if forced:
         raise MyiaTypeError(
             'Additional possibilities cannot be merged.'
         )
     else:
-        return Possibilities(x1 | x2)
+        return Possibilities(x1 + x2)
```
abergeron: And possibly here?

PR #203, opened 2019-05-31
Add reroute method to inferrers

breuleux:
The inferrer for a primitive can now define the `reroute` method to return a replacement Reference (which may point to a new node or subgraph) given a list of argrefs. This is used by `getattr`, `resolve` and the new primitive `user_switch`.

`user_switch` is output by the parser instead of `switch` for `if` statements and the like. It requires both branches to be functions, and it only defines `reroute`, which means it will always disappear after specialization.

If the condition has the form `hastype(x, t)`, `user_switch` wraps each branch to see the correctly typed version of `x`, and returns a new Reference. If the condition is not a boolean, `user_switch` adds a conversion to bool.

`ConditionalContext` is removed, which removes a potential source of bugs.


myia/abstract/prim.py
```diff
@@ -266,7 +266,7 @@ def _prim_or_graph(afn):
     if case == 'class':
         # Get field from Class
         if item_v in data.attributes:
-            return await on_dcattr(data, item_v)
+            return 'rval', await on_dcattr(data, item_v)
```
abergeron: It appears you've changed the return signature, but not the documentation.

PR #212, opened 2019-06-07
ArrayWrapper to allow backend output as a valid parameter to Myia functions

ethancaballero:
ArrayWrapper (wrapper for arrays) to allow backend output as a valid parameter to Myia functions.

Described and assigned in this issue:
https://github.com/mila-iqia/myia/issues/204


examples/mlp.py
```diff
@@ -110,15 +129,16 @@ def cost(model, x, target):
...
-@myia(backend='pytorch', backend_options={'device': device_type})
+@myia(backend=backend, backend_options=backend_options, return_backend=True)
 def step(model, x, y):
     """Returns the loss and parameter gradients."""
     # value_and_grad will return cost(model, x, y) and dcost(...)/dmodel.
     # The 'model' argument can be omitted: by default the derivative wrt
     # the first argument is returned.
-    return value_and_grad(cost, 'model')(model, x, y)
+    cost, dmodel = value_and_grad(cost, 'model')(model, x, y)
+    return cost, model - dmodel * scalar_cast(0.01, f32)
```
breuleux: The dtype is not necessarily f32 (see the `dtype = 'float32'` line). `myia.dtype.DTYPE_TO_MTYPE[dtype]` should return the correct type to cast to, although you might need to assign that to a global variable outside of the step function.

examples/mlp.py
```diff
@@ -13,16 +13,35 @@
...
+
 device_type = 'cpu'
 # device_type = 'cuda'  # Uncomment to run on the gpu
 
+if backend == 'pytorch':
+    backend_options = {'device': device_type}
+elif backend == 'nnvm':
+    backend_options = {'target': device_type, 'device_id': 0}
+elif backend == 'relay':
+    backend_options = {'target': device_type, 'device_id': 0}
```
breuleux: I'd put these in a dictionary, e.g. `backend_options = {'pytorch': {'device': device_type}, ...}`, and index it with `backend` in the decorator. It's a bit more flexible and the error will be a little nicer if there's a typo in the backend name.

myia/abstract/infer.py
```diff
@@ -27,6 +27,23 @@
     bind, type_to_abstract
 
 
+class ArrayWrapper:
```
breuleux: Add an import in `myia/abstract/__init__.py` so that we can do `from myia.abstract import ArrayWrapper`.

myia/pipeline/steps.py
```diff
@@ -23,6 +23,8 @@
 
 from .pipeline import pipeline_function, PipelineStep
 
+from ..abstract.infer import ArrayWrapper
```
breuleux: Import this from `..abstract`.

myia/pipeline/steps.py
```diff
@@ -595,7 +603,14 @@ def convert_result(self, arg, orig_t, vm_t: AbstractScalar, backend):
...
 @overload  # noqa: F811
 def convert_result(self, arg, orig_t, vm_t: AbstractArray, backend):
-    return backend.to_numpy(arg)
+    a = backend.to_numpy(arg)
+    if isinstance(backend, NumpyChecker):
+        a = ArrayWrapper(
+            a,
+            dtype.type_to_np_dtype(orig_t.element.dtype()),
+            orig_t.values[SHAPE]
+            )
```
breuleux: This is the wrong check: you only want to return an `ArrayWrapper` when the `return_backend` option is True. This means that `return_backend` should be checked here. You could add it as an extra parameter to `convert_result`, which is called from `myia.pipeline.steps.Wrap`, or maybe tuck it in the `backend` object.

tests/test_api.py
```diff
@@ -15,6 +15,11 @@
 from .common import Point, Point3D, i64, f64, to_abstract_test, ai64_of, \
     ai32_of, af64_of
 
+from myia.abstract.infer import ArrayWrapper
+from myia.compile import load_backend
+
+from dataclasses import dataclass
```
breuleux: Put stdlib imports at the top, and imports from myia above the local import from `.common`.

tests/test_specialize.py
```diff
@@ -91,6 +92,14 @@ def run_test(args):
                     raise verr
 
                 result_final = res['output'](*args)
+
+                if isinstance(result_final, ArrayWrapper):
+                    result_final = result_final.array
+                elif isinstance(result_final, tuple):
+                    result_final = tuple(a.array if isinstance(a, ArrayWrapper) else a for a in result_final)
+                elif isinstance(result_final, list):
+                    result_final = [a.array if isinstance(a, ArrayWrapper) else a for a in result_final]
```
breuleux: This should be unnecessary, because `return_backend` is False (see my other comment about this).

myia/api.py
```diff
@@ -1,10 +1,20 @@
...
 from .pipeline import standard_pipeline
-from .utils import keyword_decorator
+from .utils import keyword_decorator, overload
+
+from .abstract import AbstractTuple, AbstractList, AbstractClass, \
+    AbstractArray, TYPE, AbstractScalar, AbstractUnion
+
+from .abstract.infer import ArrayWrapper
+
+from myia.compile.backends import load_backend
```
breuleux: Import ArrayWrapper from .abstract, use a relative import for compile, and compress the whitespace.

myia/api.py
```diff
@@ -100,3 +107,97 @@ def myfun2(cond, x, y):
...
+    if isinstance(arg, np.ndarray):
+        arg = ArrayWrapper(backend.from_numpy(arg), arg.dtype, arg.shape)
+    backend.check_array(arg.array, et)
+    return arg
+
+
+"""
+TODO: AbstractUnion overload of _convert_arg_init().
+
+This overliad will look similar to AbstractUnion overload
```
breuleux:
```suggestion
This overload will look similar to AbstractUnion overload
```

tests/test_api.py
```diff
@@ -314,3 +326,165 @@ def f():
...
+
+
+__W = np.array(
+    [
+        [0.22490819, -0.71516967, -0.6626606],
+        [-0.14859799, 0.924496, 0.4032511]
+    ]
+)
+
+__b = np.array([[-0.00300545, 0.10826713, -0.29246148]])
```
breuleux: This doesn't seem to be used anywhere.

tests/test_api.py
```diff
@@ -314,3 +326,165 @@ def f():
...
+    name, options = request.param
+    return (name, options)
+
+
+__W = np.array(
+    [
+        [0.22490819, -0.71516967, -0.6626606],
+        [-0.14859799, 0.924496, 0.4032511]
+    ]
+)
```
breuleux: I don't know if the exact values for this matrix are important, but there are a few premade matrices in `tests.common`. You can use e.g. `tests.common.MA(2, 3)` instead of `__W`, for a 2x3 matrix.

examples/mlp.py
```diff
@@ -12,17 +12,41 @@
...
+from myia.dtype import DTYPE_TO_MTYPE
 
 ###########
 # Options #
 ###########
 
 
 dtype = 'float32'
+
+mtype = DTYPE_TO_MTYPE[dtype]
```
abergeron: This doesn't seem like it's used anywhere.

myia/api.py
```diff
@@ -100,3 +107,97 @@ def myfun2(cond, x, y):
...
+AbstractUnion overload from convert_arg() at time of this commit:
+https://github.com/mila-iqia/myia/blob/
+c350b341f52d2d0e3dc1e5ab1d890103a45b60c9/
+myia/pipeline/steps.py#L528-L537
+
+Current AbstractUnion overload from convert_arg():
+(Note: might have been moved to different location of codebase
+since this comment was written/committed):
+https://github.com/mila-iqia/myia/blob/master/myia/pipeline/steps.py#L528-L537
+"""
```
abergeron: I don't think this one will be needed since you can't pass in a value that is of union type.  I don't see much harm in leaving the comment though.

tests/test_abstract.py
```diff
@@ -28,6 +29,14 @@ def test_to_abstract():
     assert to_abstract(inst) == expected
 
 
+def test_numpy_scalar_to_abstract():
+    s1 = AbstractScalar({VALUE: 2, TYPE: i64})
+    assert to_abstract_test(np.int64(2)) == s1
```
abergeron: int64 is probably a bad test since it's de default type for python ints.  Maybe try int16?

tests/test_api.py
```diff
@@ -314,3 +326,155 @@ def f():
...
+    with pytest.raises(TypeError):
+        _convert((1,), (i64, i64))
+    with pytest.raises(TypeError):
+        _convert((1, 2, 3), (i64, i64))
+    with pytest.raises(TypeError):
+        _convert((1, 2, 3), [i64])
+    with pytest.raises(TypeError):
+        _convert(pt3, Point(i64, i64))
+    with pytest.raises(TypeError):
+        _convert(1, Bool)
```
abergeron: These tests seem weird to me because _convert_arg_init() should only be used for to_device(), which can't really do meaningful type checking.  So why test that type checking is done?

ethancaballero: @abergeron 
I added these tests in order cover all the
```
if not isinstance(*, *):
            raise TypeError(f'Expected *')
```
in _convert_arg_init().

Do you want me to remove all the
```
if not isinstance(*, *):
            raise TypeError(f'Expected *')
```
in _convert_arg_init().

abergeron: Well you did just remove them so I guess that's it.

PR #215, opened 2019-06-20
Add coverage for GPU parts

abergeron:
Fixes a number of issues that were present in the GPU parts of the backends.

This also updates flake8 checking to use 3.7.7


.flake8
```diff
@@ -0,0 +1,2 @@
+[flake8]
+ignore = W503,W504,E241,F722
```
breuleux: GitHub reports no newline at end of file and it's bugging me.

myia/compile/backends/nnvm.py
```diff
@@ -314,7 +314,7 @@ def convert(self, lst, context):
         target = context.MASK2STR[context.device_type]
         if target == 'cpu':
             nnvm_target = 'llvm'
-        elif target == 'cuda':  # pragma: no cover
+        elif target == 'gpu':
```
breuleux: Why are we using `'gpu'` instead of `'cuda'`?

tests/compile/test_nnvm.py
```diff
@@ -0,0 +1,42 @@
...
+from myia import dtype
+
+try:
+    from myia.compile.backends import nnvm
+except ImportError:
+    pytestmark = pytest.mark.skip(f"Can't import nnvm")
+
+
+@pytest.mark.gpu
+def test_nnvm_backend_exists():
```
breuleux: I find the test name a tad misleading.

myia/compile/backends/nnvm.py
abergeron: Because that's what tvm uses.

PR #216, opened 2019-06-25
Good errors and warnings

ethancaballero:
- Good errors for every unsupported AST node type
- Good error for when a return statement is missing (previous check did not work when there was a while loop)
- Warning when part of the code is disconnected from the output (e.g. print statements) (via either unassigned node or unused assigned node)

- Tests that assert that Errors and Warnings are displayed to look (and be colored) a specified way.


myia/debug/traceback.py
```diff
@@ -115,8 +120,12 @@ def _print_lines(lines, l1, c1, l2, c2, label='', mode='color'):
             prefix = trimmed[:start]
             hl = trimmed[start:end]
             rest = trimmed[end:]
-            eprint(f'{ln}: {prefix}{Fore.RED}{Style.BRIGHT}'
-                   f'{hl}{Style.RESET_ALL}{rest}')
+            if color == 'red':
+                eprint(f'{ln}: {prefix}{Fore.RED}{Style.BRIGHT}'
```
breuleux: Try using `getattr(Fore, color.upper())` instead of special casing the colors.

myia/debug/traceback.py
```diff
@@ -137,15 +146,58 @@ def print_inference_error(error):
...
+
+
+_previous_warning = warnings.showwarning
+
+
+def myia_warning(message, category, filename, lineno, file, line):
+    """Print out MyiaDisconnectedCodeWarning specially."""
+    if category is MyiaDisconnectedCodeWarning:
+        __warning = message
+        print_myia_warning(__warning)
```
breuleux: What is `__warning` for?

myia/parser.py
```diff
@@ -214,6 +231,35 @@ def parse(self) -> Graph:
...
+                        pos = getattr(obj, "about", None)
+                        if pos is None:
+                            break
+                        pos = getattr(pos, "debug", None)
+                        obj = pos
+                    raise MyiaSyntaxError(
+                        "Function doesn't return a value in all cases",
+                        obj.location)
+
+        # TODO: also clean_up/delete/gc any left over nodes from dfs?
```
breuleux: That shouldn't be necessary.

myia/parser.py
```diff
@@ -214,6 +231,35 @@ def parse(self) -> Graph:
...
+
+        diff_cache = self.write_cache - self.read_cache
+        if diff_cache != OrderedSet():
```
breuleux: You don't need to build an `OrderedSet()`, just `if diff_cache:` should work.

PR #219, opened 2019-06-27
Misc improvements (quaterdecies)

breuleux:
This is based on #218 

* Add `Graph.set_flags` and `Graph.has_flags`.
* Add the `reference` flag, currently not used for anything, but relevant in the rewrite of the specializer.
* Minor changes to to the `isomorphic` function. It should work like before.
* Implement `PartialInferrer.reroute`. Incidentally, this allows us to remove some code that's specific to partial in the specializer.

### DIFF
--- a/myia/composite.py
+++ b/myia/composite.py
@@ -35,6 +35,7 @@ def core(fn=None, **flags):
     flags = {
         # This is a function defined in Myia's core
         'core': True,
+        'reference': True,
         **flags,
     }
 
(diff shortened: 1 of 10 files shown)


myia/composite.py
```diff
@@ -617,10 +617,10 @@ def generate_graph(self, args):
 
         gnext = Graph()
         gnext.debug.name = self._decorate_name('lm_body')
-        gnext.flags['ignore_values'] = True
+        gnext.set_flags('reference', 'ignore_values')
```
abergeron: Are those really reference graphs?

breuleux: The graph doesn't have the core flag because we don't want to inline it.

abergeron: Ok

PR #222, opened 2019-06-28
Add support for the raise statement

breuleux:
This adds support for the  statement `raise Exception("message")` in Myia. Only `Exception` is supported for now. It maps to the new primitive `exception`.

A raise statement is treated like a return statement, in the sense that it sets the current graph's return value: `raise x` is equivalent to `return raise_(x)`. The `raise_` primitive has type `AbstractBottom`, which is the type of non-returning expressions, and its Python implementation simply raises an exception. There is no way to catch that exception within Myia, but it will terminate execution.


breuleux:
The PR failed on Jenkins because of a non-deterministic bug in the manager, which is triggered when the graph in `test_infer::test_raise_rec` is cloned right after the parse.

I made the manager deterministic, but then the bug stops happening and I can't reproduce it. I'm going to leave it at that for this PR, but I'll make an issue about the bug itself.

abergeron:
This really makes me want to add support for try/except/finally (which wouldn't be that hard).  But the code as-is is ok, so I'll merge.

PR #225, opened 2019-07-08
Changes before pytorch frontend (Class constructor, Scalar backend fixes, Bool & Nil return, isCompare)

ethancaballero:
**Don't** squash to less than these 4 commits:

- isCompare support
 
- Support for Bool and support for Nil return

- Scalar backend fixes (e.g. array_to_scalar)

- Add constructor for class types

### DIFF
--- a/myia/compile/backends/nnvm.py
+++ b/myia/compile/backends/nnvm.py
@@ -15,7 +15,7 @@ from ..utils import get_outputs
 from ..transform import CompileGraphs, nonlinear_ops
 
 from ...abstract import AbstractArray
-from ...dtype import type_to_np_dtype
+from ...dtype import type_to_np_dtype, Nil
 from ...prim import Primitive, ops as P
 
 
@@ -51,7 +51,7 @@ SIMPLE_MAP = {
 def nnvm_bool_not(c, arg):
     """Implementation of boolean not."""
     t = arg.abstract.dtype()
-    zero = c.make_constant(0, nnvm_type=nnvm_type_map(t))
+    zero = c.make_constant(0, nnvm_type=type_to_np_dtype(t))
     return sym.broadcast_equal(zero, c.ref(arg))
 
 
@@ -125,14 +125,6 @@ COMPLEX_MAP = {
     P.array_reduce: nnvm_array_reduce,
     P.transpose: nnvm_transpose,
 }
-
-
-def nnvm_type_map(type):
-    """Map a numpy type to an NNVM type."""
-    dt = type_to_np_dtype(type)
-    if dt == 'bool':
-        dt = 'uint8'
-    return dt
 
 
 def nnvm_val(val, dtype, ctx):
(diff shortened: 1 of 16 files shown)


ethancaballero:
needs Relay CI fix from @abergeron

myia/compile/backends/nnvm.py
```diff
@@ -382,13 +382,30 @@ def from_numpy(self, a):
 
     def to_scalar(self, v):
         """Convert the NNVM array to a scalar."""
-        return v.asnumpy().item()
+        if hasattr(v, "dtype"):
```
breuleux: That may not be necessary if the `v is None` test is done first.

myia/compile/backends/pytorch.py
```diff
@@ -211,13 +211,32 @@ def from_numpy(self, a):
 
     def to_scalar(self, v):
         """Convert a torch tensor to a scalar."""
-        return v.item()
+        if hasattr(v, "dtype"):
```
breuleux: That may not be necessary if the `v is None` test is done first.

myia/compile/backends/relay.py
```diff
@@ -348,13 +350,30 @@ def from_numpy(self, a):
 
     def to_scalar(self, v):
         """Convert the TVM array to a scalar."""
-        return v.asnumpy().item()
+        if hasattr(v, "dtype"):
```
breuleux: That may not be necessary if the `v is None` test is done first.

myia/composite.py
```diff
@@ -497,6 +497,11 @@ def array_tanh(xs):
     return array_map(scalar_tanh, xs)
 
 
+@core
+def nil_bool(x):
```
breuleux: Needs documentation. Other functions in this file also need documentation.

myia/composite.py
```diff
@@ -561,6 +579,40 @@ def _array_zero(xs):
...
+        g.debug.name = "is_not" if self.do_not else "is_"
+        g.set_flags('core', 'reference')
+        pa = g.add_parameter()
+        pb = g.add_parameter()
+
+        if a != b:
+            g.return_ = g.apply(P.return_, self.do_not)
+        else:
+            cmp_fn = g.apply(P.getattr, pa,
+                             "__ne__" if self.do_not else "__eq__")
```
breuleux: `is` and `is not` should only compare to Bool and None. On any other type, they do identity comparison in Python, and we probably don't want to support it. `is` and `is not` do _not_ call `__eq__` or `__ne__`.

To be clear, at least one of the arguments should be a bool or None, and then we can check if the other is also None or the right boolean. If both arguments are scalars or arrays or anything else, there should be a compile time type error.

abergeron: I think it's ok to make is equivalent to `__eq__` if the types are the same.  Because we are functional this is essentially the same.

breuleux: To do so would make Myia's behavior deviate from Python's. If you write `a is b` instead of `a == b` in Python, you mean to check whether `a` and `b` are the same object, and it would be problematic if Myia didn't do that.

For example, if a user decides to do something like `if x is layers[0].weights: ...` to do something on the first weights layer, running `x == layers[0].weights` instead is going to be both inefficient and potentially erroneous if some other parameter has the same values for some reason (e.g. if they all start out at 0). I think it's best to forbid it with an informative error message.

PR #228, opened 2019-07-10
Improve monomorphizer handling of array_map/array_reduce

breuleux:
This modifies the monomorphizer so that `array_map(f, xs)` will monomorphize f properly when `f` has type `POLY` (using the type of the elements of xs). Same for `array_reduce`.

The monomorphizer will also now pack information about all the possible input types inside the AbstractErrors it produces, and the inferrer will produce a more informative error when attempting to call an AbstractError.


tests/test_monomorphize.py
```diff
@@ -302,11 +323,13 @@ def square(x):
...
 
-    if c:
-        f = square
-    else:
-        f = double
+    def picker(c):
+        if c:
+            return square
+        else:
+            return double
```
abergeron: Did the previous code fail to work?

breuleux: `POLY` is not really an issue if it's not called, e.g. it can sometimes be the type of a dead parameter. That's why I changed validation so that it checks that a node with type `POLY` is not called. With this change, the old version of this test xpassed, which prompted me to look at the graph, and it turns out the way the graph gets structured when the if statement is inside the main function allows the inferrer and monomorphizer to avoid using the `POLY` parameter. In other words, this wasn't actually testing what I wanted to test. Now it does, although the inferrer pass after the first monomorphization now fails before validation, because it actually tries to infer a call to `POLY`. Still, that's the right test.

abergeron: Ok

PR #230, opened 2019-07-15
Frontend Support and PyTorch Frontend v0.0.0

ethancaballero:
Parts I'm unsure about:
- What should doctrings be for PartialCallable and its `__new__` method:
https://github.com/ethancaballero/myia/blob/frontend_pytorch/myia/utils/partial.py#L103-L106

- Should frontend always equal load_frontend('numpy') in typeof in these lines:
https://github.com/ethancaballero/myia/blob/frontend_pytorch/myia/prim/py_implementations.py#L240-L242

- current way I warn this might be too hacky:
"All Arrays returned by Myia Function will still be PyTorch Tensors even if backend is not pytorch and/or arrays were wrapped in ArrayWrappers via to_device"
https://github.com/ethancaballero/myia/blob/frontend_pytorch/myia/api.py#L52-L56


myia/frontends/pytorch.py
```diff
@@ -0,0 +1,345 @@
...
+        is generated by the inferrer or other methods.
+
+        Current default is to return self in order to make it easier for Myia
+        hypermap mapping function to return a different type from its input
+        (especially for pytorch modules and their contents).
+        """
+        return self
+
+
+class AbstractTensor(AbstractArray):
```
breuleux: I would rename this to `AbstractPyTorchTensor`.

myia/frontends/pytorch.py
```diff
@@ -0,0 +1,345 @@
...
+        # TODO: IS RETAIN_GRAD EVEN NEEDED? IDK YET.
+        self.retain_grad = retain_grad
+
+
+'''
+class AbstractParameter(AbstractTensor):
+    """Represents a PyTorch Parameter."""
+
+    # This might end up as an alternative way to log that this is a Parameter.
+'''
```
breuleux: Code blocks should be commented out using `#`, not triple quotes.

myia/frontends/pytorch.py
```diff
@@ -0,0 +1,345 @@
...
+
+
+AT = AbstractTensor(ANYTHING, {SHAPE: ANYTHING})
+
+
+@C.tanh.register(AT)
+@core
+def array_tanh(xs):
+    """Implementation of `array_tanh`."""
+    return P.array_map(P.scalar_tanh, xs)
```
breuleux: You should not have to do this, it should just use the registered `AbstractArray` implementation. If it doesn't, that's a bug in the current system.

myia/frontends/pytorch_functions.py
```diff
@@ -0,0 +1,53 @@
...
+#           compared with original pytorch function that function is        #
+#           replacing via pytorch_*_map in /myia/frontends/pytorch.py       #
+#                                                                           #
+#           I.e. Every function in this should have a pytest test made      #
+#           for it that asserts its Output & Gradients is equal to the      #
+#           pytorch original.                                               #
+#############################################################################
+
+
+# import torch
```
breuleux:
```suggestion
```

tests/frontends/test_pytorch.py
```diff
@@ -0,0 +1,488 @@
...
+
+def test_pytorch_dtype_to_type():
+    with pytest.raises(TypeError):
+        pytorch_dtype_to_type("fake_pytorch_type")
+
+
+# pt_debug_pipeline = standard_debug_pipeline \
+#     .select('parse', 'resolve', 'infer', 'export')
+
+# torch.set_printoptions(precision=8)
```
breuleux: Stray comments.

tests/frontends/test_pytorch.py
```diff
@@ -0,0 +1,488 @@
...
+
+    output = step_debug(model, inp)
+
+    output_expected = torch.Tensor(
+        [[ -5.20440006,   0.74380112,  -8.34329891],
+         [-18.35869789, -54.33729553,   9.38029861]]
+         )
+
+    assert torch.allclose(output, output_expected)
+#"""
```
breuleux: Remove this?

myia/abstract/prim.py
```diff
@@ -623,7 +625,7 @@ class _GetAttrInferrer(Inferrer):
 
 @standard_prim(P.scalar_to_array)
 async def _inf_scalar_to_array(engine, a: AbstractScalar):
-    return AbstractArray(a, {SHAPE: ()})
+    return engine.pipeline.resources.array_class(a, {SHAPE: ()})
```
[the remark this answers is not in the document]
abergeron: The problem is twofold:
1) Inference must know the "right" kind of array class for proper method mapping, which could be solved by what you are proposing.
2) The constant folding must produce a value that the backend understands, which would require to know the backend at a minimum.  But that is a pre-existing problem so we can ignore it for this PR.

myia/composite.py
```diff
@@ -89,7 +89,9 @@ def generate_graph(self, args):
                 p = g.add_parameter()
                 p.debug.name = f'x{i + 1}'
                 if is_array_op and not isinstance(arg, AbstractArray):
-                    p = g.apply(to_array, p)
+                    # XXX Get the right array class somewhere
+                    p = g.apply(to_array, p,
+                                AbstractArray(ANYTHING, {SHAPE: ANYTHING}))
```
breuleux: Fix this?

PR #232, opened 2019-07-16
Add AbstractTaggedUnion and related operations

breuleux:
`AbstractTaggedUnion` is the final type for unions and associates an integer tag to a type. Three new primitives are added to deal with this:

* `hastag(u, tag)` checks whether the tagged union `u` has the given tag.
* `casttag(u, tag)` extracts the value of the tagged union and gives it the type associated to the tag. 
* `tagged(x, tag)` generates a tagged union with x as the value, and the given tag.
  * `tagged(x)`, without the `tag` argument, generates a non-tagged union.

The `erase_class` pipeline pass is renamed to `simplify_types` and transforms `AbstractUnion` into `AbstractTaggedUnion`, `hastype` into `hastag`, and so on.


myia/opt/clean.py
```diff
@@ -18,12 +34,26 @@ def _reabs(self, a: AbstractDict):
...
+    * Replace AbstractClass by AbstractTuple:
+      * Class[x: t, ...] => Tuple[t, ...]
+      * getattr(data, attr) => getitem(data, idx)
+      * make_record(cls, *args) => make_tuple(*args)
+
+    * Replace AbstractUnion by AbstractTaggedUnion:
+      * Union[a, b, c, ...] => TaggedUnion[1 => a, 2 => b, 3 => c, ...]
+      * hastype(x, type) => hastag(x, tag)
+      *                  => bool_or(hastag(x, tag1), hastag(x, tag2), ...)
+      * unsafe_static_cast(x, type) => casttag(x, tag)
```
abergeron: I forgot to document that, but it also replaces AbstractDict by Tuples.

breuleux: I'll add it

PR #236, opened 2019-07-24
Misc improvements (septdecies)

breuleux:


myia/abstract/utils.py
```diff
@@ -694,11 +702,11 @@ def chk(a):
             f'Type mismatch: {type(x1)} != {type(x2)}; {x1} != {x2}'
         )
     else:
-        return __call__(x1, x2, loop, forced, bind_pending)
+        return self.map[type(x1)](self, x1, x2, forced, bind_pending)
```
abergeron: Why are you using an overload wrapper if you just ignore the result of the overload resolution?

myia/abstract/utils.py
```diff
@@ -745,122 +753,128 @@ def amerge(x1: dtype.TypeMeta, x2, loop, forced, bp):
...
 
 @overload  # noqa: F811
-def amerge(x1: object, x2, loop, forced, bp):
+def amerge(self, x1: object, x2, forced, bp):
     if x1 != x2:
         raise TypeMismatchError(x1, x2)
-    return x1
+    else:  # pragma: no cover
+        # This used to be covered before the `x1 is x2` check in amerge
+        return x1
```
abergeron: This case should be relatively easy to cover with something like `amerge([1], [1])`

myia/abstract/utils.py
```diff
@@ -745,122 +753,128 @@ def amerge(x1: dtype.TypeMeta, x2, loop, forced, bp):
...
 @overload  # noqa: F811
-def amerge(x1: AbstractJTagged, x2, loop, forced, bp):
+def amerge(self, x1: AbstractJTagged, x2, forced, bp):
     args1 = x1.element
     args2 = x2.element
-    merged = amerge(args1, args2, loop, forced, bp)
-    if forced or merged is args1:
+    merged = self(args1, args2, forced, bp)
+    if forced or merged is args1:  # pragma: no cover
+        # This used to be covered before the `x1 is x2` check in amerge
```
abergeron: I'm sure you can cover this with `amerge(AbstractJTagged(1), AbstractJTagged(ANYTHING), forced=True)`

myia/abstract/utils.py
breuleux: Because of the line that goes `x1 = x1.result()` if `x1` is `Pending`. Without this adjustment, it would call the function for `Pending`. I suppose I could replace that line with a return of a recursive call, though.

PR #237, opened 2019-07-24
Gradient and HyperMap support for Abstract[Tagged]Union

breuleux:
It is now possible to take the gradient of functions that take ADTs and union types, and use the primitives hastag, casttag and tagged.

This is based on #236


tests/test_grad.py
```diff
@@ -475,6 +488,38 @@ def test_transpose2(x, y, axis1, axis2):
     return array_to_scalar(sm)
 
 
+@grad_test(
+    (4.5,),
+    ((5.5, 1.3),),
+    pipeline=standard_debug_pipeline.configure(validate=False),
```
abergeron: You should fix the validate whitelist instead of disabling it.

PR #242, opened 2019-08-01
Use an ADT to implement lists

breuleux:
This changes the implementation and typing of lists so that they use the Cons and Empty ADTs instead:

```
List[T] = Union[Cons[T, List[T]], Empty]
```

This adds a few limitations on some edge cases of hastype(x, list) and lists of lists (we may be able to address both with an improvement in amerge), but besides that, everything works just as before.

List primitives are also removed, except for `make_list` which the inferrer replaces by the appropriate sequence of Cons cells.


myia/abstract/utils.py
```diff
@@ -183,7 +182,13 @@ def pytype_to_abstract(main: tuple, args):
 @overload  # noqa: F811
 def pytype_to_abstract(main: list, args):
     arg, = args
-    return AbstractList(type_to_abstract(arg))
+    argt = type_to_abstract(arg)
+    assert argt is ANYTHING
```
abergeron: Why even ask for args if you're only ever going to use ANYTHING?

myia/prim/grad_implementations.py
```diff
@@ -321,6 +314,18 @@ def bprop_tagged(x, t, out, dout):
     return (casttag(dout, t), zeros_like(t))
 
 
+@register_bprop(primops.raise_)
+def bprop_raise_(x, out, dout):
+    """Backpropagator for primitive `raise_`."""
+    raise x
```
abergeron: Can this really be reached?

myia/validate.py
```diff
@@ -102,9 +99,7 @@ def validate_abstract(self, a: (type(None), AbstractExternal), uses):
     P.dot,
     P.switch,
     P.return_,
-    # P.list_map,
-    P.list_append,
-    P.make_list,
+    # P.make_list,
```
abergeron: Stray comment.

tests/test_abstract.py
```diff
@@ -193,16 +195,16 @@ def test_repr():
...
+        'AbstractTaggedUnion(U(4 :: Int[16], 13 :: Float[32]))'
 
 
 def test_repr_recursive():
     sa = S(t=ty.Int[64])
     ta = T.empty()
-    la = L.empty()
-    la.__init__(ta)
+    la = T.empty()
+    la.__init__([ta])
```
abergeron: I know you can't play with AbstractList for recursive things, but the type you are building here can't match any real value.

myia/abstract/utils.py
breuleux: Sanity check, to get a failure when trying to do `hastype(x, List[Int[64]])` and whatnot. It should be a better error, I suppose, but I thought I'd leave that for later.

myia/prim/grad_implementations.py
breuleux: I'm honestly not sure. Probably not, but it's the only expression that has the right type.

myia/validate.py
breuleux: OK

tests/test_abstract.py
breuleux: These types do exist in the pipeline: recursive ADTs become recursive tuples after simplify_types.

abergeron: Yeah, I didn't think of that.

myia/validate.py
breuleux: Actually, it's not really a stray comment, unsupported operations are commented out in the whitelist to make it more obvious what is or isn't supported.

PR #245, opened 2019-08-07
Support default arguments, keyword arguments, varargs, kwargs

breuleux:
This adds support for:

* Default arguments: `def f(x=1, y=2)`
* Varargs: `def f(*args)`
* Keyword-only arguments: `def f(*, x, y)`
* Kwargs: `def f(**kwargs)`
* Calling with varargs: `f(*args, *more_args)`
* Calling with keyword arguments: `f(x=1, y=2)`
* Calling with kwargs: `f(**kwargs, **more_kwargs)`
* Any combination of these features that's valid in Python

There are mild limitations on these features:

* `*args` are resolved statically so they must always be a tuple
* `f(a, b)`, `f(x=a, y=b)` and `f(y=b, x=a)` are all considered different call signatures by the inferrer, and if the monomorphizer cannot statically resolve `f`, it will not be able to compile the code if more than one signature is found

Nonetheless, this PR should support 99% of use cases.

This PR is based on #242


myia/ir/anf.py
```diff
@@ -50,6 +51,12 @@ def __init__(self) -> None:
         self.debug = NamedDebugInfo(self)
         self.flags = {}
         self.transforms: Dict[str, Union[Graph, Primitive]] = {}
+        self.vararg = False
+        self.kwarg = False
+        self.defaults = []
+        self.kwonly = 0
+        self.user_graph = None
+        self.sig = None
```
abergeron: These new attributes are not documented.

breuleux:
I'm not super happy about hijacking `return_` either, but it's not necessarily terrible: it's a way to keep extra nodes reachable for cloning/managing purposes, and the inferrer will raise an error if we ever forget to get rid of them. Possibly a different "primitive" than `P.return_` could be used.

myia/ir/anf.py
breuleux: I documented most of them, `user_graph` and `sig` are sort of for internal use so I put an underscore in front.

PR #246, opened 2019-08-08
Frontend pytorch v0_0_2 

ethancaballero:
- nn.sequential fwd support (fwd only because aliasing issue)
- pt frontend support of grad from size(1) array output without .item (conversion to 0-d float) via MultiTypeGraph (and .item better matches pt other than detach that .item causes in pt)
- values method for dictionary and hypermap for dictionary
- test_pytorch_ops.py file for automating testing of fwd and/or bwd of pytorch ops; currently can test ops that receive a single tensor as input.
- `exp`, `log`, `relu`, `sigmoid`, `sum` (all_reduce but not reduce over specified dim yet), `tanh`, and `item` pass pytorch_ops tests
- `reshape` primitive support from all backends
- `reshape`, `t`, `view` are supported as pytorch_ops but not explicitly tested in test_pytorch_ops.py yet (because need to finish way to automatically test ops with 2 args); these are implicitly tested by other pytorch ops that use them
- `zeros_like` (for torch.Tensor) is implicitly tested when performing update step using nn.functional.linear when bias is None (in test_pytorch.py); (I think torch.Tensor.`zeros_like` is a hidden method that only `.backward` can use); to later support `zeros_like` (as `torch.` method), relay and nnvm files in backend folder will need be altered (it already can work with pt backend).


tests/frontends/test_pytorch.py
```diff
@@ -23,7 +23,7 @@ def test_pytorch_dtype_to_type():
 
 
 # Uncomment this line to print values at specific precision
-# torch.set_printoptions(precision=8)
+torch.set_printoptions(precision=8)
```
breuleux: Did you mean to uncomment it?

tests/frontends/test_pytorch_ops.py
```diff
@@ -0,0 +1,458 @@
...
+    def __init__(self):
+        # device used
+        self.dev = 'cpu'
+        # backend used
+        self.backend = 'pytorch'
+        # numerical precision
+        self.dtype = 'float32'
+
+
+args = Args()
```
breuleux: What does this do and where is it used?

tests/frontends/test_pytorch_ops.py
```diff
@@ -0,0 +1,458 @@
...
+    (nn.Parameter(torch.Tensor([-2.2]))),
+    (nn.Parameter(torch.Tensor(MA(2, 3)))),
+)
+
+
+@pytest.mark.parametrize(
+    'name,args',
+    [(op, single_tensor_args) for op in all_torch_ops__1_tensor_arg]
+)
+@pytest.mark.timeout(10)
```
breuleux: Why is there a timeout?

tests/frontends/test_pytorch_ops.py
```diff
@@ -0,0 +1,458 @@
...
+        parse=steps.step_parse,
+        resolve=steps.step_resolve,
+        infer=steps.step_infer,
+        specialize=steps.step_specialize,
+        opt=steps.step_debug_opt,
+        validate=step_grad_validate,
+        # compile=steps.step_compile,
+        export=steps.step_debug_export,
+    )
+)
```
breuleux: You should import `grad_whitelist`, `grad_validate_abstract`, `grad_wrap` and `grad_pipeline` from `test_grad`, to avoid unnecessary code duplication.

tests/frontends/test_pytorch_ops.py
```diff
@@ -315,6 +270,9 @@ def _name_args_helper(name, args):
 # all_torch_ops = dir(torch)
 # all_torch_tensor_ops = dir(torch.Tensor([5.49670]))
 
+# pytest_timeout used so that unmapped ops with infinite call recursion
+# don't take very long time to fail
+pytest_timeout_time = 10
```
breuleux: #238 adds a 30s timeout for every test, so I would remove specific timeouts like the ones below.

PR #251, opened 2019-08-14
Misc improvements (novodecies)

breuleux:
* Some error messages are improved
* `debug.logword` is replaced by the `breakword` package (install from pip if needed)
* Minor efficiency improvements

Each commit is self-contained. Do not squash.


myia/abstract/infer.py
```diff
@@ -315,12 +317,15 @@ def get_inferrer_for(self, mg: MetaGraphFunction):
                 f'{fn.values[VALUE]} {fn.values[DATA] or ""}.'
             )
 
-        elif not isinstance(fn, AbstractFunction):
+        elif isinstance(fn, AbstractClassBase):
             g = ref.node.graph
```
abergeron: Is that to be able to call classes?

breuleux: Calling dataclass instances if they define `__call__`, yes.

PR #252, opened 2019-08-14
Macro system

breuleux:
Add a macro system. A macro simply replaces a subtree with another. The following primitives are removed and replaced by macros:

* `apply`
* `dict_values`
* `getattr` (except on `AbstractClassBase`)
* `make_list`
* `resolve`
* `user_switch`

The `getattr` and `setattr` primitives are renamed to `record_getitem` and `record_setitem` and now only work on records/ADTs.

A hack in `vm` explicitly maps some macros such as `resolve` to the appropriate vm implementations to keep tests working. Ultimately, we will remove support for running the DebugVM without inference and monomorphization, at which point the hack can be removed.

Furthermore, the `grad` MetaGraph is replaced by a macro that returns a MetaGraph. This brings a few improvements along the way:

* The derivative wrt the output can now be provided easily, using the `dout` keyword argument: `grad(f)(x, y, dout=2)`
* `value_and_grad(f)` is now equivalent to `grad(f, return_value=True)`
* `grad` can now work on a Primitive or MetaGraph.

This PR is based on #251


tests/test_lang.py
```diff
@@ -130,11 +130,6 @@ def test_swap(x, y):
 ###################
 
 
-@parse_compare(13)
-def test_list(x):
-    return [x, x + 1, x + 2]
```
abergeron: Probably skip this until we add inference to the debug pipeline.

Issue #239, opened 2019-07-29
Dealing with aliasing


breuleux:
Myia is currently not able to handle aliased tensors in data structures. This issue can crop up in the Pytorch frontend, in code like this:

```python
class LinearSeq(torch.nn.Module):
    def __init__(self, a, b):
        super(LinearSeq, self).__init__()
        self.lin = torch.nn.Linear(a, b)
        self.seq = torch.nn.Sequential(self.lin)

    def forward(self, x):
        return self.seq(x)
```

The problem is that Myia sees both `self.lin` and `self.seq[0]`, but it understands them as different parameters rather than the same parameter. Thus, if `forward` only uses `self.seq`, the gradient wrt `self.lin` is zero, and the update will be applied on `seq`, but not `lin`. Furthermore, if both `seq` and `lin` are used, they will accumulate gradients separately and will diverge.

This is a difficult problem, and if we handle it, I believe it would be best to consider the aliasing patterns statically (by which I mean specialize graphs wrt aliasing patterns). The fact that two tensors in opposite corners of a data structure may be aliased seems particularly difficult to deal with, but maybe we can get away with only supporting a few simple patterns.

So the question is, how do we deal with this?

PR #255, opened 2019-08-21
Merge gradients for aliased parameters

breuleux:
This implements a framework to deal with #239.

The `ALIASID` track can be optionally computed and represents an equivalence class such that if x and y have the same `ALIASID` they must refer to the same object. `grad` takes these into account in order to make sure that df/dx and df/dy are also equal by adding up these contributions.

It is possible to customize the function that determines which data is "aliasable". One is defined, `myia.abstract.aliasing.ndarray_aliasable`, which marks numpy arrays. Data inside lists or ADTs is considered non-aliasable because each element of a list or ADT must be associated to the same AbstractValue and therefore the same `ALIASID`, so there is no way to track aliasing on individual elements.

The PR also adds support for `record_setitem` and `dict_setitem` and makes necessary adjustments to `amerge` and a few other functions.

### DIFF
--- a/myia/opt/opt.py
+++ b/myia/opt/opt.py
@@ -87,7 +87,7 @@ class PatternSubstitutionOptimization:
         else:
             return None
 
-    def __str__(self):
+    def __str__(self):  # pragma: no cover
         return f'<PatternSubstitutionOptimization {self.name}>'
 
     __repr__ = __str__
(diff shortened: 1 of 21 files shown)


myia/opt/opt.py
```diff
@@ -87,7 +87,7 @@ def __call__(self, optimizer, node):
         else:
             return None
 
-    def __str__(self):
+    def __str__(self):  # pragma: no cover
```
abergeron: This was covered in the last PR.

abergeron:
```suggestion
    def __str__(self):
```

PR #260, opened 2019-09-03
Improvements to method resolution on abstract types

breuleux:
Based on #259 

* Remove `methods` field from `AbstractClassBase`. It is unnecessary.
* Method resolution now *always* relies on `a.values[TYPE]`
* Remove the `type_token` function
* Replace `AbstractPyTorchTensor` by `AbstractArray(..., {TYPE: PyTorchTensor})`
* Remove overloads in the pytorch frontend that are now unnecessary.

### DIFF
--- a/myia/abstract/data.py
+++ b/myia/abstract/data.py
@@ -242,7 +242,7 @@ class AbstractValue(Interned, PossiblyRecursive):
         self.values = TrackDict(values)
 
     def dtype(self):
-        """Return the type of this scalar."""
+        """Return the type of this AbstractValue."""
         t = self.values[TYPE]
         if isinstance(t, Pending) and t.done():
             t = t.result()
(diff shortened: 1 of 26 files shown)


myia/abstract/data.py
```diff
@@ -241,6 +241,13 @@ def __init__(self, values):
         super().__init__()
         self.values = TrackDict(values)
 
+    def dtype(self):
+        """Return the type of this scalar."""
```
abergeron: The documentation is wrong.

myia/frontends/pytorch.py
```diff
@@ -172,61 +161,38 @@ def new_module(*args):
...
-
 
-@convert_arg.register
-def _convert_arg(self, arg, orig_t: AbstractPyTorchTensor):
-    et = orig_t.element
-    assert isinstance(et, AbstractScalar)
-    et = et.values[TYPE]
-    assert issubclass(et, Number)
+@convert_arg_array.register
+def _convert_arg_array(arg, t: PyTorchTensor, et, orig_t):
```
abergeron: Since you changed the location of the type checks in pipeline/steps.py, they should be present here too.

Alternatively this could only check that it is a pytorch tensor and convert it to numpy, leaving the type/shape checks as common code.

breuleux: Yes, good idea.

PR #262, opened 2019-09-03
string compare and return

ethancaballero:
- support for comparing strings and returning strings

### DIFF
--- a/myia/opt/__init__.py
+++ b/myia/opt/__init__.py
@@ -1,6 +1,6 @@
 """Optimization submodule."""
 
-from .clean import _strmap_tag, simplify_types, str_to_tag, type_to_tag  # noqa
+from .clean import simplify_types, str_to_tag, type_to_tag  # noqa
 from .cse import CSE, cse  # noqa
 from .dde import DeadDataElimination  # noqa
 from .opt import (  # noqa
--- a/myia/opt/__init__.py
+++ b/myia/opt/__init__.py
@@ -1,6 +1,6 @@
 """Optimization submodule."""
 
-from .clean import simplify_types, type_to_tag  # noqa
+from .clean import _strmap_tag, simplify_types, str_to_tag, type_to_tag  # noqa
 from .cse import CSE, cse  # noqa
 from .dde import DeadDataElimination  # noqa
 from .opt import (  # noqa
--- a/myia/opt/clean.py
+++ b/myia/opt/clean.py
@@ -4,6 +4,7 @@ import weakref
 from itertools import count
 
 from ..abstract import (
+    ANYTHING,
     TYPE,
     VALUE,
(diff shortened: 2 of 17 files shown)


myia/opt/__init__.py
```diff
@@ -1,6 +1,6 @@
 """Optimization submodule."""
 
-from .clean import simplify_types, type_to_tag  # noqa
+from .clean import _strmap_tag, simplify_types, str_to_tag, type_to_tag  # noqa
```
abergeron: Don't expose private stuff.
```suggestion
from .clean import simplify_types, str_to_tag, type_to_tag  # noqa
```

myia/opt/clean.py
```diff
@@ -35,11 +35,34 @@ def type_to_tag(t):
...
 @abstract_clone.variant
 def _reabs(self, a: AbstractClassBase):
     return (yield AbstractTuple)(self(x) for x in a.attributes.values())
 
 
+@overload  # noqa: F811
+def _reabs(self, a: AbstractScalar):
+    if a.values[TYPE] == String:
+        v = a.values[VALUE]
+        if isinstance(v, str):
```
abergeron: This should always be true.  Make it an assert or remove it.

myia/pipeline/steps.py
```diff
@@ -668,7 +678,11 @@ def convert_result(self, res, orig_t, vm_t: AbstractTuple, backend,
 @overload  # noqa: F811
 def convert_result(self, arg, orig_t, vm_t: AbstractScalar, backend,
                    return_backend):
-    return backend.to_scalar(arg)
+    ret = backend.to_scalar(arg)
+    if orig_t.values[TYPE] == dtype.String and \
+            vm_t.values[TYPE] == dtype.Int[64]:
```
abergeron: You don't need to check that vm_t is Int[64].  This is implicit.

myia/pipeline/steps.py
```diff
@@ -614,11 +618,17 @@ def convert_arg(self, arg, orig_t: AbstractScalar, backend):
...
+        if not isinstance(arg, str):
+            raise MyiaInputTypeError(f'Expected string')
+        arg = str_to_tag(arg)
+        t = dtype.Int[64]
     else:
         raise MyiaInputTypeError(f'Invalid type: {t}')
-    expected_value = orig_t.values[VALUE]
-    if expected_value is not ANYTHING and expected_value != arg:
-        raise MyiaInputTypeError(f'Invalid value: {arg}')
+    if not issubclass(orig_t.values[TYPE], dtype.String):
```
breuleux: Why is that check here? If the program is compiled assuming that we are giving the string "hello" as a parameter, then we should check that this is the right string.

PR #267, opened 2019-09-06
Reorganize the pipeline

breuleux:
The pipeline interface is simplified in the following ways:

* `PipelineStep` is gone. Any plain function can be used as a step. `PipelineResource` is also gone. Just define a `Partializable` class.
* Resources are not specially handled by the `Pipeline` class anymore. `PipelineDefinition` now only takes the steps as keyword arguments.
* Resources are now created by the `resources` step and are threaded as an argument to each step alongside graph, argspec, etc.
* To configure resource `x` you need to configure `resources.x` instead of just `x`.
* Some parameters that were previously on steps are moved to resources (`backend`, `return_backend`, `operation_whitelist`, `validate_abstract`)
* Optimizers are given resources directly instead of an Optimizer object


abergeron:
This doesn't have 100% coverage, but it appears that codecov is fine with it.

Hmmm...

PR #269, opened 2019-09-09
Reorganization

breuleux:
Based on #268 

* Replace the embed, typeof, array_len and tuple_len primitives by macros.
* Remove py_implementations for all operations that are removed by simplify_types.
* Remove the hack in DebugVM to support a few macros and metagraphs and adjust tests accordingly.
* Make method_map use inheritance patterns, so that xtype.Int can use the methods defined on xtype.Number and xtype.Int.


myia/vm.py
```diff
@@ -231,8 +229,9 @@ def _call(self, graph: Graph, args: List[Any]):
 
         assert isinstance(graph, Graph)
 
-        if graph not in self._vars:
-            self._acquire_graph(graph)
+        assert graph in self._vars
+        # if graph not in self._vars:
```
abergeron:
```suggestion
```

myia/vm.py
```diff
@@ -231,8 +229,9 @@ def _call(self, graph: Graph, args: List[Any]):
...
+        #     self._acquire_graph(graph)
```
abergeron:
```suggestion
```

PR #270, opened 2019-09-10
Reorganize operations

breuleux:
This is the beginning of an attempt to reorganize all operations in Myia. This sets up a new system that cohabits with the old system, for now.

* All operations, whether they be macros, composites or primitives, would be moved to `myia.operations`
* Each primitive would have its own file, e.g. `scalar_pow` would live in `myia.operations.prim_scalar_pow` (`scalar_pow` is the only primitive to move to the new system at the moment).
* `myia.operations.xyz` is an instance of `Operation`, which optionally defines a Python implementation and a default mapping (which can be a Primitive, a Macro, etc.) The default mapping can be overridden in the object_map.
* A `Primitive` now defines an optional Python implementation (for the debug VM), a default inferrer, and a default bprop, which can all be overridden.

Under the new system, the following files would disappear:

* `myia.prim.ops`: now `myia.operations.primitives`
* `myia.prim.py_implementations`: import from `myia.operations` instead.
* `myia.prim.grad_implementations`
* `myia.abstract.prim`
* `myia.composite`, now in individual modules under `myia.operations`
* `myia.macros`, now in individual modules under `myia.operations`


myia/compile/backends/nnvm.py
```diff
@@ -411,3 +411,22 @@ def from_scalar(self, s, t):
 def NNVMBackendR(target, device_id):
     """NNVM Proxy."""
     return HandleBackend(NNVMBackend(target, device_id))
+
+
+__all__ = [
+    'NNVMBackend',
+    'NNVMBackendR',
```
abergeron: This should only export NNVMBackend and NNVMBackendR

myia/compile/backends/pytorch.py
```diff
@@ -288,3 +288,21 @@ def from_scalar(self, s, t):
 def PyTorchBackendR(device):
     """Pytorch proxy."""
     return HandleBackend(PyTorchBackend(device))
+
+
+__all__ = [
+    'PyTorchBackend',
+    'PyTorchBackendR',
```
abergeron: Same here

myia/compile/backends/relay.py
```diff
@@ -400,3 +400,21 @@ def from_scalar(self, s, t):
 def RelayBackendR(target, device_id):
     """Relay proxy."""
     return HandleBackend(RelayBackend(target, device_id))
+
+
+__all__ = [
+    'CompileGraph',
+    'MAP',
+    'RelayBackend',
+    'RelayBackendR',
```
abergeron: And here.

myia/compile/vm.py
```diff
@@ -361,3 +361,9 @@ def inst_external(self, fn, args):
         outs = fn(*(self._ref(a) for a in args))
         for o in outs:
             self._push(o)
+
+
+__all__ = [
+    'FinalVM',
+    'struct_partial',
```
abergeron: No struct_partial export
```suggestion
```

myia/frontends/pytorch.py
```diff
@@ -183,3 +183,9 @@ def _to_abstract(self, v: torch.nn.Parameter, **kwargs):
         }),
         {SHAPE: tuple(v.shape), TYPE: PyTorchTensor},
     )
+
+
+__all__ = [
+    'mod_sub',
+    'pytorch_dtype_to_type',
+]
```
abergeron: I would argue that this shouldn't have an `__all__`

myia/grad.py
```diff
@@ -623,3 +623,18 @@ def bprop_to_grad_transform(prim, **flags):
...
+    'BPropAppRemapper',
+    'BPropRemapper',
+    'FPropAppRemapper',
+    'FPropRemapper',
+    'GradRemapper',
+    'J',
+    'SensRemapper',
+    'SlaveRemapper',
+    'bprop_to_grad_transform',
+    'wrap_grad_transform',
```
abergeron: This might be a bit too much exporting.

myia/graph_utils.py
```diff
@@ -102,3 +102,12 @@ def toposort(root: T,
         done.add(node)
         yield node
         todo.pop()
+
+
+__consolidate__ = True
+__all__ = [
+    'T',
```
abergeron: Not this.
```suggestion
```

myia/info.py
```diff
@@ -206,3 +206,14 @@ def __exit__(self, type, value, tb):
...
+
+
+__consolidate__ = True
+__all__ = [
+    'About',
+    'DebugInfo',
+    'DebugInherit',
+    'NamedDebugInfo',
+    'StackVar',
+    'current_info',
```
abergeron: Maybe current_info and StackVar are somewhat private.

myia/ir/manager.py
```diff
@@ -897,3 +897,35 @@ def __enter__(self):
...
+    'NestingStatisticGraphWise',
+    'NestingStatisticWholesale',
+    'NodesStatistic',
+    'ParentProxy',
+    'ParentStatistic',
+    'PerGraphStatistic',
+    'RecursiveStatistic',
+    'ScopeStatistic',
+    'UsesStatistic',
+    'manage',
```
abergeron: I think we only want manage and ManagerError here.

myia/opt/cse.py
```diff
@@ -99,3 +99,10 @@ def __call__(self, root):
         """Apply CSE on root."""
         chg = cse(root, self.resources.manager)
         return chg and self.report_changes
+
+
+__all__ = [
+    'CSE',
+    'cse',
+    'group_nodes',
```
abergeron: Perhaps not this.
```suggestion
```

myia/opt/lib.py
```diff
@@ -1048,3 +1048,17 @@ def __call__(self, root):
             node.abstract = newtype
 
         return len(nodes) > 0
+
+
+__all__ = [
+    'JElim',
+    'M',
```
abergeron: Maybe not this
```suggestion
```

myia/utils/env.py
```diff
@@ -106,3 +106,13 @@ def __len__(self):
...
 newenv = EnvInstance()
+
+
+__consolidate__ = True
+__all__ = [
+    'EnvInstance',
+    'SymbolicKeyInstance',
+    'newenv',
+    'require_same',
+    'smap',
```
abergeron: Not this?

myia/utils/intern.py
```diff
@@ -307,3 +307,26 @@ def intern(inst):
...
+    'PossiblyRecursive',
+    'RecursionException',
+    'Wrapper',
+    'deep_eqkey',
+    'eq',
+    'eqkey',
+    'eqrec',
+    'hash',
+    'hashrec',
+    'intern',
```
abergeron: I'm pretty sure some of those should not be exported.

myia/utils/misc.py
```diff
@@ -435,3 +435,31 @@ def assert_scalar(*args):
...
+    'TaggedValue',
+    'UNKNOWN',
+    'assert_scalar',
+    'core',
+    'dataclass_fields',
+    'is_dataclass_type',
+    'keyword_decorator',
+    'list_str',
+    'repr_',
+    'resolve_from_path',
```
abergeron: A bunch of those should not be here.

myia/utils/trace.py
```diff
@@ -335,3 +335,18 @@ def new_fn(**kwargs):
...
+__consolidate__ = True
+__all__ = [
+    'DoTrace',
+    'ProfileResults',
+    'Profiler',
+    'TraceExplorer',
+    'TraceListener',
+    'Tracer',
+    'TracerContextManager',
+    'glob_to_regex',
```
abergeron: Maybe not this.  And some of the above ones too.

myia/utils/unify.py
```diff
@@ -613,3 +613,25 @@ def reify(self, v, equiv: EquivT) -> Any:
...
+
+
+__consolidate__ = True
+__all__ = [
+    'FilterVar',
+    'PredicateSet',
+    'RestrictedVar',
+    'SVar',
+    'Seq',
+    'T',
```
abergeron: Not this.  And perhaps some others too.

myia/utils/variables.py
```diff
@@ -50,3 +50,9 @@ def _is_c(n):
 Xs = SVar(Var())
 Ys = SVar(Var())
 Cs = SVar(constvar())
+
+
+__consolidate__ = True
+__all__ = [
+    'constvar',
```
abergeron: We might not want this, but all the other ones instead.

myia/vm.py
```diff
@@ -290,3 +290,11 @@ def _handle_node(self, node: ANFNode, frame: VMFrame):
 
         else:
             raise AssertionError("Unknown node type")  # pragma: no cover
+
+
+__all__ = [
+    'Closure',
+    'Partial',
+    'VM',
+    'VMFrame',
```
abergeron: Probably not VMFrame.

myia/xtype.py
```diff
@@ -318,3 +318,41 @@ class External(typing.Generic[T]):
...
+    'Float',
+    'Int',
+    'NDArray',
+    'Nil',
+    'NotImplementedType',
+    'Number',
+    'Object',
+    'String',
+    'SymbolicKeyType',
+    'T',
```
abergeron: Not T, TypeMeta, as_frozen and maybe not Type too.

myia/frontends/pytorch.py
breuleux: I'd keep `pytorch_dtype_to_type` I think.

myia/info.py
breuleux: Could be useful to export `current_info`

myia/utils/misc.py
breuleux: T1 and T2 don't belong, but the rest seems fine.

myia/utils/variables.py
breuleux: opt/lib does use constvar.

myia/xtype.py
breuleux: `TypeMeta` is used elsewhere.

PR #272, opened 2019-09-10
Split the backend in separate processes

abergeron:
This creates a new process for each (backend, device) pair and talks to it over stdin/stdout.

This is required to stop huge c++ librairies (pytorch, tvm, ...) from fighting over private c++ standard header symbols and causing weird crashes and hangs depending on the load order.


abergeron:
It seems travis didn't upload the coverage properly, which is why we get this problem.

PR #277, opened 2019-09-17
docstring fix in find_aliases and other small fixes.

fosterrath-mila:
In the context of regenerating the Myia documentation, I started looking at the definitions. Here are a few little tweaks to get rid of doc compilation issues.

### DIFF
--- a/myia/abstract/aliasing.py
+++ b/myia/abstract/aliasing.py
@@ -41,7 +41,6 @@ def _explore(self, v: object, vseq, path):
         for k, x in dataclass_fields(v).items():
             yield from self(x, vseq, (*path, k))
 
-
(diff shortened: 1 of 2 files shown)


abergeron:
I don't think this fix is good for the format of the docstring.  We use: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

myia/abstract/aliasing.py
```diff
@@ -41,7 +41,6 @@ def _explore(self, v: object, vseq, path):
         for k, x in dataclass_fields(v).items():
             yield from self(x, vseq, (*path, k))
 
-
```
abergeron: We need two blank lines between methods.

PR #285, opened 2019-09-26
Misc improvements (tervicies)

breuleux:
The commits are clean, so there is no need to squash them.


abergeron:
Coverage is not 100%, codecov is lying.

PR #286, opened 2019-09-26
Add some missing primitives to Relay

abergeron:


abergeron:
What is missing now is max_pool2d and array_setitem

PR #287, opened 2019-09-30
Add support for ruamel.yaml

abergeron:
Since ruamel.yaml merged the size_t fix after a couple of days (https://bitbucket.org/ruamel/yaml.clib/pull-requests/1/use-size_t-for-marks-in-cython-to-handle/diff) and PyYAML has apparently been "considering" a very similar fix for at least a month now: https://github.com/yaml/pyyaml/pull/310, I believe it is in our best interest to possibly switch to ruamel.yaml in order to have a "fixed" version as soon as possible.

Also @ethancaballero and I did some runs using the python version of PyYAML and it is so slow that I consider it unusable for our purposes and therefore removed support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ env:
   - TEST_SUITE=static
   - TEST_SUITE=unit
 script:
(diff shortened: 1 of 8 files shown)


abergeron:
The CI install modifications are done and working now.

PR #289, opened 2019-09-30
Proper implementation of Python's dunder method protocol

breuleux:
This PR is based on #288

When Python encounters a binary operator like `+` or `@`, it follows a protocol. Supposing we have the call `x + y`, Python does the following:

* Try `x.__add__(y)`
* If that method does not exist or returns the special value `NotImplemented`, try `y.__radd__(x)`.
* If that method does not exist or returns the special value `NotImplemented`, raise an exception.

The methods to try depend on the operator, for example `x < y` will try `x.__lt__(y)` then `y.__gt__(x)`. Some operators, like `x == y`, have a fall back to `x is y`.

This protocol now works properly in Myia. One immediate consequence is that the expression `model = model - dmodel * lr` in some of the model examples can now be written as `model = model - lr * dmodel`.


myia/operations/ops_dunder.py
```diff
@@ -41,43 +40,126 @@ def protocol(data, x):
...
+mod = dunder_protocol_binary(
+    name='mod', lattr='__mod__', rattr='__rmod__',
+    pyop=operator.mod
+)
+pow = dunder_protocol_binary(
+    name='pow', lattr='__pow__', rattr='__rpow__',
+    pyop=operator.pow
+)
+truediv = dunder_protocol_binary(
+    name='truediv', lattr='__truediv__', rattr='__truediv__',
```
abergeron:
```suggestion
    name='truediv', lattr='__truediv__', rattr='__rtruediv__',
```

myia/operations/ops_dunder.py
```diff
@@ -41,43 +40,126 @@ def protocol(data, x):
...
+    pyop=operator.truediv
+)
+floordiv = dunder_protocol_binary(
+    name='floordiv', lattr='__floordiv__', rattr='__floordiv__',
```
abergeron:
```suggestion
    name='floordiv', lattr='__floordiv__', rattr='__rfloordiv__',
```

PR #291, opened 2019-10-02
frontend_pytorch_v0_0_4: runs mnist example from pytorch examples with small changes

ethancaballero:
- runs mnist example from pytorch examples with small changes
- also adds support for cast option of pytorch function that needed support for `.dtype` attribute

### DIFF
--- a/myia/operations/op_gadd.py
+++ b/myia/operations/op_gadd.py
@@ -1,7 +1,7 @@
 """Implementation of the 'gadd' operation."""
 
 from ..lib import HyperMap, MultitypeGraph, core
-from ..xtype import EnvType, Nil, Number
+from ..xtype import Bool, EnvType, Nil, Number
 from .primitives import env_add, scalar_add
 
 _leaf_add = MultitypeGraph('gadd')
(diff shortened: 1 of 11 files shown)


myia/operations/op_gadd.py
```diff
@@ -25,6 +25,13 @@ def _nil_add(x, y):
     return None
 
 
+@_leaf_add.register(Bool, Bool)
+@core
+def _bool_add(x, y):
+    assert x is y
+    return y
```
breuleux: Shouldn't we use `bool_or(x, y)`? That's essentially boolean addition.

tests/multitest.py
```diff
@@ -26,6 +26,14 @@ def eqtest(a1: (np.ndarray, int, float), a2,
     return np.allclose(a1, a2, rtol=rtol, atol=atol)
 
 
+@overload  # noqa: F811
+def eqtest(x: type(None), y, **kwargs):
+    if y is None:
+        return True
+    else:
+        return y.sum().item() == 0
```
breuleux: I think we want to check that `y` is all zeros, not that its sum is zero. The code for that would be `(y == 0).all()`.

PR #292, opened 2019-10-02
Add more relay ops

abergeron:
- array_max
- conv2d (and grads, except for dilation and groups)
- max_pool2d
- max_pool2d_grad


abergeron:
I apparently have a bug in the input gradient for conv2d, I'll try to fix it.

abergeron:
The bug is in the implementation of conv2d_transpose on relay that can't do proper gradient for inputs that were strided in a certain way.

abergeron:
I've also added abs, sign, concat and split.

tests/frontends/test_pytorch_ops.py
```diff
@@ -130,12 +147,19 @@ def mksens(x):
...
 
 
 @myia_function_test(marks=[pytest.mark.grad], id='grad')
-def fwd_and_bwd(self, fn, args, broad_specs=None, pipeline=standard_pipeline):
-    _fwd_and_bwd(fn, args, broad_specs, pipeline)
+def _fwd_and_bwd(self, fn, args, broad_specs=None, pipeline=standard_pipeline,
+                 backend=False, atol=1e-8, rtol=1e-5,
+                 grad_atol=1e-6, grad_rtol=1e-5):
+    __fwd_and_bwd(fn, args, broad_specs, pipeline, backend, atol=atol,
+                  rtol=rtol, grad_atol=grad_atol, grad_rtol=grad_rtol)
```
breuleux: Inline `__fwd_and_bwd` in here, since it's the only use.

abergeron: done

PR #294, opened 2019-10-04
Add Universe, Handle and auto-monadization

breuleux:
Based on #293

This adds an auto-monadization system of sorts so that we can handle stateful operations. The `Universe` and `Handle` types are added, which essentially functions as follows (the `use_universe` flag is required for now):

```python
@myia(use_universe=True)
def iadd(h, x):
    current = handle_get(h)
    new = add(current, x)
    handle_set(h, new)
    return None

h = HandleInstance(10)
iadd(h, 42)
assert h.state == 52
```

The above function is internally transformed into the following one:

```python
def iadd(U, h, x):
    U, current = handle_get(U, h)
    U, new = add(U, current, x)
    U, _ = handle_set(U, h, new)
    return U, None
```

So every function gets a new parameter, the "universe",  which is called `U`, which is a key value store where keys are handles. They also always return a pair, where the first element is the updated universe. At the end of the call, Myia updates all the handles according to their values in the universe returned by the top level function.

A handle can be used as a sort of cell to store state, like in the example. For printing, there would be a global `print_handle`, and for random numbers, the handle would point to the random state.

The system as implemented in this PR has the following limitations, which will have to be addressed in future PRs:

* There is no backend support. The debug VM must be used.
* There is no gradient support. This shouldn't be too difficult, and in theory it shouldn't matter for features like print or random numbers (since we're not taking gradients wrt these states).
* It doesn't work with partials, and it's not clear how to fix that: a partial sets the first argument(s) of a function, but the `U` argument should be threaded in every actual call, it should never be stored in a partial. One solution is to special-case this argument with partial (except we want to be able to eliminate it if it is unused, so there would be two versions of partial?), another is to always have `U` as the *last* argument (but this doesn't mesh well with varargs).
* There are no optimizations on universes yet. I have written one, but I think it'll be more appropriate to add it later.


abergeron:
I'm reasonably certain that I can make this work in the relay backend.

myia/pipeline/steps.py
```diff
@@ -450,7 +467,11 @@ def wrapped(*args):
             raise MyiaInputTypeError('Wrong number of arguments.')
         args = tuple(_to_backend(to_canonical(arg, ot), backend, vt)
                      for arg, ot, vt in zip(args, orig_arg_t, vm_arg_t))
-        res = fn(*args)
+        if resources.universal:
+            unv, res = fn(new_universe, *args)
+            unv.commit()
```
abergeron: Why do you need a separate commit step here?

breuleux: To update all the states stored in the handles. So if e.g. you have a RNG, you will have a handle with its initial state, then Myia will keep track of the state using the universe, but nothing actually gets changed. `unv.commit()` will perform the update, so that when you call fn again, it will start where the first call left off.

PR #297, opened 2019-10-22
Frontend pytorch v0 0 6; grads of sequantial alias

ethancaballero:
- grads of sequantial alias are correct, but has some duplicates in named_parameters


myia/api.py
```diff
@@ -131,15 +131,16 @@ def myfun2(cond, x, y):
 # Move value to target accelerator hardware #
 #############################################
 
-def to_device(value, backend, backend_options=None, *, orig_t=None, vm_t=None):
+def to_device(value, backend, backend_options=None, broaden=True, *,
```
abergeron: Please make broaden a keyword-only argument.

tests/frontends/test_pytorch.py
```diff
@@ -455,6 +456,95 @@ def step(model, inp):
...
+
+    inp = torch.Tensor(MA(2, 4, dtype=args.dtype))
+    model = Linear_Seq(4, 2, 3)
+    target = torch.Tensor([2.5])
+
+    """
+    from myia.abstract.aliasing import find_aliases
+    al = find_aliases(model, aliasable=tensor_pytorch_aliasable)
+    print("alias", al)
+    # """
```
abergeron: Please don't leave code like this in the tests

tests/frontends/test_pytorch.py
```diff
@@ -455,6 +456,95 @@ def step(model, inp):
     assert torch.allclose(output, output_expected)
 
 
+def test_module_linear_seq_bwd():
```
abergeron: Is there a way to make this test much shorter using some of the mt stuff?

myia/abstract/aliasing.py
```diff
@@ -41,6 +38,16 @@ def _explore(self, v: object, vseq, path):
...
+    else:
+        vseq = (*vseq, v)
+
+        try:
+            fields = get_fields(v)
+        except TypeError:
+            return None
+
+        for k, x in fields.items():
+            yield from self(x, vseq, (*path, k))
```
breuleux: Since `get_fields` also works for dataclasses, you can replace the entire body of this overload by the contents of the else branch (you don't need to check for is_dataclass).

tests/test_edge_cases.py
```diff
@@ -37,3 +39,22 @@ def g(x, y):
...
+        """Apply the layer."""
+        return self.W
+
+
+def test_get_dclass_fields():
+    from myia.utils import get_fields
+    lin = Linear(np.ones((2, 1)))
+    f = get_fields(lin)
+
+    assert eqtest(list(f)[0][1], np.ones((2, 1)))
```
breuleux: If you do the change I suggested in the other comment, I think this test is unnecessary.

myia/utils/get_fields.py
```diff
@@ -0,0 +1,22 @@
...
+
+from .misc import dataclass_fields
+from .overload import overload
+
+
+@overload
+def get_fields(instance: object):
+    """Returns fields of an instance."""
+    if is_dataclass(instance):
+        return dataclass_fields(instance).items()
```
breuleux:
```suggestion
        return dataclass_fields(instance)
```

The overload for Module returns a dictionary, and _explore expects a dictionary.

PR #300, opened 2019-10-24
Try to fix jenkins coverage report

abergeron:


abergeron:
I'm 98% sure that this will fix the problem of jenkins sending its coverage report into the void sometimes.

PR #301, opened 2019-11-05
Implementation of random number generator for Myia

notoraptor:
Hi @abergeron @breuleux 

I just make this PR to ease any discussion around the RNG implementation.

Currently, I have implemented the RNG in pure Python, based on Theano code. For reference:
- Theano code is here: https://github.com/Theano/Theano/blob/master/theano/sandbox/rng_mrg.py
- Original Java implementation by Pierre Lecuyer is here: https://github.com/umontreal-simul/ssj/blob/master/src/main/java/umontreal/ssj/rng/MRG31k3p.java

Implementation currently seems to work. I also added a test to compare output with the Theano output, and it currently produces same results.

Then, my next step was to try to compile the python implementation into myia, but I am facing many issues:
- Python implementation checks input seeds, as all values are not allowed, and generate some exceptions if bad seeds are given. It seems I can not currently compile exceptions raising with Myia, though I saw a `raise` primitive in operations. I don't yet know why compilation fails.
- I need to generate the output array and populate it with random values, but I don't yet have a way to create new data. @breuleux told me about `scalar_to_array` + `distribute` op, but, from what I saw, it seems these operations are currently used in `zero_like` op (isn't it?), which is not suited here. So I guess I need to create a `zeros` operation that can take just shape and dtype as argument, and I guess I should use `np.zeros` as entry point (ie. value to be replaced by the op later in compilation).
- I also need to create the initial random state from given seed, which can be either an integer or a ready-to-use integer vector with 6 values. I guess I should create something like `np.asarray` op here to ease this step.
- I simplified the random generator so that it uses only one vector as state to update, instead of using a matrix as in Theano (in Theano, this should be equivalent to get random numbers with parameter `nstreams=1`). Thus, in intermediate functions, I could just return the new updated state vector (it contains only 6 values) instead of trying to update a potential huge matrix inplace. So, I removed a place where inplace update was used, but I will still need to make inplace updates when generating random numbers, as I will need to a) create the output array, and b) update each array value with a generated random scalar.

This is where I am, currently. What do you think?

### DIFF
--- a/myia/compile/backends/pytorch.py
+++ b/myia/compile/backends/pytorch.py
@@ -56,8 +56,6 @@ def pytorch_random_initialize(seed):
 def pytorch_random_uint32(rstate, shape):
     """Implementation of random_uint32 for pytorch."""
     shape = tuple(dim.item() for dim in shape)
-    if not shape:
-        shape = (1,)
     rng = torch.Generator()
     rng.set_state(rstate)
     output = torch.zeros(shape, dtype=torch.int64)
@@ -551,8 +549,6 @@ class PyTorchBackend(Backend):
         elif isinstance(t, abstract.AbstractTaggedUnion):
             real_t = t.options.get(v.tag)
             return TaggedValue(v.tag, self.to_backend_value(v.value, real_t))
-        elif isinstance(t, abstract.AbstractRandomState):
-            return v.state
         else:
             raise NotImplementedError(f'to_backend_value for {t}')
 
--- a/myia/compile/backends/pytorch.py
+++ b/myia/compile/backends/pytorch.py
@@ -6,7 +6,7 @@ import torch
 from ... import abstract, xtype
 from ...ir import manage
 from ...operations import Primitive, primitives as P
-from ...utils import TaggedValue
+from ...utils import RandomStateWrapper, TaggedValue
 from ...xtype import Bool, Float, Int, UInt, type_to_np_dtype
 from ..cconv import closure_convert
 from ..transform import CompileGraphs, nonlinear_ops
@@ -44,6 +44,25 @@ def pytorch_take_grad_inp(nb_indices, indices, values):
                      .reshape((-1, row_size))
                      .sum(dim=(0,)))
     return output
+
+
+def pytorch_random_initialize(seed):
+    """Implementation of random_initialize for pytorch."""
+    rng = torch.Generator()
+    rng.manual_seed(seed.item())
+    return rng.get_state()
+
+
+def pytorch_random_uint32(rstate, shape):
+    """Implementation of random_uint32 for pytorch."""
+    shape = tuple(dim.item() for dim in shape)
+    if not shape:
+        shape = (1,)
+    rng = torch.Generator()
+    rng.set_state(rstate)
+    output = torch.zeros(shape, dtype=torch.int64)
+    output.random_(0, 2**32, generator=rng)
+    return rng.get_state(), output
 
 
 simple_mapping = {
@@ -93,6 +112,8 @@ simple_mapping = {
     P.take_grad_inp: pytorch_take_grad_inp,
 
     P.array_to_scalar: pytorch_array_to_scalar,
+    P.random_initialize: pytorch_random_initialize,
+    P.random_uint32: pytorch_random_uint32,
 }
 
 
@@ -484,13 +505,22 @@ class PyTorchBackend(Backend):
         if isinstance(t, abstract.AbstractScalar):
             return self.to_scalar(v)
         elif isinstance(t, abstract.AbstractArray):
-            return self.to_numpy(v)
+            # Convert torch tensor to numpy tensor.
+            output = self.to_numpy(v)
+            # If possible and necessary, cast numpy tensor to expected tensor.
+            array_type = t.element.xtype()
+            if array_type and array_type not in _type_map:
+                # Probably u16, u32 or u64. Let's cast.
+                output = output.astype(type_to_np_dtype(array_type))
+            return output
         elif isinstance(t, abstract.AbstractTuple):
             return tuple(self.from_backend_value(ve, te)
                          for ve, te in zip(v, t.elements))
         elif isinstance(t, abstract.AbstractTaggedUnion):
             return TaggedValue(v.tag, self.from_backend_value(
                 v.value, t.options.get(v.tag)))
+        elif isinstance(t, abstract.AbstractRandomState):
+            return RandomStateWrapper(self.to_numpy(v))
         else:
             raise NotImplementedError(f"Don't know what to do for {t}")
 
@@ -521,6 +551,8 @@ class PyTorchBackend(Backend):
         elif isinstance(t, abstract.AbstractTaggedUnion):
             real_t = t.options.get(v.tag)
             return TaggedValue(v.tag, self.to_backend_value(v.value, real_t))
+        elif isinstance(t, abstract.AbstractRandomState):
+            return v.state
         else:
             raise NotImplementedError(f'to_backend_value for {t}')
 
--- a/myia/compile/backends/relay.py
+++ b/myia/compile/backends/relay.py
@@ -442,10 +442,14 @@ def relay_random_uint32(c, ref_rstate, ref_shape):
     counter = relay.TupleGetItem(relay_state, 1)
     impl = relay_philox.Philox2x32(output_size)
     ctr = impl.generate_relay_counter_array(counter)
-    random = impl.philox_2x(ctr, key, to_float=False)
+    random = impl.philox_2x(ctr, key)
     # Reshape vector to expected shape.
     if shape:
+        # Reshape vector to output shape.
         random = relay.op.reshape(random, shape)
+    else:
+        # Convert 1-element vector to scalar
+        random = relay.op.take(random, relay.const(0), mode='fast')
     # Generate next state: same key, counter + 1
     next_rstate = relay.Tuple((
         key, relay.add(counter, relay.const(1, 'uint32'))))
--- a/myia/compile/backends/relay.py
+++ b/myia/compile/backends/relay.py
@@ -12,11 +12,11 @@ from ...abstract import AbstractTaggedUnion
 from ...graph_utils import toposort
 from ...ir import manage
 from ...operations import Primitive, primitives as P
-from ...utils import HandleInstance, TaggedValue
+from ...utils import HandleInstance, RandomStateWrapper, TaggedValue
 from ...xtype import UniverseType, type_to_np_dtype
 from ..channel import handle
 from ..transform import get_prim_graph, return_handles, wrap_result
-from . import Backend, Converter, HandleBackend
+from . import Backend, Converter, HandleBackend, relay_philox
 from .relay_helpers import (
     TypeHelper,
     add_functions,
@@ -413,6 +413,46 @@ def relay_take_grad_inp(c, _nb_indices, _indices, _values):
     return relay.concatenate(outputs, 0)
 
 
+def relay_random_initialize(c, ref_seed):
+    """Create a random state for Philox2x32 RNG.
+
+    State is a couple (key, counter).
+    key is given seed, or a default value if seed is None.
+    counter starts with 0 and is incremented after each generation batch.
+    """
+    assert ref_seed.is_constant(type(None)) or ref_seed.is_constant(int)
+    seed = ref_seed.value
+    key = relay.const(seed, 'uint32')
+    counter = relay.const(0, 'uint32')
+    rstate = relay.Tuple((key, counter))
+    return rstate
+
+
+def relay_random_uint32(c, ref_rstate, ref_shape):
+    """Generate a random tensor using Philox2x32 RNG."""
+    assert ref_shape.is_constant(tuple)
+    shape = ref_shape.value
+    relay_state = c.ref(ref_rstate)
+    # Compute output size.
+    output_size = 1
+    for dim in shape:
+        output_size *= dim
+    # Generate random uint32 values.
+    key = relay.TupleGetItem(relay_state, 0)
+    counter = relay.TupleGetItem(relay_state, 1)
+    impl = relay_philox.Philox2x32(output_size)
+    ctr = impl.generate_relay_counter_array(counter)
+    random = impl.philox_2x(ctr, key, to_float=False)
+    # Reshape vector to expected shape.
+    if shape:
+        random = relay.op.reshape(random, shape)
+    # Generate next state: same key, counter + 1
+    next_rstate = relay.Tuple((
+        key, relay.add(counter, relay.const(1, 'uint32'))))
+    # Return next state and random tensor.
+    return relay.Tuple((next_rstate, random))
+
+
 COMPLEX_MAP = {
     P.distribute: relay_distribute,
     P.transpose: relay_transpose,
@@ -443,6 +483,8 @@ COMPLEX_MAP = {
     P.universe_setitem: relay_universe_setitem,
     P.universe_getitem: relay_universe_getitem,
     P.take_grad_inp: relay_take_grad_inp,
+    P.random_initialize: relay_random_initialize,
+    P.random_uint32: relay_random_uint32,
 }
 
 
@@ -776,6 +818,9 @@ class RelayOutputConverter(Converter):
         except TypeError:
             conv_val = self(v.fields[0], t.options.get(tag))
         return TaggedValue(tag, conv_val)
+
+    def convert_random_state(self, v, t):
+        return RandomStateWrapper(tuple(el.asnumpy().item() for el in v))
 
 
 class RelayBackend(Backend):
--- a/myia/operations/prim_random_initialize.py
+++ b/myia/operations/prim_random_initialize.py
@@ -1,0 +1,36 @@
+"""Definitions for the primitive `random_initialize`."""
+
+import numpy as np
+
+from .. import xtype
+from ..lib import AbstractRandomState, standard_prim
+from . import primitives as P
+
+
+def pyimpl_random_initialize(seed):
+    """Implement `random_initialize`."""
+    return np.random.RandomState(seed)
+
+
+@standard_prim(P.random_initialize)
+async def infer_random_initialize(self, engine, seed: xtype.u32):
+    """Infer the return type of primitive `random_initialize`."""
+    return AbstractRandomState()
+
+
+__operation_defaults__ = {
+    'name': 'random_initialize',
+    'registered_name': 'random_initialize',
+    'mapping': P.random_initialize,
+    'python_implementation': pyimpl_random_initialize,
+}
+
+
+__primitive_defaults__ = {
+    'name': 'random_initialize',
+    'registered_name': 'random_initialize',
+    'type': 'backend',
+    'python_implementation': pyimpl_random_initialize,
+    'inferrer_constructor': infer_random_initialize,
+    'grad_transform': None,
+}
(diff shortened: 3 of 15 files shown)


abergeron:
> * Python implementation checks input seeds, as all values are not allowed, and generate some exceptions if bad seeds are given. It seems I can not currently compile exceptions raising with Myia, though I saw a `raise` primitive in operations. I don't yet know why compilation fails.

`raise` should work.  However I see that some exceptions use string formatting and that doesn't work.  Also, you can't have anything other than a single string in the exception args.

Alternatively, we could require the seed to be a constant and do that processing outside of myia.

> * I simplified the random generator so that it uses only one vector as state to update, instead of using a matrix as in Theano (in Theano, this should be equivalent to get random numbers with parameter `nstreams=1`). Thus, in intermediate functions, I could just return the new updated state vector (it contains only 6 values) instead of trying to update a potential huge matrix inplace. So, I removed a place where inplace update was used, but I will still need to make inplace updates when generating random numbers, as I will need to a) create the output array, and b) update each array value with a generated random scalar.

I don't get why you need to do inplace updates.  There should be a better way than creating an empty array and updating the values one by one.  Not to mention that it's going to be super slow.

notoraptor:
Hi @abergeron 

Ok, exceptions seem to work now. I removed string formatting, extra exception arguments, and fixed other things and now it does not produce errors anymore.

About inplace updates: current generator code produces 1 scalar at a time, so I need to call the generator for as many values as I need. It's something like:

```python
rstate = ...
values = []
for _ in range(n_elements):
    val, rstate = next_value(rstate)
    values.append(val)
```

But now I found a `array_scan` implementation in Myia. So, if I use it, with something like `distribute(scalar_to_array(0))` as array input, and an adequate function, I guess I could avoid create an array to be updated, and just let the scan function generate it for me, right?

My problem here is that, it seems this scan primitive is not yet used. Is it?

myia/compile/backends/pytorch.py
```diff
@@ -452,9 +477,14 @@ def from_backend_value(self, v, t):
 
     def to_backend_value(self, v, t):
         """Convert an intermediate value to a backend value."""
-        if (isinstance(t, (abstract.AbstractError, abstract.AbstractType))
+        if (isinstance(t, (abstract.AbstractError))
```
breuleux:
```suggestion
        if (isinstance(t, abstract.AbstractError)
```

myia/operations/macro_to_scalar_type.py
```diff
@@ -0,0 +1,63 @@
...
+    :param sync_data: arbitrary data to convert
+    :param default: optional default value to return.
+        If not provided and if sync_data cannot be converted,
+        then an exception is raised.
+    :return: an abstract scalar object if data can be converted,
+        or default value if provided, otherwise raise an exception.
+    """
+    if isinstance(sync_data, AbstractType):
+        return Constant(sync_data.xvalue())
+    if isinstance(sync_data, AbstractScalar):
```
breuleux: This is mostly a style convention, but I find it preferable to use `elif` here. It's a bit easier to read (you can immediately see they are mutually exclusive branches), and it behaves better if for some reason you forget a `return` statement somewhere.

myia/operations/op_full.py
```diff
@@ -0,0 +1,45 @@
...
+)
+
+
+def pyimpl_full(shape, fill_value, dtype):
+    """Python implementation for operation full."""
+    return np.full(shape, fill_value, dtype)
+
+
+@core
+def full(shape, fill_value, dtype):
```
breuleux: You should be able to have default arguments here.

```suggestion
def full(shape, fill_value, dtype=None):
```

myia/operations/op_full.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    Main code for operation full.
+
+    :param shape: a tuple of integers
+    :param fill_value: a scalar value
+    :param dtype: either a string (e.g. 'int32')
+        or a numpy dtype (e.g. np.int32)
+    :return: an array
+    """
+    abstract_scalar_type = to_scalar_type(dtype, typeof(fill_value))
```
breuleux: I'm not a huge fan of the `default` argument on `to_scalar_type`. I think it would be better to do it like this:

```suggestion
    if dtype is None:
        dtype = typeof(fill_value)
    abstract_scalar_type = to_scalar_type(dtype)
```

myia/operations/ops_dunder.py
```diff
@@ -164,6 +164,7 @@ def protocol(x, y):
     name='or', lattr='__or__', rattr='__ror__',
     pyop=operator.or_
 )
+
```
breuleux: Why are you separating xor from its friends? :(
```suggestion
```

tests/operations/test_operations.py
```diff
@@ -51,3 +54,25 @@ def test_bitwise_lshift(a, b):
 )
 def test_bitwise_rshift(a, b):
     return a >> b
+
+
+def test_op_full():
```
breuleux: Use `@mt` for this test and `test_op_prod`.

myia/operations/op_full.py
```diff
@@ -0,0 +1,45 @@
...
+@core
+def full(shape, fill_value, dtype):
+    """
+    Main code for operation full.
+
+    :param shape: a tuple of integers
+    :param fill_value: a scalar value
+    :param dtype: either a string (e.g. 'int32')
+        or a numpy dtype (e.g. np.int32)
+    :return: an array
```
breuleux: This isn't the docstyle we've been using so far (not that we've been super consistent about it). You can check in other files.

myia/abstract/infer.py
```diff
@@ -296,9 +296,19 @@ def get_inferrer_for(self, m: MacroFunction):
         argrefs = [self.ref(node, ctx) for node in n_args]
 
         if isinstance(fn, AbstractType):
+            # Replace abstract type instanciation with
```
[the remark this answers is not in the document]
notoraptor: Done!

myia/compile/backends/pytorch.py
```diff
@@ -179,11 +167,48 @@ def _impl(array):
...
         if len(tshp) < len(ashp):
             res = torch.reshape(res, shape=tshp)
         return (res,)
-    return _impl, (op.inputs[2],)
+
+    return _impl
+
+
+def _pytorch_array_reduce_mul(tshp):
+    """Generate implementation for product reduction based on givne axes."""
```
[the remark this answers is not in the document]
notoraptor: Done!

myia/abstract/amerge.py
```diff
@@ -247,10 +247,12 @@ def amerge(self, x1: TaggedPossibilities, x2, forced, bp):
 def amerge(self, x1: xtype.TypeMeta, x2, forced, bp):
     if issubclass(x2, x1):
         return x1
-    elif not forced and issubclass(x1, x2):
-        return x2
-    else:
-        raise TypeMismatchError(x1, x2)
+    elif not forced:
+        # Try to infer the smallest super-type containing both types x1 and x2.
+        super_type = xtype.number_super_set(x1, x2)
```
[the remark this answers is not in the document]
notoraptor: You're right, I needed upcast because I have to multiply an integer with a floating value in RNG, and the floating value could be either 16, 32 or 64 bits. But ultimately, I was able to solve the problem by casting integer to the right floating type using a dictionary.

myia/abstract/infer.py
```diff
@@ -296,9 +296,19 @@ def get_inferrer_for(self, m: MacroFunction):
...
+            # a make_record for all other cases.
+            val = fn.xvalue()
+            cls = type_to_abstract(val)
             g = ref.node.graph
-            newfn = g.apply(P.partial, P.make_record, fn.xvalue())
-            newcall = g.apply(newfn, *n_args)
+            if isinstance(cls, AbstractScalar):
+                newcall = g.apply(P.scalar_cast, *n_args,
+                                  AbstractScalar({VALUE: ANYTHING,
+                                                  TYPE: cls.xtype()}))
```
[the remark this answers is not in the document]
notoraptor: Done!

myia/abstract/infer.py
```diff
@@ -296,9 +296,19 @@ def get_inferrer_for(self, m: MacroFunction):
...
+            else:
+                newfn = g.apply(P.partial, P.make_record, fn.xvalue())
```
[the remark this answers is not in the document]
notoraptor: Done!

myia/compile/backends/pytorch.py
notoraptor: Done!

myia/operations/macro_to_scalar_type.py
notoraptor: Done!

myia/operations/op_full.py
notoraptor: Done!

myia/operations/op_full.py
notoraptor: Done!

myia/operations/ops_dunder.py
notoraptor: Useless line removed!

notoraptor:
Hi @breuleux , ultimately, I split again the PR into 2 branches: this one with just the code related to RNG, and another with operations full and prod : https://github.com/notoraptor/myia/tree/ops-full-and-prod

After checking the code, I realized that I finally don't use prod and full in my current RNG implementation, because, ultimately, I simplified the code so that the RNG just generates 1 scalar at a time (allowing me to avoid having to deal with any array to update). So these both ops are not currently useful here. But I still applied your suggestions to whole code before split.

Also, when trying to test full op with `@mt`, I got some issues, one with relay backend, and some related to string handling (when passing d-type as a string value). I don't yet know why I have these errors, but I could work on it later.

breuleux:
@notoraptor Okay, great! Can you create a PR with the full and prod operations?

notoraptor:
@breuleux done here ! https://github.com/mila-iqia/myia/pull/310

notoraptor:
Rebased, and commits cleaned.

notoraptor:
PR updated. Add primitives `random_initialize` and `random_uniform` + placeholder functions `myia_random_initialize` and `myia_random_uniform` that allow to generate uniform random tensors.

Actually works only with Pytorch backend.

My old RNG implementation is still here.

breuleux:
So I kind of feel that `random_uniform` as a primitive may be a bit too high-level, one issue being that it throws away 8 bits of information for each float, and there's no way for the user to recover them, so if e.g. they actually wanted a matrix of random uint32 from 0 to MAXINT, they couldn't do that. Would it be possible to have `random_uint` instead, and do the conversion to the [0, 1) range using Myia operations? This would require finding out how to generate a max range uint32 with PyTorch.

On the other hand, in the future, I think we want the RNG to become a Myia graph (once the compiler's in a better shape to handle it), so we could also just kick that can down the road.

myia/operations/prim_random_initialize.py
```diff
@@ -0,0 +1,35 @@
...
+from . import primitives as P
+
+
+def pyimpl_random_initialize(seed):
+    """Implement `random_initialize`."""
+    return np.random.RandomState(seed)
+
+
+@standard_prim(P.random_initialize)
+async def infer_random_initialize(self, engine, seed):
```
breuleux: Do we require the seed to be an integer? If so, add a type check.

myia/operations/prim_random_uniform.py
```diff
@@ -0,0 +1,58 @@
...
+    value = rstate.uniform(low, high, size)
+    return rstate, value
+
+
+@standard_prim(P.random_uniform)
+async def infer_random_uniform(self, engine,
+                               rstate: AbstractRandomState,
+                               low: AbstractScalar,
+                               high: AbstractScalar,
+                               size: AbstractTuple):
```
breuleux: I think we should standardize on the name `shape` for shape arguments.

myia/operations/prim_random_uniform.py
```diff
@@ -0,0 +1,58 @@
...
+    """Infer the return type of primitive `random_uniform`."""
+    output_scalar_type = AbstractScalar({VALUE: ANYTHING, TYPE: xtype.f32})
+    output_shape = tuple(
+        self.require_constant(e, argnum=f'"3:size[{edx}]"')
+        for edx, e in enumerate(size.elements))
+    if output_shape:
+        value_type = AbstractArray(output_scalar_type,
+                                   {SHAPE: output_shape, TYPE: xtype.NDArray})
+    else:
+        value_type = output_scalar_type
```
breuleux: It does not seem possible to generate a random array with shape `()`. In general, operations should be as regular as possible, so if the size is a tuple, the output should be an array with that shape. If the user wants a scalar they can use `array_to_scalar`.

myia/pipeline/standard.py
```diff
@@ -86,6 +86,8 @@
     np.tanh: operations.array_tanh,
     np.sum: operations.sum,
     np.prod: operations.prod,
+    public_api.myia_random_initialize: operations.random_initialize,
+    public_api.myia_random_uniform: operations.random_uniform,
```
breuleux: I'm not sure what `public_api` is for. `myia.operations` already is a public API that the user is allowed to import from.

tests/test_public_api.py
```diff
@@ -0,0 +1,58 @@
...
+
+def test_pytorch():
+    run_random_generator(run_pytorch)
+
+
+def test_relay():
+    rstate = run_random_generator(run_relay)
+    key, counter = rstate
+    assert key == 12345678
+    assert counter == 4
```
breuleux: At least for the Relay implementation, I would like to see an additional test that tests the exact values returned by the RNG with some specific seed. For example, initialize with seed 1234, generate one 2x2 matrix, check the values, generate a second 2x2 matrix, check the values. That way, if something is changed in the RNG implementation, we can verify that the same numbers come out.

myia/compile/backends/relay.py
```diff
@@ -403,6 +411,56 @@ def relay_take_grad_inp(c, _nb_indices, _indices, _values):
...
+    """Create a random state for Philox2x32 RNG.
+
+    State is a couple (key, counter).
+    key is given seed, or a default value if seed is None.
+    counter starts with 0 and is incremented after each generation batch.
+    """
+    assert ref_seed.is_constant(type(None)) or ref_seed.is_constant(int)
+    seed = ref_seed.value
+    if seed is None:
+        seed = 12345678
```
breuleux: I don't think there should be a default seed, especially not in the backend implementation.

myia/operations/prim_random_initialize.py
notoraptor: Yes, seed is the key, and it should be a uint32. So I will add a typecheck.

myia/operations/prim_random_uniform.py
notoraptor: Ok!

myia/operations/prim_random_uniform.py
notoraptor: OK. So I should check that given shape is not an empty tuple ? Because it's valid in pytorch:
```
>>> import torch
>>> torch.zeros(())
tensor(0.)
>>> torch.zeros(()).uniform_()
tensor(0.9694)
```

myia/compile/backends/relay.py
notoraptor: Ok!

myia/operations/prim_random_uniform.py
breuleux: No, it's fine if the shape is the empty tuple, but it should return an AbstractArray in that case (with shape == ()). Basically you can just remove the conditional.

notoraptor: Ok!

tests/test_public_api.py
notoraptor: Ok!

myia/pipeline/standard.py
notoraptor: Oh, ok, I did not know. So, user could just directly import operations `random_initialize` and `random_uint` and use them as functions in python code.

notoraptor:
Hi @breuleux I added 2 commits to fix all your points !

By the way, @abergeron was right, some tess (not RNG ones, but other in tests suite) won't work because of changes in TVM api. So, I guess this PR will need to wait for PR #322 at least !

myia/pipeline/standard.py
breuleux: Yes, exactly.

notoraptor:
@abergeron @breuleux PR rebased!

myia/compile/backends/pytorch.py
```diff
@@ -484,13 +505,22 @@ def from_backend_value(self, v, t):
         if isinstance(t, abstract.AbstractScalar):
             return self.to_scalar(v)
         elif isinstance(t, abstract.AbstractArray):
-            return self.to_numpy(v)
+            # Convert torch tensor to numpy tensor.
+            output = self.to_numpy(v)
+            # If possible and necessary, cast numpy tensor to expected tensor.
+            array_type = t.element.xtype()
```
abergeron: If you ever need to do that, then whatever function produced the array has a bug.  All primitives should return the right type.

myia/operations/prim_random_uint32.py
```diff
@@ -0,0 +1,62 @@
...
+async def infer_random_uint32(self, engine,
+                              rstate: AbstractRandomState,
+                              shape: AbstractTuple):
+    """Infer the return type of primitive `random_uint32`."""
+    output_scalar_type = AbstractScalar({VALUE: ANYTHING, TYPE: xtype.u32})
+    output_shape = tuple(
+        self.require_constant(e, argnum=f'"3:size[{edx}]"')
+        for edx, e in enumerate(shape.elements))
+    if not output_shape:
+        output_shape = (1,)
```
abergeron: This shouldn't be supported at the primitive level.

myia/compile/backends/pytorch.py
notoraptor: @abergeron It seems we cannot cast values to neither `uint16` nor `uint32` nor `uint64` (unsigned integers except `uint8`) in Pytorch. This code currently fails:
```python
import numpy as np
from myia import myia
from myia.operations import array_cast


@myia(backend='pytorch')
def f(val):
    return array_cast(val, np.uint32)


def main():
    val = np.arange(4, dtype='float32')
    res = f(val)
    print(res.dtype)


if __name__ == '__main__':
    main()
```

With this error:
```
  File "/home/notoraptor/mila/dev/git/myia/myia/compile/backends/pytorch.py", line 178, in pytorch_array_cast
    dt = _type_map[t.value.xtype()]

KeyError: UInt[32]
```

Indeed, `_type_map` does not handle unsigned types very well: https://github.com/mila-iqia/myia/blob/master/myia/compile/backends/pytorch.py#L16

Checking torch documentation, it seems indeed some torch dtypes are missing: https://pytorch.org/docs/stable/tensor_attributes.html#torch-dtype

myia/operations/prim_random_uint32.py
notoraptor: Ok, removed!

abergeron:
I ran coverage locally:
[report.txt](https://github.com/mila-iqia/myia/files/4232097/report.txt)

There are some spots missing.

notoraptor:
Code updated ! Except for this: https://github.com/mila-iqia/myia/pull/301#discussion_r382143774

myia/compile/backends/pytorch.py
abergeron: Ok then for this change, but this is just more motivation for me to drop the pytorch backend.

PR #303, opened 2019-11-07
New algorithm for dead data elimination

breuleux:
The current algorithm for dead data elimination relies on information propagated by the inferrer as to which set of functions may be called at which point. That information cannot be reliably tracked through optimizations, in particular through inlining of functions that contain closures (these closures are copied, but they will not be put in the abstract fields of the nodes downstream). Because of this, DDE cannot work well with live/incremental inference.

This new implementation does not use the information in abstract, therefore it should be compatible with live inference.

NOTE: This is a complete rewrite of the dde.py file, so don't look at the diff, it will just confuse you.


myia/opt/dde.py
```diff
@@ -1,226 +1,479 @@
...
+            for node2, i in self.manager.uses[node]:
+                self.todo.add(node2)
+            for other_node in self.flow[node]:
+                self._add_value(other_node, need, value)
+
+    def add_need(self, node, need):
+        """Declare that node has the given need."""
+        needs = self.need[node]
+        if isinstance(need, tuple) and len(need) > MAX_NEED_DEPTH:
+            need = need[:MAX_NEED_DEPTH]
```
abergeron: I understand that this is here to prevent running forever, but I don't see any devolving mechanism to convert to a WILDCARD need or something.  That would be more prudent.

There might be something I don't get which alleviates this requirement.

myia/opt/dde.py
```diff
@@ -1,226 +1,479 @@
...
-                                     for subp in _subslices((g, *p))))
-        return missing
+    def make_dead(self, node):
+        """Create a dead version of the node."""
+        a = node.abstract
+        val = DEAD
+        if isinstance(a, abstract.AbstractScalar):
+            if a.xtype() == xtype.EnvType:
+                val = newenv
+            assert a.xtype() != xtype.UniverseType  # TODO
```
abergeron: A workaround for now might be to just leave it alone.  It might be better than crashing.

myia/opt/dde.py
breuleux: The need `(x, y)` subsumes the need `(x, y, z)` for all `z`, so it's safe to cut the tuple to any length.

myia/opt/dde.py
breuleux: True, I could do that.

PR #306, opened 2019-11-15
rnn (lstm_cell) pt_fe support

ethancaballero:
* adds these ops to pytorch frontend: mse_loss, chunk, & lstm_cell

* should allow support for this example: https://github.com/pytorch/examples/blob/master/time_sequence_prediction/train.py
(other than the part where torch.zeros is declared inside of myia step function).


myia/frontends/pytorch_functions.py
```diff
@@ -350,6 +397,20 @@ def mean(self, dim=None, keepdim=False, *, dtype=None):
             / P.scalar_cast(x.shape[dim], x.dtype)
 
 
+@core
+def mse_loss(input, target, reduction='mean'):
+    """Map of 'nll_loss' pytorch method."""
```
abergeron:
```suggestion
    """Map of 'mse_loss' pytorch method."""
```

tests/frontends/test_pytorch_ops.py
```diff
@@ -105,7 +105,7 @@ def make_argspec(args, broad_specs):
 @myia_function_test(marks=[pytest.mark.grad], id='grad')
 def _fwd_and_bwd(self, fn, args, broad_specs=None, pipeline=standard_pipeline,
                  backend=False, atol=1e-8, rtol=1e-5,
-                 grad_atol=1e-6, grad_rtol=1e-5):
+                 grad_atol=1e-5, grad_rtol=1e-5):
```
abergeron: Can you set that only on the test that requires it instead of globally.

tests/frontends/test_pytorch_ops.py
```diff
@@ -380,6 +403,12 @@ def test_torch_mean(x):
     return torch.mean(x)
 
 
+@fwd_and_bwd_no_relay(nn.Parameter(torch.Tensor(MA(2, 3))),
+                      nn.Parameter(torch.Tensor(MB(2, 3))))
+def test_torch_mse_loss(x, y):
+    return torch.nn.functional.mse_loss(x, y)
```
abergeron: Does that really not work with relay?  I'm surprised.

PR #307, opened 2019-11-21
Bitwise ops

notoraptor:
This is a PR to add bitwise operations for scalars (and, or, xor, left shift and right shift).

This PR can be reviewed before the RNG pull request.

@breuleux @abergeron


tests/operations/test_operations.py
```diff
@@ -0,0 +1,34 @@
...
+def run_bit_xor(a, b):
+    return a ^ b
+
+
+def test_bitwise_operations():
+    assert run_lsfhit(3, 2) == 12
+    assert run_rshift(12, 2) == 3
+    assert run_bit_and(5, 7) == 5
+    assert run_bit_or(5, 2) == 7
+    assert run_bit_xor(10, 8) == 2
```
breuleux: Each operation should be tested separately. You should use the `@mt` decorator defined in `tests/multitest.py`, look in `tests/test_algos.py` for examples. Something like this, basically:

```python
@mt(
    infer(i64, i64, result=i64),  # test type inference
    run(10, 8, result=2),         # test implementation and result
    # other tests go here
)
def test_bit_xor(a, b):
    return a ^ b
```

This will also test all backends, although there is also `run_debug` and `run_no_relay`. I believe you don't have Relay support, but I think it should be as easy to add as for Pytorch, assuming Relay supports these operations at all.

You can omit the result in `run()`, in that case it'll compare to the Python result.

abergeron: Relay supports right and left shift but doesn't have the others.

notoraptor:
Hi @breuleux @abergeron I added all suggested changes.

For relay, I added left shift and right shifts, but I did not find operations for bitwise `and`, `or` and `xor`.

myia/pipeline/standard.py
```diff
@@ -137,6 +137,15 @@
...
     },
+    xtype.Integral: {
+        '__and__': lop(operations.scalar_bit_and, xtype.Integral, '__and__'),
+        '__or__': lop(operations.scalar_bit_or, xtype.Integral, '__or__'),
+        '__xor__': lop(operations.scalar_bit_xor, xtype.Integral, '__xor__'),
+        '__lshift__': lop(
+            operations.scalar_bit_lshift, xtype.Integral, '__lshift__'),
+        '__rshift__': lop(
+            operations.scalar_bit_rshift, xtype.Integral, '__rshift__'),
+    },
```
breuleux: Now that we have `Integral`, I think we can also add the entries currently under `Int` and `UInt` in here, which will remove some code duplication.

breuleux:
@notoraptor I see. Seems like an odd omission. Maybe Relay's `logical_and` and `logical_or` are actually bitwise?

notoraptor:
@breuleux Updated, I merged Int and UInt into Integral.

Relay `logical_and` and `logical_or` seem to be boolean operations. Following code fails asking for an entry to be a boolean:

```python
from tvm import relay


def test_relay():
    x = relay.var('x', dtype='int32', shape=())
    y = relay.var('y', dtype='int32', shape=())
    z = relay.logical_and(x, y)
    f = relay.Function([x, y], z)
    mod = relay.Module({})
    fname = relay.GlobalVar('f')
    mod[fname] = f
    e = relay.create_executor(mod=mod)
    c = e.evaluate(mod[fname])
    print(c(7, 5))


if __name__ == '__main__':
    test_relay()
```

breuleux:
@notoraptor Okay, good to know. We'll have to wait until these operations are available in Relay.

tests/prim/test_py_implementations.py
```diff
@@ -396,6 +400,30 @@ def test_prim_dot():
...
+    res = scalar_bit_or(5, 2)
+    assert ref == res == 7
+
+
+def test_prim_scalar_bit_xor():
+    ref = 10 ^ 8
+    res = scalar_bit_xor(10, 8)
+    assert ref == res == 2
+
+
```
breuleux: The coverage results are in, and it seems you forgot `test_prim_scalar_bit_and` :(

notoraptor:
@breuleux updated!

PR #308, opened 2019-11-22
Universe support for the Relay backend

abergeron:
This also modifies/fixes a couple of things outside of relay to make sure backend support is possible.

### DIFF
--- a/myia/api.py
+++ b/myia/api.py
@@ -43,7 +43,7 @@ class MyiaFunction:
         self.alias_tracker = alias_tracker
         self.specialize_values = set(specialize_values)
         self.pip = standard_pipeline.configure({
-            'resources.universal': True,
+            'resources.universal': use_universe,
             'resources.backend.name': backend,
             'resources.backend.options': backend_options,
             'resources.return_backend': return_backend,
--- a/myia/api.py
+++ b/myia/api.py
@@ -109,6 +109,11 @@ class MyiaFunction:
                 pass
         return self.compile(args)(*args)
 
+    def to_device(self, v, *, broaden=True, vm_t=None, orig_t=None):
+        backr = self.pip.steps['resources'].keywords['backend'].keywords
+        return to_device(v, backr['name'], backr['options'],
+                         broaden=broaden, vm_t=vm_t, orig_t=orig_t)
+
 
 @keyword_decorator
 def myia(fn, *, specialize_values=[], backend=None, backend_options=None,
--- a/myia/api.py
+++ b/myia/api.py
@@ -42,18 +42,12 @@ class MyiaFunction:
         self.fn = fn
         self.alias_tracker = alias_tracker
         self.specialize_values = set(specialize_values)
-        if use_universe:
-            from .pipeline import standard_debug_pipeline
-            self.pip = standard_debug_pipeline.configure({
-                'validate': False,
-                'resources.universal': True,
-            })
-        else:
-            self.pip = standard_pipeline.configure({
-                'resources.backend.name': backend,
-                'resources.backend.options': backend_options,
-                'resources.return_backend': return_backend,
-            })
+        self.pip = standard_pipeline.configure({
+            'resources.universal': True,
+            'resources.backend.name': backend,
+            'resources.backend.options': backend_options,
+            'resources.return_backend': return_backend,
+        })
         self._cache = {}
         self.latest = None
         if tracer is ABSENT:
--- a/myia/api.py
+++ b/myia/api.py
@@ -41,7 +41,7 @@ class MyiaFunction:
         """Initialize a MyiaFunction."""
         # Change this once relay becomes the default backend.
         if use_universe and backend != 'relay':
-            raise RuntimeError(
+            raise RuntimeError(  # pragma: no cover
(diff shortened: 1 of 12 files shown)


abergeron:
Now with 100% coverage!

myia/compile/backends/relay.py
```diff
@@ -523,24 +543,27 @@ def run(self, graph, context, target):
 
         for g in mng.graphs:
             if g.parent is None:
-                self.graph_map[g] = relay.GlobalVar(g.debug.debug_name)
+                if g is graph:
+                    self.graph_map[g] = relay.GlobalVar("main")
```
breuleux: This could collide with functions named "main" by the user, no?

myia/compile/backends/relay.py
```diff
@@ -582,32 +605,66 @@ def ref(self, node):
...
 class RelayInputConverter(Converter):
     """Convert values to Relay."""
 
     def __init__(self, context):
         """Set the context."""
         self.context = context
+        self.cst_conv = RelayConstantConverter(self.context)
+        mod = relay.Module({})
+        th = TypeHelper()
+        th.initialize(mod, DummyManager())
```
breuleux: Maybe edit TypeHelper.initialize so that the manager is optional?

myia/ir/utils.py
```diff
@@ -232,7 +232,8 @@ def sexp_to_node(sexp, graph, multigraph=False, sub=None):
     elif isinstance(sexp, ANFNode):
         return sexp
     else:
-        return Constant(sexp)
+        node = Constant(sexp)
+    return node
```
breuleux: Not sure what the point of these changes is.

myia/operations/prim_handle.py
```diff
@@ -1,14 +1,9 @@
 """Definitions for the primitive `handle`."""
 
-from ..lib import AbstractHandle, HandleInstance, standard_prim
+from ..lib import AbstractHandle, standard_prim
 from . import primitives as P
 
 
-def pyimpl_handle(init):
-    """Implement `handle`."""
-    return HandleInstance(init)
```
breuleux: Why are you removing this?

myia/compile/transform.py
```diff
@@ -79,6 +82,43 @@ def wrap_primitives(graph):
     return graph
 
 
+def gather_handles_params(params):
+    """Return list of all handles in params.
+
+    This list also has graph nodes that correspond to the handle value
+    and functions to extract the value from the arguments.
+    """
```
breuleux: That function doesn't seem to do...  much.

myia/api.py
```diff
@@ -115,6 +113,12 @@ def __call__(self, *args):
                 pass
         return self.compile(args)(*args)
 
+    def to_device(self, v, *, broaden=True, vm_t=None, orig_t=None):
+        """Move value to the function's accelerator hardware."""
+        backr = self.pip.steps['resources'].keywords['backend'].keywords
```
breuleux: I get it, but this looks pretty bad. Might be worth it to split resources into settings and resources, the former of which can be shared by all specializations in the MyiaFunction, but that's neither here nor there, I suppose.

myia/compile/backends/relay.py
abergeron: Yes it could.  However relay requires that the entry function is named "main" and that there is only one.  So we would have to mangle the "main" from the user somehow.

myia/compile/backends/relay.py
abergeron: Done.

myia/ir/utils.py
abergeron: Leftovers from something else.

myia/operations/prim_handle.py
abergeron: Because it's not called anymore.  Unless you want me to add a DebugVM test to keep it.

myia/compile/transform.py
abergeron: It did more in the beginning and might go back to doing more if I ever add back the support that was cut to get this working.

Unless you would rather we cut the whole thing for now.

myia/api.py
abergeron: The problem is not sharing between specialization (since the backend will be shared anyway), but rather access to the values before any specialization has happened.  Therefore we don't have access to a built pipeline.

myia/compile/backends/relay.py
breuleux: We could probably just mangle all function names with a `myia_` prefix.

myia/operations/prim_handle.py
breuleux: The Universe feature won't work with the DebugVM anymore, so leave it removed, I suppose.

myia/compile/transform.py
breuleux: I mean that `gather_handles_params` doesn't do anything at all. It's a docstring and then no function body.

abergeron:
Rebased

myia/compile/transform.py
abergeron: Oh right.  This is more leftovers then.

PR #310, opened 2019-11-25
[wip] Add operations np.full and np.prod

notoraptor:
TODO:
- testing
- some issues for op full with relay backend
- some issues sometimes for op full when passing string as dtypes (e.g. `'int32'`)

@breuleux

### DIFF
--- a/myia/abstract/data.py
+++ b/myia/abstract/data.py
@@ -860,17 +860,6 @@ def listof(t):
 def u64tup_typecheck(engine, tup):
     """Verify that tup is a tuple of uint64."""
     tup_t = engine.check(AbstractTuple, tup)
-    # If tup is a constant tuple of positive integers,
-    # then consider it as a tuple of uint64.
-    try:
-        if all(isinstance(el, AbstractScalar)
-               and issubclass(el.xtype(), xtype.Integral)
-               and el.xvalue() >= 0
-               for el in tup_t.elements):
-            return tup_t
-    except TypeError:
-        pass
-    # Otherwise, check merged type.
     for elem_t in tup_t.elements:
         engine.abstract_merge(xtype.UInt[64], elem_t.xtype())
     return tup_t
--- a/myia/abstract/data.py
+++ b/myia/abstract/data.py
@@ -860,6 +860,17 @@ def listof(t):
 def u64tup_typecheck(engine, tup):
     """Verify that tup is a tuple of uint64."""
     tup_t = engine.check(AbstractTuple, tup)
+    # If tup is a constant tuple of positive integers,
+    # then consider it as a tuple of uint64.
+    try:
+        if all(isinstance(el, AbstractScalar)
+               and issubclass(el.xtype(), xtype.Integral)
+               and el.xvalue() >= 0
+               for el in tup_t.elements):
+            return tup_t
+    except TypeError:
+        pass
+    # Otherwise, check merged type.
     for elem_t in tup_t.elements:
         engine.abstract_merge(xtype.UInt[64], elem_t.xtype())
     return tup_t
--- a/myia/operations/macro_to_scalar_type.py
+++ b/myia/operations/macro_to_scalar_type.py
@@ -23,14 +23,16 @@ from ..lib import (
 from ..xtype import Number, String, pytype_to_myiatype
 
 
-def convert_to_scalar_type(sync_data):
+@macro
+async def to_scalar_type(info, data):
     """
-    Main code to convert given data to abstract scalar.
+    Convert given data to abstract scalar.
 
-    :param sync_data: arbitrary data to convert
+    :param data: arbitrary data to convert
     :return: an abstract scalar object if data can be converted,
         otherwise raise an exception.
     """
+    sync_data = await data.get()
     if isinstance(sync_data, AbstractType):
         return Constant(sync_data.xvalue())
     elif isinstance(sync_data, AbstractScalar):
@@ -41,13 +43,6 @@ def convert_to_scalar_type(sync_data):
             myia_type = pytype_to_myiatype(np.dtype(sync_data.xvalue()).type)
             return Constant(AbstractScalar({VALUE: ANYTHING, TYPE: myia_type}))
     raise Exception('Unable to convert data to scalar type: %s' % sync_data)
-
-
-@macro
-async def to_scalar_type(info, data):
-    """Return a constant with the result of `data.xvalue()`."""
-    sync_data = await data.get()
-    return convert_to_scalar_type(sync_data)
 
 
 __operation_defaults__ = {
--- a/myia/operations/macro_to_scalar_type.py
+++ b/myia/operations/macro_to_scalar_type.py
@@ -8,8 +8,11 @@ If data is an abstract scalar with a string type, try to infer type
 from given string and return an abstract scalar with inferred type.
 Used in myia/operations/op_full.py
 """
+import inspect
 
 import numpy as np
+
+from myia.utils.errors import MyiaTypeError
 
 from ..lib import (
     ANYTHING,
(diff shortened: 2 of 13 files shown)


notoraptor:
PR seems fixed:
- added multitest for prod and full
- using broad_specs=(False,...) for multitesting to avoid errors when passing strings and tuples
- updated `u64tup_typecheck` to easily accept constant tuples of positive integers.

@breuleux

myia/abstract/data.py
```diff
@@ -860,6 +860,17 @@ def listof(t):
 def u64tup_typecheck(engine, tup):
     """Verify that tup is a tuple of uint64."""
     tup_t = engine.check(AbstractTuple, tup)
+    # If tup is a constant tuple of positive integers,
+    # then consider it as a tuple of uint64.
```
breuleux: That's probably not going to cause any issues, but I'm a bit hesitant that this warrants a special case.

tests/operations/test_operations.py
```diff
@@ -51,3 +53,22 @@ def test_bitwise_lshift(a, b):
...
+@mt(
+    run(np.arange(10), result=np.prod(np.arange(10))),
+    run(np.asarray([-2, -1, 2, 11], dtype='float32'), result=np.float32(44)),
+)
+def test_prod(arr):
+    return np.prod(arr)
+
+
+@mt(
+    run((2, 3), 0, None, result=np.zeros((2, 3), 'float32')),
```
breuleux: You can import `Shp` from `.common` and use `Shp(2, 3)`. That will create a u64 tuple.

tests/operations/test_operations.py
```diff
@@ -51,3 +53,22 @@ def test_bitwise_lshift(a, b):
 )
 def test_bitwise_rshift(a, b):
     return a >> b
+
+
+@mt(
+    run(np.arange(10), result=np.prod(np.arange(10))),
```
breuleux: The arange includes zero, so the result is going to be zero. I'm not sure if that's what you meant to do.

notoraptor: Indeed, I will use `np.arange(1, 11)` instead and hardcode the expected result value.

myia/abstract/data.py
notoraptor: Ok, I will remove it.

tests/operations/test_operations.py
notoraptor: Ok ! So, i am using `Shp`, but now I have issues because op_full now receives an abstract type of abstract tuple, instead of just an abstract tuple, so I am looking for a way to fix it.

breuleux: Oh yeah my bad, `Shp` is for for `infer(...)`. You can use `(np.uint64(2), np.uint64(3))`, for what it's worth.

notoraptor:
@breuleux Code updated, it should work now. I removed special case from `u64tup_typecheck`
 and used `np.uint64` to generate shapes in testing. Using np.uint64 raised many other errors in relay backend, because:
- relay converted np.uint64 to relay class `tvm.expr.UIntImm` when mapping shapes, and at some point the shape dimensions should be typecasted to builtin type `int`, which seems not implemented for `UIntImm` (although it works with `IntImm`).
- at some point, testing compares expected output shape with result shape, and this raises an error if shapes do not have same type (ie. `int` for a shape and `UIntImm` for the other)

I resolved these issues by making sure to always convert shapes to tuples of builtin integers before using them in relay backend. It seems enough to resolve the issues.

It's done in latest commit.

abergeron:
Your full gpu test problem should solve itself with #313.

I've also modified the way you "fix" shapes in relay a bit.

notoraptor:
@abergeron ok !

notoraptor:
Rebased to integrate new master commits.

notoraptor:
Updated, trying to fix coverage issue.

myia/operations/macro_to_scalar_type.py
```diff
@@ -0,0 +1,55 @@
...
+    :param data: arbitrary data to convert
+    :return: an abstract scalar object if data can be converted,
+        otherwise raise an exception.
+    """
+    sync_data = await data.get()
+    if isinstance(sync_data, AbstractType):
+        sync_data = sync_data.xvalue()
+    if isinstance(sync_data, AbstractScalar):
+        xtype = sync_data.xtype()
+        if issubclass(xtype, Number):
```
breuleux: There are two issues here:

1. This will raise a TypeError if xtype is a future. One of your tests triggers this (I put a comment under it). The test still passes, but that's because it is expecting the other TypeError, the one raised by this function.
2. If you do force the future/pending value, then `to_scalar_type(10)` will work, which it shouldn't.

tests/operations/test_operations.py
```diff
@@ -51,3 +69,71 @@ def test_bitwise_lshift(a, b):
...
+    """Test op full with wrong given d-types."""
+    for dtype, should_raise in (
+            ('float16', False),
+            (int, False),
+            (np.float16, False),
+            ('bad string', True),
+            # Bad value (integer, neither type not string)
+            (10, True),
+            # Bad value (a tuple)
+            ((), True),
```
breuleux: You can use `@mt(infer(...))` for this. For example, `infer('bad string', result=TypeError)` will check that a TypeError is raised. You should merge these tests with `test_full`.

tests/operations/test_operations.py
```diff
@@ -51,3 +69,71 @@ def test_bitwise_lshift(a, b):
...
+
+def test_full_bad_dtype():
+    """Test op full with wrong given d-types."""
+    for dtype, should_raise in (
+            ('float16', False),
+            (int, False),
+            (np.float16, False),
+            ('bad string', True),
+            # Bad value (integer, neither type not string)
+            (10, True),
```
breuleux: This test raises a TypeError, but it's the *wrong* TypeError (see other comment).

tests/operations/test_operations.py
```diff
@@ -51,3 +69,71 @@ def test_bitwise_lshift(a, b):
...
+    infer(Shp(2, 3), f64, 'int16', result=ai16_of(2, 3)),
+    infer(Shp(2, 3), f64, 'int32', result=ai32_of(2, 3)),
+    infer(Shp(2, 3), f64, 'uint64', result=au64_of(2, 3)),
+    # Numpy d-types should also be accepted as d-types.
+    infer(Shp(2, 3), i64, np.int16, result=ai16_of(2, 3)),
+    infer(Shp(2, 3), i64, np.float16, result=af16_of(2, 3)),
+    infer(Shp(2, 3), f64, np.uint64, result=au64_of(2, 3)),
+)
+def test_infer_full(shape, value, dtype):
+    return np.full(shape, value, dtype)
```
breuleux: Why are the inference tests listed separately from test_full?

notoraptor: When I put inference tests in test_full, I got the following error:

`TypeError: infer() got an unexpected keyword argument 'broad_specs'`

I pass `broad_specs` to the `mt` call decorating `test_full`, and it seems `infer` does not receive such argument, so I need to separate inference tests and standard tests.

I need to use `broad_specs` to make tests work correctly. otherwise I got other issues, as if the strings are not sent correctly to the operations.

breuleux: Oh, that's true. I'm not sure what's the best approach to solve this, so let's keep it this way for now.

notoraptor:
Hi @breuleux all your comments should have been addressed:

- tests for wrong type values were added to `test_infer_full`
- to_scalar_type was rewritten to make sure `to_scalar_type(10)` won't be accepted.

myia/operations/macro_to_scalar_type.py
```diff
@@ -0,0 +1,90 @@
...
+    Convert given data to abstract scalar.
+
+    :param data: arbitrary data to convert
+    :return: an abstract scalar object if data can be converted,
+        otherwise raise an exception.
+    """
+    sync_data = await data.get()
+
+    # We expect either:
+    # - an abstract string containing an abstract scalar with scalar type
```
breuleux:
```suggestion
    # - an abstract type containing an abstract scalar with scalar type
```

notoraptor: Done!

PR #311, opened 2019-11-26
Pytorch frontend op cross_entropy

notoraptor:
Hi @breuleux @abergeron @ethancaballero 

This is a PR to implement pytorch frontend operation `cross_entropy`.

@ethancaballero I did not yet try to implement all cross entropy function parameters (weight, size_average, ignore_index, reduce and reduction), I just implemented input and target parameters. Could you check if I modified right files in right places ? Should I implement all parameters ?

Thanks!

### DIFF
--- a/myia/frontends/pytorch.py
+++ b/myia/frontends/pytorch.py
@@ -5,7 +5,6 @@ import types
 from collections import OrderedDict
 
 import torch
-import torch.nn.functional
 
 from .. import operations
 from ..abstract.data import (
--- a/myia/frontends/pytorch.py
+++ b/myia/frontends/pytorch.py
@@ -5,6 +5,7 @@ import types
 from collections import OrderedDict
 
 import torch
+import torch.nn.functional
 
 from .. import operations
 from ..abstract.data import (
@@ -36,6 +37,7 @@ from .pytorch_functions import (
     cat,
     chunk,
     conv2d,
+    cross_entropy,
     gather,
     item,
     linear,
@@ -98,6 +100,7 @@ standard_object_map.update({
     torch.var: var,
     # torch.zeros_like: C.zeros_like,  # currently only works with pt backend
     torch.nn.functional.conv2d: conv2d,
+    torch.nn.functional.cross_entropy: cross_entropy,
     torch.nn.functional.linear: linear,
     torch.nn.functional.max_pool2d: max_pool2d,
     torch.nn.functional.mse_loss: mse_loss,
--- a/myia/frontends/pytorch_functions.py
+++ b/myia/frontends/pytorch_functions.py
@@ -19,7 +19,6 @@ import operator
 from functools import reduce
 
 import torch
-import torch.nn.functional
 
 from .. import operations
 from ..abstract import myia_static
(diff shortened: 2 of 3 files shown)


myia/frontends/pytorch.py
```diff
@@ -5,6 +5,7 @@
 from collections import OrderedDict
 
 import torch
+import torch.nn.functional
```
ethancaballero: you don't need to import torch.nn.functional

myia/frontends/pytorch_functions.py
```diff
@@ -19,6 +19,7 @@
 from functools import reduce
 
 import torch
+import torch.nn.functional
```
ethancaballero: you don't need to import torch.nn.functional

tests/frontends/test_pytorch_ops.py
```diff
@@ -311,6 +311,17 @@ def test_torch_conv2d__group3(inp, w, b):
     return torch.sum(value)
 
 
+@mt(
+    fwd_and_bwd(nn.Parameter(torch.Tensor(torch.randn(3, 5))),
+                nn.Parameter(torch.randint(5, (3,)),
+                             requires_grad=False)),
+    run(torch.randn(3, 5), torch.randint(5, (3,))),
+    backend=backend_no_relay
+)
```
ethancaballero: This should be:
```
@mt(
    fwd_and_bwd(nn.Parameter(torch.Tensor(torch.randn(3, 5))),
                nn.Parameter(torch.randint(5, (3,)),
                             requires_grad=False)),
)
```

myia/frontends/pytorch_functions.py
```diff
@@ -268,6 +269,14 @@ def conv2d(input, weight, bias=None, stride=1, padding=0, dilation=1,
     return ret
 
 
+@core
+def cross_entropy(input, target):
+    """Map of method torch.nn.functional.cross_entropy."""
+    a = torch.nn.functional.log_softmax(input, 1)
+    b = torch.nn.functional.nll_loss(a, target)
+    return b
```
ethancaballero: This should be:
```
@core
def cross_entropy(input, target, reduction='mean'):
    """Map of method torch.nn.functional.cross_entropy."""
    a = torch.nn.functional.log_softmax(input, 1)
    b = torch.nn.functional.nll_loss(a, target, reduction=reduction)
    return b
```

ethancaballero:
The additional parameter that needs to be added to cross_entopy is `reduction`. I've added how to so in the line comment of mine.

myia/frontends/pytorch.py
notoraptor: Done!

myia/frontends/pytorch_functions.py
notoraptor: Done!

tests/frontends/test_pytorch_ops.py
notoraptor: Done! I used `fwd_and_bwd_no_relay` instead of just `fwd_and_bwd`, because I got some compilation errors with relay, and I guess we don't need relay to work here as we are in Pytorch frontend, do we?

I also duplicated the test for each currently supported reduction (none, mean and sum)

myia/frontends/pytorch_functions.py
notoraptor: Done!

PR #314, opened 2019-12-03
Add pytorch frontend operation `detach`.

notoraptor:
@ethancaballero @breuleux @abergeron 

This is a PR to add Pytorch frontend operation `detach`.


breuleux:
* I don't think our primitive should be called `detach`. That's PyTorch terminology, and not a very good one. I would prefer `stop_gradient` as in TensorFlow, or something else equally descriptive.
* After the gradient transform, `detach` is functionally identical to the identity function, and it should be removed from the graph. An optimization that removes `detach` should be added to `step_opt2`, which occurs after gradient expansion.
* Given that it should always be optimized always, `detach` should not be a `BackendPrimitive`. It should be a `PlaceholderPrimitive` like `J` or `Jinv`. `PlaceholderPrimitive` is for primitives that serve a purpose after inference, but should be removed or optimized away before they get to the backend.

notoraptor:
Hi @breuleux , I updated the code based on your suggestion!

PS: Should primitive `identity` be considered as a placeholder, too ?

tests/frontends/test_pytorch_ops.py
```diff
@@ -330,6 +330,20 @@ def test_torch_cross_entropy(inp, target, reduction):
     return F.cross_entropy(inp, target, reduction=reduction)
 
 
+@mt(
+    fwd_and_bwd_no_relay(nn.Parameter(torch.Tensor(torch.randn(7, 3)))),
+    fwd_and_bwd_no_relay(nn.Parameter(torch.Tensor(torch.randn(5, 8)))),
+    fwd_and_bwd_no_relay(nn.Parameter(torch.Tensor(torch.randn(10)))),
```
abergeron: This should run with relay.

notoraptor: Done!

notoraptor:
I also removed stop_gradient mapping from pytorch backend, as the primitive should be removed by optimizations and then no go until backend.

breuleux:
@notoraptor Good question. Yes, `identity` should probably be the same. You could try to change it and see if that breaks anything.

myia/frontends/pytorch.py
```diff
@@ -116,6 +118,7 @@
     'abs': operations.array_abs,
     'argmax': argmax,
     'chunk': chunk,
+    'detach': detach,
```
breuleux: You can just set this directly to `operations.stop_gradient`

myia/validate.py
```diff
@@ -90,6 +90,7 @@ def validate_abstract(self, a: (type(None), AbstractExternal), uses):
     P.bool_and,
     P.bool_or,
     P.bool_eq,
+    P.stop_gradient,
```
breuleux: PlaceholderPrimitives should not be in the whitelist (if you rebase on top of current master, the whitelist was removed and replaced by a check for BackendPrimitive, so we won't need to keep updating this list anymore).

myia/frontends/pytorch.py
notoraptor: Done!

myia/validate.py
notoraptor: Rebased!

notoraptor:
@breuleux I also added a commit to make `identity` a placeholder primitive, to check if all tests run correctly with that modification.

notoraptor:
All tests passed, so making `identity` a placeholder does not seem to create any problem.

I added another commit to fix coverage issue.

PR #315, opened 2019-12-05
Add scalar instantiation and casting (like int(val) or np.float16(val))

notoraptor:
Hi, @breuleux @abergeron .

This PR should add support for scalar instantiation and casting (when doing `x = int(val)` or `x = np.float16(-2)`). I extracted code from my branch about RNG, and extended it to support relay backend.


myia/compile/backends/relay_helpers.py
```diff
@@ -180,6 +181,12 @@ def to_relay_type(self, a: AbstractScalar):
         return relay.ty.scalar_type(type_to_np_dtype(tp))
 
 
+@overload  # noqa: F811
+def to_relay_type(self, a: AbstractType):
+    # Let's just replace abstract type object by type wrapped inside it.
```
breuleux: That's a bit fishy. As far as I understand, values with type AbstractType are not really used, so I think we can just flat out replace it with an int type.

myia/operations/prim_scalar_cast.py
```diff
@@ -34,7 +34,8 @@ def pyimpl_scalar_cast(x, t):
     if not isinstance(a, AbstractScalar):
         raise MyiaTypeError('scalar_cast must cast to a scalar type')
     t = a.xtype()
-    engine.check(xtype.Number, t)
+    # Let's accept both numbers and boolean.
+    engine.check((xtype.Number, xtype.Bool), t)
```
breuleux: Casting a number to bool is currently implemented as `x != 0` and I'm not sure we want two ways of doing it. Also, this allows int->bool, but bool->int will not work (because scalar is still required to only have type Number).

myia/abstract/infer.py
```diff
@@ -353,7 +361,11 @@ def check_predicate(self, predicate, x):
             * A Myia type (xtype.Int[64] etc.)
             * A Python class
             * A callable that returns a boolean
+            * A tuple of values listed above. In such case, we just need
+              one of given predicates to apply to x.
         """
+        if isinstance(predicate, tuple):
+            return any(self.check_predicate(pred, x) for pred in predicate)
         if isinstance(predicate, xtype.TypeMeta):
```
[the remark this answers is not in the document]
notoraptor: Done!

myia/compile/backends/pytorch.py
```diff
@@ -475,9 +475,14 @@ def from_backend_value(self, v, t):
 
     def to_backend_value(self, v, t):
         """Convert an intermediate value to a backend value."""
-        if (isinstance(t, (abstract.AbstractError, abstract.AbstractType))
+        if (isinstance(t, abstract.AbstractError)
                 or v is abstract.DEAD):
             return None
+        if isinstance(t, abstract.AbstractType):
```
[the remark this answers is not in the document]
notoraptor: Done!

myia/compile/backends/relay.py
```diff
@@ -645,6 +646,14 @@ def convert_tagged(self, v, t):
         conv_val = self(v.value, real_t)
         return interpreter.ConstructorValue(ctr.tag, [conv_val], None)
 
+    def convert_type(self, v, t):
+        # v is already a scalar type
+        # (ie. numpy type or Python builtin numeric type)
+        # and relay does not want anything but scalar or ndarray in his graph,
```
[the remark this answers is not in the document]
notoraptor: Done

myia/abstract/infer.py
```diff
@@ -296,9 +296,17 @@ def get_inferrer_for(self, m: MacroFunction):
...
+            # Replace abstract type instantiation with
+            # either a cast for abstract scalars, or
+            # a make_record for all other cases.
+            val = fn.xvalue()
+            cls = type_to_abstract(val)
             g = ref.node.graph
-            newfn = g.apply(P.partial, P.make_record, fn.xvalue())
-            newcall = g.apply(newfn, *n_args)
+            if isinstance(cls, AbstractScalar):
+                newcall = g.apply(P.scalar_cast, *n_args, cls)
```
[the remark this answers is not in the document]
notoraptor: Done

myia/compile/backends/relay_helpers.py
notoraptor: OK!

myia/operations/prim_scalar_cast.py
notoraptor: Ok, I will remove that !

myia/abstract/infer.py
```diff
@@ -296,9 +296,23 @@ def get_inferrer_for(self, m: MacroFunction):
...
+            cls = type_to_abstract(val)
             g = ref.node.graph
-            newfn = g.apply(P.partial, P.make_record, fn.xvalue())
-            newcall = g.apply(newfn, *n_args)
+            if isinstance(cls, AbstractScalar):
+                typ = cls.xtype()
+                if issubclass(typ, xtype.Number):
+                    newcall = g.apply(P.scalar_cast, *n_args, cls)
+                elif typ is xtype.Bool:
+                    newcall = g.apply(P.scalar_ne, *n_args, 0)
```
breuleux: Sorry, I wasn't totally clear. This should map to the `bool` operator,  which normally calls `obj.__bool__` and will produce `x != 0` on int/float.

```suggestion
                    newcall = g.apply(operations.bool, *n_args)
```

notoraptor: OK !

PS: I am also trying to write the macro `call_object`, but I got other issues. I call the macro in infer code using `g.apply(operations.call_object, fn, *n_args)`, but then it wrap the abstractype into another abstract type again (ie., macro receives `AbstractType(AbstractType(AbstractScalar))` instead of just `AbstractType(AbstractScalar)`). I'm trying to figure out why this happens.

breuleux: I assume that's because it's wrapping it with `to_abstract`. Try using `n_fn` instead of `fn`.

notoraptor: Thanks !

notoraptor:
@breuleux Updated with new macro `call_object`.

breuleux:
@notoraptor Can you rebase on current master? Thanks!

notoraptor:
@breuleux Rebased !

PR #316, opened 2019-12-06
Add pytorch frontend operation torch.nn.embedding.

notoraptor:
@breuleux @abergeron @ethancaballero 

I added pytorch frontend operation `torch.nn.embedding` using new primitives.


ethancaballero:
Are P.embedding and P.grad_embedding_weights primitives actually needed to add embedding op to pytorch frontend? (They might be, I'm just not sure). @breuleux @abergeron

notoraptor:
Hi ! So, ultimately, should I rename anything in this PR ? Maybe just `take_grad_weights` to `take_grad_inputs` ?

breuleux:
@notoraptor I think `take_grad_weights` ought to be made into a more useful primitive like `put(array, indices, values)`, such that `take_grad_weights(weights, indices, dout) <=> put(zeros_like(weights), indices, dout)`.

notoraptor:
Hi @breuleux !

Currently, `take_grad_weights(weights, indices, dout)` computes, for each index in given matrix `weights`, the sum of rows in `dout` where that index was used. This is how I compute the gradient of embedding wrt/ weights.

So, if I try to replace it with a more independent primitive, I guess I just need matrix shape (not even a zero-like matrix), and primitive should be something like:


```
sum_usages(nb_indices, row_size, indices, values)
=>
return a matrix with shape (nb_indices, row_size)
```

NB: `values` should have shape `indices.shape` + `(row_size,)`


What do you think ?

Is such primitive reusable anywhere else ?

breuleux:
@notoraptor My point in replacing it with an operation like `put` is that `put` can be used to set rows in a non-zero matrix, so e.g. `put(weights, (4, 3), values)` would do something similar to `weights[4] = values[0]; weights[3] = values[1]`. For the gradient of `take` you set the first argument to zeros, but in other uses, that need not be the case.

breuleux:
@notoraptor Can you rebase this on top of the current master branch? There are changes to the pydocstyle version and configuration and we need to verify that this code passes the new criteria.

notoraptor:
@breuleux Rebased !

notoraptor:
Tests passed !

notoraptor:
PS: Well, not yet, after all ... But it seems whole code is covered ...

PR #326, opened 2020-02-17
(wip) Add more scalar ops in both backends (&, |, ^, ~, sin, cos, trunc)

notoraptor:
Based on PR #322 so that new TVM release can be used easily. Add scalar operations:
- bitwise and
- bitwise or
- bitwise xor
- bitwise not
- sin
- cos
- trunc

Bitwise operations left shift and right shift were already implemented.


abergeron:
I've checked the coverage locally and it's fine.  So we can merge this.

PR #327, opened 2020-02-24
Myia CNN example and Policy Gradient Example

ethancaballero:
- also adds part of public_api (myia library)


myia/public_api.py
```diff
@@ -388,21 +388,16 @@ def item(x):
...
-
     if dim is None:
         dim = _build_fwd_tuple(x.shape)
     elif isinstance(dim, int):
         dim = (dim,)
-    elif isinstance(dim, tuple):
-        pass
-    else:
-        assert keepdim is False
-        return P.array_map(P.scalar_max, x, dim)
```
abergeron: This adds the 2-array compare case, why do you remove it?

myia/public_api.py
```diff
@@ -21,13 +21,13 @@
 import numpy as np
 import torch
```
abergeron: If this is to become the public myia API it can't have a dependency on PyTorch.

As it clearly still does, I don't approve of the rename.

ethancaballero:
this is now subsumed by https://github.com/mila-iqia/myia/pull/340

breuleux:
As Ethan said: subsumed by #340

PR #328, opened 2020-02-24
Misc improvements (quinvicies)

breuleux:
This fixes a few bugs at various places in the code and adds some tracing and debugging.


debug/butest.py
```diff
@@ -237,10 +237,10 @@ def pytest_runtest_setup(item):
...
-        def __init__(self):
-            super().__init__(None)
+        def __init__(self, skip=None):
+            super().__init__(None, skip=skip)
 
-        def set_trace(self, frame=None):
+        def set_trace(self, frame=None, skip=None):
             interactor.show(synchronous=True)
             self.repl = interactor.repl
             super().set_trace(frame)
```
abergeron: Should you pass on the skip argument here?

breuleux: As it turns out, `skip` is an argument to the Pdb/Bdb constructor, not to `set_trace`, except within pytest which does some black magic with pdb, but `super().set_trace(frame, skip=skip)` won't work here.

PR #329, opened 2020-02-25
Implementation of einsum

abergeron:
This doesn't support diag/traces, but should support everything else.  If we ever add support for diag, then I can probably add support for them in the macro.

There is also a small fix to torch.max to support an extra argument pattern.

### DIFF
--- a/myia/operations/macro_einsum.py
+++ b/myia/operations/macro_einsum.py
@@ -45,32 +45,59 @@ def _tensordot(g, a, b, *, axes):
     return (g.apply(P.reshape, res, res_shp), res_shp)
 
 
+def _reduce_transpose(g, input_spec, output_spec, arg):
+    if input_spec == output_spec:
+        return arg
+
+    out_idx = set(output_spec)
+    reduce_axes = [i for i, c in enumerate(input_spec) if c not in out_idx]
+    shp = arg[1]
+
+    target_shape = [s if i not in reduce_axes else 1
+                    for i, s in enumerate(shp)]
+    res = g.apply(P.array_reduce, P.scalar_add, arg[0], tuple(target_shape))
+
+    for i in reversed(reduce_axes):
+        del target_shape[i]
+    res = g.apply(P.reshape, res, tuple(target_shape))
+
+    mid_spec = [c for c in input_spec if c in out_idx]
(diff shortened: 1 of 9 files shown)


tests/operations/test_macro_einsum.py
```diff
@@ -0,0 +1,105 @@
...
+import pytest
+
+from myia.lib import InferenceError
+from myia.operations import einsum
+
+from ..common import MA, MB
+from ..multitest import infer, mt, run_debug
+
+
+@run_debug(MA(4, 1)[0])
```
breuleux: `MA(4, 1)[0]` returns a vector of size 1, did you mean `MA(1, 4)[0]`? (Same comment for the tests below)

tests/operations/test_macro_einsum.py
```diff
@@ -0,0 +1,105 @@
...
+    return einsum('ij->j', a)
+
+
+@run_debug(MA(3, 4))
+def test_einsum_sum2d_01(a):
+    return einsum('ij->', a)
+
+
+@run_debug(MA(3, 4), MB(3, 4))
+def test_einsum_elemwise2d(a, b):
```
breuleux: This is not an elemwise operation.

myia/operations/macro_einsum.py
```diff
@@ -0,0 +1,181 @@
+"""Implementation of the 'einsum' macro."""
+
+from opt_einsum import contract, contract_expression
+
+from ..lib import InferenceError, macro
+from . import primitives as P
+
+
+def _tensordot(g, a, b, *, axes):
```
breuleux: Could you write some light documentation for this and the other private functions in this module? It'd help a little for review/maintenance.

tests/operations/test_macro_einsum.py
abergeron: Yes

tests/operations/test_macro_einsum.py
abergeron: Yes, that should be ij on output.

myia/operations/macro_einsum.py
abergeron: Done

PR #332, opened 2020-03-06
Add tan to relay backend.

notoraptor:
@abergeron @breuleux 

This PR adds `tan` to relay backend.

**NB**:

TVM API changed again, and now it seems `tvm.relay.Module` does not exist anymore. So, I guess we would need to update whole relay backend again to support new TVM API before merging this pull request.

On my computer, I use this code to compile relay with current TVM code:

```python
import tvm
from tvm import relay
from tvm.contrib import graph_runtime


# Old way, does not work anymore currently.
def compile_function_old(inputs, output):
    f = relay.Function(list(inputs), output)
    m = relay.Module({})
    fn = relay.GlobalVar('f')
    m[fn] = f
    e = relay.create_executor(mod=m)
    c = e.evaluate(m[fn])
    return c


# New way. I don't know if another way exists, but this works now.
def compile_function(inputs, output):
    ir = relay.Function(list(inputs), output)
    mod = tvm.IRModule.from_expr(ir)
    target = tvm.target.create('llvm')
    graph, lib, params = relay.build(mod, target=target)
    ctx = tvm.cpu()
    module = graph_runtime.create(graph, lib, ctx)

    def fn(*args):
        for i in range(len(inputs)):
            module.set_input(i, args[i])
        module.set_input(**params)
        module.run()
        out = module.get_output(0)
        return out

    return fn
```

### DIFF
--- a/myia/compile/backends/relay.py
+++ b/myia/compile/backends/relay.py
@@ -61,6 +61,7 @@ SIMPLE_MAP = {
     P.scalar_tanh: relay.op.tanh,
     P.scalar_sin: relay.op.sin,
     P.scalar_cos: relay.op.cos,
+    P.scalar_tan: relay.op.tan,
     P.scalar_trunc: relay.op.trunc,
     P.scalar_sign: relay.sign,
     P.scalar_abs: relay.abs,
--- a/tests/operations/test_operations.py
+++ b/tests/operations/test_operations.py
@@ -114,6 +114,19 @@ def test_cos(a):
 
 
 @mt(
+    # seems to not accept integers
+    run(0., result=0),
+    run(np.float16(0), result=0),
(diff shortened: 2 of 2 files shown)


abergeron:
This seems fine, but it'll need a new tvm build for it to work.

I'll do that, but I'm waiting for a bug in the VM to be fixed.

notoraptor:
OK !

abergeron:
#334 Supports the new API, it also comes with newer packages that integrate the tan changes in TVM so once that one is merged, this one should be good to go.

breuleux:
#334 is merged, so you can rebase.

notoraptor:
@breuleux rebased !

notoraptor:
Tests succeed !

PR #333, opened 2020-03-09
Fix the tracer/profiler for accuracy.

abergeron:
This also collapses some of the hierarchy where it makes sense.

### DIFF
--- a/tests/utils/test_trace.py
+++ b/tests/utils/test_trace.py
@@ -125,9 +125,22 @@ def test_focus():
     }
 
 
+def badsleep():
+    with tracer('sleep') as tr:
+        with tracer('fence', profile=False) as tr:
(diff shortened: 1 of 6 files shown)


tests/utils/test_trace.py
```diff
@@ -125,10 +125,23 @@ def test_focus():
...
+            with tracer('stare', at='ceiling') as tr:
+                tracer().emit_time(hours=5)
+
+
 def test_profile():
     with Profiler():
         day()
 
+    with Profiler():
+       badsleep()
```
breuleux:
```suggestion
        badsleep()
```

abergeron:
The failure is jenkins tripping over itself just before the end it seems.  It did run coverage and it's fine.  All the tests pass.

PR #334, opened 2020-03-13
Enable the relay VM

abergeron:
This enables usage of the relay VM.  It means new TVM packages and follows the API changes required.

It also sequences the execution in the backend directly to avoid nesting issues that were preventing the VM from working properly in certain edge cases.  This may have a performance impact (positive or negative) on some graphs.

It also changes the way grad environments are implemented, but I don't believe this change is essential for the relay VM to work.  If there are concerns about that it can be reverted or split into a separate PR.

All in all, the VM itself is much faster than the interpreter we were using previously so I expect this to be a net positive for all tasks.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,4 @@ prettyprinter
 ruamel.yaml
 ruamel.yaml.clib>=0.2
 pydocstyle==5.0.1
+antlr4-python3-runtime
--- a/myia/compile/backends/__init__.py
+++ b/myia/compile/backends/__init__.py
@@ -34,9 +34,9 @@ def channel_loader(pkg, name):
     return loader
 
 
-def relay_defaults(target='cpu', device_id=0):
+def relay_defaults(target='cpu', device_id=0, exec_kind='vm'):
     """Format options for relay."""
-    return dict(target=target, device_id=device_id)
+    return dict(target=target, device_id=device_id, exec_kind=exec_kind)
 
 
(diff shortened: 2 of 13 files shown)


abergeron:
I just noticed that this breaks the universe stuff (which is kind of unavoidable because the VM doesn't support references for now).  Those tests are not tagged with "relay" so I did not run them when checking.

I could add an option to use the interpreter for these or we could remove the universe support.  I am not partial to one solution so please voice your opinion.

abergeron:
@breuleux @ethancaballero @notoraptor

breuleux:
I'd say temporarily have an option to use the interpreter until the Relay VM supports references, if that's not too much trouble.

abergeron:
I've added the option and updated the tests.  I hope the VM will grow to support references soon.

abergeron:
The jenkins failure is due to #336.

Separately from that, there is a lack of coverage that I need to address.

requirements.txt
```diff
@@ -4,3 +4,4 @@ prettyprinter
 ruamel.yaml
 ruamel.yaml.clib>=0.2
 pydocstyle==5.0.1
+antlr4-python3-runtime
```
breuleux: What is this for?

abergeron: The relay text parser.  The VM loads some ressources from a text file that needs to be parsed.

breuleux: Shouldn't that dependency be declared by the tvm/relay package?

PR #339, opened 2020-03-25
Reformat the code using Black

breuleux:
I'm reformatting the entire codebase using Black, which should take care of a good chunk of common flake8 issues. Simply run `black .` to reformat all Python files in the directory. The way it reformats imports is compatible with `isort`, so it should all work together.

I have added a `pyproject.toml` file in which I put the configuration for `black`. I also moved `.isort.cfg` and `.coveragerc` into `pyproject.toml` so that we can have as much configuration as possible in a single file.


breuleux:
I only thought of it this morning (sorry @notoraptor :(), but there's a trick to rebase on top of this easily, which is to:

1. Checkout the `black` branch locally
2. For every file you have modified, `git checkout master -- file.py` to revert it to the master version
3. Commit these reverts to your local `black` branch
4. Rebase on top of that commit with `git rebase black`, that should yield no conflicts thanks to the revert.
5. Run `black` to fix the formatting on your own files. That will basically redo the changes of this PR, but on the new code.

And then you should be good to go.

notoraptor:
@abergeron @breuleux this PR is ok for me !

ethancaballero:
idk, most of my changes are in file called public_api.py based on pytorch_function.py and test_all_ops.py based on test_pytorch_ops.py

breuleux:
@ethancaballero If your changes are in new files then it should be fine, you will just have to run `black` on them to fix the formatting.

PR #340, opened 2020-03-29
cnn and policy_gradient examples; numpy inputs compatibility for public_api

ethancaballero:


myia/public_api.py
```diff
@@ -523,10 +527,16 @@ def norm(inp, p=None, dim=None):
...
+    if P.hastype(inp, APT):
+        # Convert p to input type.
+        p = P.scalar_to_array(P.scalar_cast(p, inp.dtype), APT)
+        # Create a value 1 with input type.
+        one = P.scalar_to_array(P.scalar_cast(1, inp.dtype), APT)
+    else:
+        # Convert p to input type.
+        p = P.scalar_to_array(P.scalar_cast(p, inp.dtype), AA)
+        # Create a value 1 with input type.
+        one = P.scalar_to_array(P.scalar_cast(1, inp.dtype), AA)
```
breuleux: I think this should work:

```suggestion
    # Convert p to input type.
    p = P.scalar_to_array(P.scalar_cast(p, inp.dtype), operations.typeof(inp))
    # Create a value 1 with input type.
    one = P.scalar_to_array(P.scalar_cast(1, inp.dtype), operations.typeof(inp))
```

myia/public_api.py
```diff
@@ -603,7 +613,8 @@ def reshape(x, *shp):
 def scatter(self, dim, index, src):
     """Map of 'scatter' pytorch method."""
     if not P.hastype(src, APT):
-        src = P.scalar_to_array(P.scalar_cast(src, self.dtype), APT)
+        if P.hastype(self, APT) or P.hastype(src, AS):
+            src = P.scalar_to_array(P.scalar_cast(src, self.dtype), APT)
```
breuleux: Same idea as above.

```suggestion
        src = P.scalar_to_array(P.scalar_cast(src, self.dtype), operations.typeof(self))
```

myia/public_api.py
```diff
@@ -603,7 +613,8 @@ def reshape(x, *shp):
 def scatter(self, dim, index, src):
     """Map of 'scatter' pytorch method."""
     if not P.hastype(src, APT):
```
breuleux: Consider using this instead of APT:

```python
AA = AbstractArray(
    AbstractScalar({TYPE: ANYTHING, VALUE: ANYTHING}),
    {SHAPE: ANYTHING, TYPE: ANYTHING},
)
```

The `TYPE: ANYTHING` should make it work for any kind of array, whether it's an ndarray or a Tensor.

myia/public_api.py
```diff
@@ -19,15 +19,14 @@
...
-from .. import operations
-from ..abstract import myia_static
-from ..hypermap import hyper_map
-from ..operations import primitives as P
-from ..utils import MyiaValueError, core
-from ..xtype import TupleT, f32, i64, u64
-from .pytorch_abstract_types import APT, APT_bool
+from . import operations
+from .abstract import myia_static
+from .frontends.pytorch_abstract_types import AA, APT, AS, APT_bool
```
breuleux: Consider using this instead of APT_bool:

```python
AA_bool = AbstractArray(
    AbstractScalar({TYPE: Bool, VALUE: ANYTHING}),
    {SHAPE: ANYTHING, TYPE: ANYTHING},
)
```

myia/public_api.py
ethancaballero: Proposed change doesn't seem to work. It throws this error:
```/Users/ethancaballero/mila-iqia/myia_outer/myia/myia/public_api.py:614
in scatter(
    self :: Float[32] x 3 x 4,
    dim :: <PendingFromList finished result=Int[64]> = 0,
    index :: Int[64] x 2 x 4,
    src :: Float[32] x 2 x 4
)
614: src = P.scalar_to_array(P.scalar_cast(src, self.dtype), operations.typeof(self))
================================================================================
in scalar_cast(Float[32] x 2 x 4, Ty(AbstractScalar(Float[32])))
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
MyiaTypeError: Expected Number, not PyTorchTensor
```

myia/public_api.py
ethancaballero: this suggestion doesn't seem to necessarily work.

myia/public_api.py
breuleux: Which test?

examples/policy_gradient.py
```diff
@@ -0,0 +1,234 @@
...
+
+import gym
+import numpy as np
+import torch
+from numpy.random import RandomState
+
+from myia import ArithmeticData, myia, value_and_grad
+from myia.api import to_device
+
+# The following import installs custom tracebacks for inference errors
```
breuleux, line 17: Travis didn't publish the CI results here which I'm looking into, but it did run: https://travis-ci.org/github/mila-iqia/myia/jobs/669909314

The error is because `isort` and `black` have an incompatibility when there are comments in the middle of imports. `black` puts a newline, `isort` puts none. I say we just remove the comment.

```suggestion
```

PR #341, opened 2020-04-03
Split the manager into infer_manager and opt_manager

breuleux:
This splits the single manager used in the pipeline up to now into two managers, `infer_manager` and `opt_manager`, the former for the parse/inference part of the pipeline, the latter for optimization. Monomorphize takes the graphs that are in the `infer_manager` and creates monomorphized versions to handle with `opt_manager`.

This separation makes it possible to reuse the pipeline's managers to perform the gradient expansion, since the graphs added during inference won't interfere with the ones in the `opt_manager`. This, in turn, means that we can reuse cached graphs that are used in multiple grad expansions, like `zeros_like`.

Here is a further list of changes:

* The monomorphizer will no longer change any graphs in place, because of the manager split (the reuse_existing option is removed).
* The monomorphizer can now work incrementally: it can accept new graphs.
* The manager gains a `set_opaque_condition` method. Using this method, the manager can be told to ignore certain nodes. This is used when we need to perform inference on new graphs that refer to graphs that are in `opt_manager`, because we don't want the `infer_manager` to see them.
* The "quarantine" hack in `clone` is no longer necessary and is removed.
* `test_grad::test_free[var/graph]_outside_grad` are no longer xfail.
* There is now a `validator` resource, which can easily be called elsewhere in the pipeline for debugging purposes.
* The `inferrer` resource is split into three resources, `inferrer`, `live_inferrer` and `monomorphizer`, because they don't really need to be together anymore.
* The `resolve` step is removed from the standard inferrer because it is not compatible with the new manager structure, but since some tests require it, it is added to the pipelines for these specific tests that don't perform inference.

**Performance improvements**: not sure on the whole test suite, but on examples/rnn, the time spent in gradient expansion goes down from 4.2s to 3.2s on my machine, and there are similar gains on other tests. The maximal number of nodes in the manager goes down from a peak of 2,500 before these changes to about 1,700, because more graphs can be reused.

### DIFF
--- a/myia/ir/manager.py
+++ b/myia/ir/manager.py
@@ -560,6 +560,7 @@ class GraphManager(Partializable):
         if allow_changes is None:
             allow_changes = self.manage
         self.allow_changes = allow_changes
+        self.check_opaque = None
         self.reset()
 
     def clear(self):
@@ -612,6 +613,10 @@ class GraphManager(Partializable):
         for root in roots:
             self.add_graph(root, root=True)
 
+    def set_opaque_condition(self, fn):
+        """Set a condition on nodes that should not be included."""
+        self.check_opaque = fn
+
     def add_graph(self, graph, root=False):
         """Add a graph to this manager, optionally as a root graph."""
         if root:
@@ -647,7 +652,9 @@ class GraphManager(Partializable):
         """Ensure that the graph is managed by this manager."""
         if self.manage:
             if graph._manager and graph._manager is not self:
-                raise ManagerError("A graph can only have one manager.")
+                raise ManagerError(
+                    "A graph can only have one manager. " f"Graph: {graph}"
+                )
             graph._manager = self
         self.graphs.add(graph)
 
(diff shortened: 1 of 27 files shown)


myia/ir/manager.py
```diff
@@ -730,9 +731,12 @@ def limit(x):
 
         acq = OrderedSet()
         for node in nodes:
-            new_nodes = OrderedSet(dfs(node, succ_deeper, limit))
-            self.all_nodes |= new_nodes
-            acq |= new_nodes
+            acq |= OrderedSet(dfs(node, succ_deeper, limit))
+
+        if self.check_opaque:
+            acq = [n for n in acq if not self.check_opaque(n)]
```
abergeron, line 737: I'm worried that this can balloon for large graph clusters.

It might be better to add a limit on the dfs that respects check_opaque().

breuleux: You're right, I should've put this in the `limit` function that's defined right above this. I made a commit to fix this, we'll see if everything passes.

PR #342, opened 2020-04-03
Add update_tag option to HyperMap

breuleux:
This expands the interface to HyperMap to have a way to only process dataclass fields that are annotated with a specific tag. For example:

```python
from myia.utils import tags

@dataclass
class ConstYPoint:
    x: tags.Update
    y: object

add_update = HyperMap(fn_leaf=scalar_add, update_tag=tags.Update)

assert add_update(ConstYPoint(5, 10), 1) == ConstYPoint(6, 10)
```

Only the field marked with `tags.Update` will be updated. This works recursively and on all dataclasses, but only if the `update_tag` option is given to `HyperMap`.

Note that if you have nested dataclasses, updating fields in the nested dataclass requires tagging both that field and the dataclass with the update tag. In other words, if you want to update `x.y.z`, then both `x.y` and `x.y.z` must be annotated with `tags.Update`. If only `x.y.z` is marked, it won't work.

### DIFF
--- a/myia/hypermap.py
+++ b/myia/hypermap.py
@@ -48,6 +48,7 @@ class HyperMap(MetaGraph):
         trust_union_match=False,
         infer_value=False,
         name=None,
+        update_tag=None,
     ):
         """Initialize a HyperMap."""
         if name is None:
@@ -68,6 +69,7 @@ class HyperMap(MetaGraph):
             abstract.Possibilities,
             abstract.TaggedPossibilities,
         )
+        self.update_tag = update_tag
 
     async def normalize_args(self, args):
         """Return broadened arguments."""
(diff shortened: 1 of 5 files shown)


myia/hypermap.py
```diff
@@ -367,10 +382,19 @@ def _leafcall(args):
...
-                args = [
-                    getattr(x, name) if nonleaf else x for x, nonleaf in argmap
-                ]
-                results[name] = _reccall(args)
+                if (
+                    self.update_tag
+                    and not main.__annotations__[name] is self.update_tag
+                ):
+                    (arg0, nonleaf), *_ = argmap
+                    assert nonleaf
```
notoraptor, line 390: Hi @breuleux , sorry, I forgot to review this PR ! It's ok for me ! Just a question: why arg0 must not be a leaf ? (same above at line 259)

breuleux: That's because `arg0` is where we get the data for fields that don't have the update tag, and leaves don't have these fields by definition.

So if we only update field `x` of `Point(x, y)`, this means we allow `Point(x, y) + z == Point(x + z, y)` but we don't allow `z + Point(x, y)`. In theory, we *could* allow it, but I think it is better (and simpler) to always get unchanged fields from the first argument than to have a more complicated rule like "it comes from the first non-leaf argument." We can always revise this if there is a need.

notoraptor: Ok !

PR #343, opened 2020-04-07
Improved interning and caching

breuleux:
This PR is based on #341 

* The interned versions of data structures are now cached inside a field on the object
* Caching is added to fix_types and jelim using the standard CloneState and CheckState

This significantly improves performance of several tests. For the RNN test, the compilation time goes from ~10s (master) to ~9s (#341) to ~7s (this PR). Running `test_algos` is about 2x faster than it is on master, although some of the savings appear to be because there is less performance degradation from running many tests one after the other (which is also good). The test suite overall is something like 1.2x faster.


breuleux:
The PR is no longer based on #341

breuleux:
Rebased on the current master

PR #338, opened 2020-03-23
(wip)(do not merge) Run all tests on relay

notoraptor:
Missing operations in relay backend to make all tests pass:
- array_setitem
- conv_transpose2d
- gather
- scatter
- scatter_add

I am currently working on `conv_transpose2d`. Issues:
- Seems limited to only `groups=1` and `dilation=(1, 1)`: https://github.com/apache/incubator-tvm/blob/master/python/tvm/relay/op/strategy/x86.py#L169
- Persistent error when running `pytest -xvvs tests/frontends/test_pytorch_ops.py::test_torch_conv2d` on this branch:

```log
(myia) notoraptor@notoraptor-linux:~/mila/dev/git/myia$ pytest -xvvs tests/frontends/test_pytorch_ops.py::test_torch_conv2d
=========================================================================================== test session starts ============================================================================================
platform linux -- Python 3.7.6, pytest-5.4.1, py-1.8.1, pluggy-0.13.1 -- /home/notoraptor/anaconda3/envs/myia/bin/python
cachedir: .pytest_cache
rootdir: /media/win/Users/notoraptor/mila/dev/git/myia, inifile: pytest.ini
plugins: cov-2.8.1
collected 6 items                                                                                                                                                                                          

tests/frontends/test_pytorch_ops.py::test_torch_conv2d[relay-cpu-grad0] <- tests/multitest.py ANTLR runtime and generated code versions disagree: 4.8!=4.7.2
ANTLR runtime and generated code versions disagree: 4.8!=4.7.2
Cannot find config for target=llvm, workload=('conv2d_NCHWc.x86', ('TENSOR', (2, 6, 4, 5), 'float32'), ('TENSOR', (3, 6, 3, 3), 'float32'), (2, 3), (3, 2, 3, 2), (1, 1), 'NCHW', 'NCHW', 'float32'). A fallback configuration is used, which may bring great performance regression.
FAILED

================================================================================================= FAILURES =================================================================================================
____________________________________________________________________________________ test_torch_conv2d[relay-cpu-grad0] ____________________________________________________________________________________

test = <tests.multitest.MyiaFunctionTest object at 0x7f7570a72650>

    def runtest(test):
>       test.run(fn)

tests/multitest.py:66: 
[...]

PR #346, opened 2020-04-14
Add relay implementation to fully support conv2d_transpose in relay backend

notoraptor:
This PR extracts code from #338 and adds supplementary code to fix `conv2d_transpose` in relay backend. It is currently implemented as a full relay graph and seems to correctly support all values for groups, dilations, and other parameters.

This PR needs relay operation `dilate` to be added in TVM backend. I made a PR for that: https://github.com/apache/incubator-tvm/pull/5331

This way to do was more easy than changing all conv2d_transpose implementations in TVM backend. If everything is confirmed and all tests pass, I may work on hardcoding the relay graph implementation directly into TVM backend.

@breuleux @abergeron @fosterrath-mila

### DIFF
--- a/myia/compile/backends/pytorch.py
+++ b/myia/compile/backends/pytorch.py
@@ -381,9 +381,7 @@ def pytorch_conv2d(op):
 def pytorch_conv_transpose2d(op):
     """Implementation of conv_transpose2d."""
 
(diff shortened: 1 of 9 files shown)


notoraptor:
All tests succeed!

PR #347, opened 2020-04-23
Add lambda lifting step

breuleux:
This adds a `llift` step that performs lambda lifting. That is to say, every closure which is only found in call position is modified so that it takes its free variables as additional arguments.


myia/opt/llift.py
```diff
@@ -0,0 +1,82 @@
...
+                new_node.abstract = node.abstract
+                tr.replace(node, new_node)
+
+    with mng.transact() as tr:
+        for entry in todo.values():
+            # Redirect the fvs to the parameter (those in scope)
+            for fv, param in entry.fvs.items():
+                for node, idx in mng.uses[fv]:
+                    if node.graph in entry.graph.scope:
+                        tr.set_edge(node, idx, param)
```
abergeron, line 76: I see that you have a test for multi-level nesting, but I just don't see how it can work.  I probably just don't understand it.

Can you try to explain?

abergeron: I'll try to document what we discussed for future generations:

This is the starting graph:
```
 def f1(x, y):
     def g(z):
         return x + z

     def h():
         return g(y)

     return h()
```

The first loop adds parameters for all the call sites like this:
```
 def f1(x, y):
     def g(z, x_):
         return x + z

     def h(y_, x_):
         return g(y, x)

     return h(y, x)
```

Then the second loop then swaps the freevars for the new args:
```
 def f1(x, y):
     def g(z, x_):
         return x_ + z

     def h(y_, x_):
         return g(y_, x_)

     return h(y, x)
```

PR #349, opened 2020-04-28
pytorch_1.5_compatibility

ethancaballero:


tests/frontends/test_pytorch.py
```diff
@@ -418,85 +418,6 @@ def forward(self, x):
         return x
 
 
-def test_module_2_layer_mlp_seq_fwd():
```
abergeron, line 421: Please don't delete tests like this.  Mark them as xfail or something like that.

If you really think they have no more value, then please explain why.

requirements-gpu.conda
```diff
@@ -2,14 +2,14 @@ python>=3.7,<3.8a0
 colorama
 numpy
 opt_einsum
-cudatoolkit=10.0.*
+cudatoolkit=10.2.*
 cudnn
 abergeron::tvm==0.7dev1+0.*
```
abergeron, line 7: TVM version should be bumped to `0.7dev1+1.*`.  That is the one that has cuda 10.2 packages.

PR #350, opened 2020-04-28
Run backend on same process as frontend.

notoraptor:
@breuleux This is a pull request to remove call to sub-processes when compiling graph.

I will let all tests run to check if there are some problems.


notoraptor:
@abergeron @breuleux Updated. I think it's done, but I needed to temporarily deactivate some GPU tests because they fail and @ethancaballero seems to work on it on PR #349 :
- tests.frontends.test_all_ops.py::test_torch_scatter_broadcast_source
- tests/frontends/test_pytorch.py::test_module_2_layer_mlp_seq_fwd
- tests/frontends/test_pytorch.py::test_module_linear_seq_bwd

(fails seems related to pytorch 1.5)

Once we find a way to fix these tests, I could reactivate them here before merging.
NB: In his PR, @ethancaballero just removed those tests. Maybe can I do the same ?

tests/test_parser.py
```diff
@@ -220,37 +226,17 @@ def c5():
...
 
 
-def test_no_return__format(capsys):
+def test_no_return__format():
     def f():
         pass
 
     try:
         parse(f)
-    except MyiaSyntaxError:
```
abergeron, line 229: This doesn't have anything to do with the multi-process.  It is testing that exception printout for myia parser errors give the proper stack trace on output.

Can you revert that back to what it was (this whole file)?

notoraptor: Hi @abergeron I can revert, but this will make tests fail.

notoraptor:
Hi @breuleux @abergeron ! Just to say that everything seems to work here.
- I fixed TVM using this branch : https://github.com/apache/incubator-tvm/compare/master...notoraptor:fix-excepthook
- then I compiled tvm packages and uploaded them on my own conda channel: https://anaconda.org/notoraptor/tvm and https://anaconda.org/notoraptor/tvm-libs 
- and I use those packages in this branch (see requirement files)

So, this PR still cannot be merged until some more official TVM packages are used, but apart from that, it seems fine.

I also removed uncovered code.

myia/compile/backends/__init__.py
```diff
@@ -52,11 +52,11 @@ def pytorch_default(device="cpu:0"):
 
 _backends = {
     "relay": (
-        channel_loader("myia.compile.backends.relay", "RelayBackendR"),
+        channel_loader("myia.compile.backends.relay", "RelayBackend"),
```
abergeron, line 55: You might want to rename channel_loader to something else.

requirements-cpu.conda
```diff
@@ -2,7 +2,7 @@ python>=3.7,<3.8a0
 colorama
 numpy
 opt_einsum
-abergeron::tvm==0.7dev1+1.*
+notoraptor::tvm==0.7dev1+1.*
```
abergeron, line 5: You need to change the version if you change the code (the +1 should be +2).  Your +1 should be the same as my +1.

myia/compile/backends/__init__.py
notoraptor: Renamed to backend_loader.

requirements-cpu.conda
notoraptor: By the way, this is changes applied to tvm: https://github.com/apache/incubator-tvm/compare/master...notoraptor:fix-excepthook

Should I still recompile and rename my package to ` notoraptor::tvm==0.7dev1+2.*` ?

notoraptor:
@abergeron rebased, channel_loader renamed to backend_loader, and yaml and ruamel removed from requirements.

requirements-cpu.conda
abergeron: Yes, otherwise it can easily create problems where people don't know which version of the code they have exactly and it makes it harder to debug problems.

This is why it's not a good practice to release two packages with the same version and different code.

abergeron:
I think you forgot an import with your rebase and it makes the random tests fail.

requirements-cpu.conda
notoraptor: Ok, so I will recompile and re-upload a tvm package with version `+2`!

notoraptor:
Indeed, fixed!

abergeron:
I'm ok with the changes here, the only thing we should wait for is the TVM PR.

notoraptor:
@abergeron Ok !

notoraptor:
@abergeron updated!

PR #351, opened 2020-05-04
Forward references

breuleux:
This PR adjusts the parser so that closures can refer to functions or variables defined after themselves.

### DIFF
--- a/myia/parser.py
+++ b/myia/parser.py
@@ -260,7 +260,8 @@ class Parser:
         tree = asttokens.ASTTokens(src, parse=True).tree
         function_def = tree.body[0]
         assert isinstance(function_def, ast.FunctionDef)
-        main_block = self._process_function(None, function_def)
+        main_block, _finalize = self._create_function(None, function_def)
+        _finalize()
         for node in dfs(main_block.graph.return_, succ_deeper):
             if node.is_constant_graph():
                 if node.value.return_ is None:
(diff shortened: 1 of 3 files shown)


abergeron:
This seems like it's a bit too permissive, because this works:

```python
def test_forward_reference_in_closure2():
    def g(x):
        def h():
	    return a * a

	r = h()
        a = x * x
	return r

    parse(g)
```

It might be ok to let it work, but it could cause problems where python and myia don't interpret something the same way like:

```python
def test_forward_reference_in_closure3():
    def g(x):
        def h():
	    return a * a

        a = x
	r = h()
        a = x * x
	return r

    parse(g)
```

abergeron:
Maybe we should call the finalizer if a function is called before the end of the block to solve this?

But then again python allows this:

```python
def main(x):
    def h():
        return a * a

    a = x
    r = h()
    a = x * x
    return r, h

r, h = main(2)
print(r, h())  # prints "4 16"
```

breuleux:
@abergeron Ah, yeah, I forgot about this issue. Not sure what the best solution is.

breuleux:
Also, technically, Myia already had this issue:

```python
def f(x):
    a = x * x
    def g():
        return a
    a = a * a
    return a, g()

print(f(4))        # => (256, 256)
print(myia(f)(4))  # => (256, 16)
```

Ironically, this PR fixes that particular example, because `g` will now take the last version of `a`.

PR #353, opened 2020-05-04
Remove unused parameters in functions

breuleux:
This add an optimization phase that removes unused parameters in functions, when possible.

The manager gains `call_sites` and `higher_order_sites` to track where a function is called/used.

Some code is not covered anymore since e.g. dead parameters can be removed.

### DIFF
--- a/myia/validate.py
+++ b/myia/validate.py
@@ -21,7 +21,7 @@ from .abstract import (
 )
 from .operations import Primitive
 from .operations.primitives import BackendPrimitive
-from .utils import ErrorPool, Partializable, overload
+from .utils import ErrorPool, Partializable, overload, untested_legacy
 
 
 class ValidationError(Exception):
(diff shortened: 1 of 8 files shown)


myia/validate.py
```diff
@@ -41,15 +41,16 @@ def validate_abstract(self, a: (AbstractClass, AbstractJTagged), uses):
...
-        raise ValidationError(f"Illegal type in the graph: {a}", type=a)
+    with untested_legacy():
+        kind = a.xvalue()
+        if kind is DEAD:
+            return True
+        elif kind is POLY:
+            return not any(key == 0 for node, key in uses)
+        else:  # pragma: no cover
+            # As it turns out, the inferrer now catches this error before
+            # we get to validation.
```
abergeron, line 52: You might want to remove that pragma

breuleux: Okie dokie. Done.

PR #355, opened 2020-05-05
Changes to the universe mechanism

breuleux:
* The universe argument is moved to the end of the parameter list, rather than the beginning, in order to be compatible with partial application.
* The `handle` primitive becomes `make_handle` and is changed.
  * It does not set an initial value in the universe. `universe_setitem` must be called to do so.
  * `make_handle` takes a universe and a type and returns a new universe and a handle, which allows us to order calls to `make_handle` and guarantee that CSE cannot merge similar handles.
  * A graph transform is applied in the Relay backend so that it matches the behavior of `RefCreate`.
* The `make_cell` composite behaves like `handle` used to.
* The `handle_get` and `handle_set` composites are renamed to `cell_get` and `cell_set`.
* The feature works in more situations, and more tests are added.


abergeron:
The only question I have is why not have make_cell as the primitive?  Otherwise I'm ok.

breuleux:
> The only question I have is why not have make_cell as the primitive?

Essentially, because it is simpler for optimization purposes if only one primitive sets things in the universe. For example, we can simplify `getitem(setitem(U, h, v), h)` to `v`, but if `make_cell` was a primitive we would also need to consider `getitem(make_cell(U, v)[1], h)`, and so on.

PR #356, opened 2020-05-05
Add AbstractFunctionUnique to replace AbstractFunction after monomorphization

breuleux:
`VirtualFunction` essentially becomes `AbstractFunctionUnique` which inherits from `AbstractValue`. After monomorphization, `AbstractFunction` is removed from the graph and all functions have `AbstractFunctionUnique` as their type.

`AbstractFunctionUnique` does not contain a `Possibilities` list, just a plain list of argument types and an output type. As such it is much closer to a conventional representation of a function type. This simplifies processing in the optimizer and in the backend.

Note that in the future, we might want to replace `AbstractFunction` during inference by an `AbstractUnion` of `AbstractFunctionUnique`.

`JTransformedFunction(fn)` is also replaced by `TransformedFunction(fn, P.J)` which creates opportunities to have other transforms, like `TransformedFunction(fn, P.Jinv)` to represent the inverse.

### DIFF
--- a/myia/abstract/data.py
+++ b/myia/abstract/data.py
@@ -38,8 +38,6 @@ VOID = Named("VOID")
 # Represents specialization problems
 DEAD = Named("DEAD")
 register_serialize(DEAD, "DEAD")
-DUMMY = Named("DUMMY")
-register_serialize(DUMMY, "DUMMY")
 POLY = Named("POLY")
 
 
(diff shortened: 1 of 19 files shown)


myia/abstract/data.py
```diff
@@ -63,15 +63,6 @@ def __init__(self, options):
         # list is different.
         super().__init__(OrderedSet(options))
 
-    def _serialize(self):
```
abergeron, line 66: You should remove the `@serializable` tag from the class too.

breuleux: Ah, yes, good catch.

breuleux: Done.

PR #357, opened 2020-05-06
Add a tracing garbage collector to the manager

breuleux:
As it turns out, the manager needs a garbage collector in some situations. It does have a GC of sorts which is based on refcounts of graph uses, but if there is a cycle somewhere, removing a pointer to the whole cycle won't reclaim it.

A `gc` method is therefore added to the manager which basically checks the `graphs_reachable` statistic from the roots and resets the manager if this is not the same as the current set of graphs. In theory it should be possible to do better than a reset, but there is unfortunately an issue with a new test in future PR #359 (which is not available yet, but it will come in its own time).

### DIFF
--- a/myia/ir/manager.py
+++ b/myia/ir/manager.py
@@ -612,7 +612,7 @@ class GraphManager(Partializable):
         Normally this is done incrementally through reference counting, but
         because of circular references, some graphs might remain.
         """
-        reach = set(self.roots)
+        reach = OrderedSet(self.roots)
         for root in self.roots:
             reach.update(self.graphs_reachable[root])
         # TODO: Ideally the two lines below should be replaced by the commented
--- a/myia/ir/manager.py
+++ b/myia/ir/manager.py
@@ -603,6 +603,7 @@ class GraphManager(Partializable):
             allow_changes = self.manage
         self.allow_changes = allow_changes
         self.check_opaque = None
+        self.events = None
         self.reset()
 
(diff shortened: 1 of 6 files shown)


myia/ir/manager.py
```diff
@@ -583,13 +583,31 @@ def __init__(self, *roots, manage=True, allow_changes=None):
...
+        """
+        reach = OrderedSet(self.roots)
+        for root in self.roots:
+            reach.update(self.graphs_reachable[root])
+        # TODO: Ideally the two lines below should be replaced by the commented
+        # out line, but it causes an error in
+        # tests/test_grad.py::test_recursive_closure
+        # # self._drop_all(self.graphs - reach, drop_nodes=True)
+        if reach != self.graphs:
+            self.reset()
```
abergeron, line 603: I'm wondering if you could modify the check in _maybe_drop_graph() to use graphs_reachable instead.  But that might trigger the same bug referred to above.

Also, that bug should be investigated to figure out why it happens.  A full reset is non trivial and apparently involves deleting all registered events.

breuleux: `graphs_reachable` is kind of expensive, we want to avoid computing it unless necessary, especially on the root graph.

A full reset is, well, rare, so I thought it might be acceptable.

I think the bug only happens on a test which is in another PR that I'm preparing. I suppose I could look deeper into it.

PR #358, opened 2020-05-07
Extend lambda lifting to the branches of switch

breuleux:
This extends lambda_lift to perform the following transform:

```python
def f(x, y, z):
    def true_branch():
        return y

    def false_branch():
        return z

    return switch(x < 0, true_branch, false_branch)()

# ==>

def f(x, y, z):
    def true_branch(_y, _z):    # <- Add free variables of both branches
        return _y

    def false_branch(_y, _z):   # <- Add free variables of both branches
        return _z

    return switch(x < 0, true_branch, false_branch)(y, z)   # <- Add args
```

This transform is able to remove free variables in a lot more situations than before.


breuleux:
I have also extended removing unused parameters to branches of switch.

The `GraphInterfaceRewriter` class implements a generic framework for lambda lifting, removing unused parameters and generally any optimization that modifies the external interface of a graph (number of parameters, type of output, and so on).

myia/opt/rewrite.py
```diff
@@ -1,70 +1,472 @@
...
+            chg = rewriter.run()
             if chg:
                 tracer().emit_success(**args, new_node=None)
             return chg
 
 
-__all__ = ["RewriteGraphs", "rewrite_graphs"]
+##########################
+# LAMBDA LIFTING EXAMPLE #
+##########################
```
abergeron, line 412: This should probably go with the LambdaLiftRewriter above

myia/opt/rewrite.py
```diff
@@ -1,70 +1,472 @@
...
+        if not getattr(self.filter, "noop", False):
+            new_entries = dict(entries)
+            for g, entry in entries.items():
+                if g in new_entries and not self.filter(entry, entries):
+                    for gg in entry.eqv:
+                        del new_entries[gg]
+            entries = new_entries
+
+        # 3. Sort the entries, if operations must be done in a specific order
+        # if self.order_key is NotImplemented:
```
abergeron, line 169: This comment seems like leftover code.

myia/opt/rewrite.py
breuleux: OK

myia/opt/rewrite.py
breuleux: Yeah. It just... blended in, so I missed it :P

PR #360, opened 2020-05-13
Change gradient generation around free variables

breuleux:
The generation of the gradient around free variables is modified in order to support recursive closures. More specifically, we use `free_variables_extended` to generate an entry in the env for the gradient of the closure for each non-graph free variable instead of packing some of them into entries that are themselves environments.

A test is added for recursive closures. That test would not work before this PR -- mutually recursive nodes would be created by the grad function, which constitutes an invalid computation graph.

### DIFF
--- a/myia/grad.py
+++ b/myia/grad.py
@@ -441,8 +441,9 @@ class SensRemapper(GradRemapper):
         fv_sens = Constant(newenv)
         for fv in g.free_variables_total:
             sens = self.get(g, fv)
-            # NOTE: If sens is an application of zeros_like, it would be
-            # possible to skip adding it to the env.
+            if sens.is_apply(zeros_like):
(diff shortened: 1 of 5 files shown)


myia/grad.py
```diff
@@ -319,7 +319,7 @@ def gen_child(self, g, ng, child):
         with About(child.debug, self.relation):
             self.remap_node((g, child), g, child, ng, ng.apply())
 
-    def gen_fv(self, g, ng, node):
+    def gen_fv_extended(self, g, ng, node):
         """Generate sensitivities for free variables.
 
         Note that the default gen_fv does nothing, so this is different
```
abergeron, line 325: Maybe this comment should refer to gen_fv_extended?

breuleux: True

PR #361, opened 2020-05-19
Allow a RNG state to be returned from a compiled function and then used as argument for another compiled function.

notoraptor:
@breuleux @abergeron

This PR should allow to return an RNG state from a compiled function and then use it as argument for other compiled functions. Python <-> Myia communication should be solved now.

### DIFF
--- a/myia/simplify_types.py
+++ b/myia/simplify_types.py
@@ -35,7 +35,6 @@ from .compile import BackendValue
 from .ir import Constant
 from .operations import primitives as P
 from .utils import HandleInstance, MyiaInputTypeError, TaggedValue, overload
-from .utils.misc import RandomStateWrapper
 from .xtype import Int, NDArray, String
 
 ####################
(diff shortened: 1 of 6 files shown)


myia/simplify_types.py
```diff
@@ -303,6 +303,12 @@ def to_canonical(self, arg, orig_t: AbstractTuple, coerce):
     return tuple(self(x, o, coerce) for x, o in zip(arg, oe))
 
 
+@overload  # noqa: F811
+def to_canonical(self, arg, orig_t: AbstractRandomState, coerce):
+    # arg must be a RandomStateWrapper
+    return arg
```
breuleux, line 309: You need to check that arg actually is a RandomStateWrapper. Once a function is compiled for the first time, `to_canonical` is used to validate the arguments.

notoraptor: Done!

PR #362, opened 2020-05-19
analogous_to_modules

ethancaballero:


abergeron:
I'm not waiting for the branch build to finish since the pr build finished.  BTW, you get duplicate builds like this when you push a branch on the main repo and not your fork.

abergeron:
Please resubmit this PR and fix the coverage.  I've reverted it for now with #363

breuleux:
The lines that are not covered would be compiled and run by Myia in the examples, so I think it is probably sufficient to add a `# pragma: no cover` comment for the corresponding functions.

PR #368, opened 2020-07-09
Add remaining ops to relay (gather, scatter, scatter_add) and run all tests on relay.

notoraptor:
`gather` and `scatter` already added. Fixed `conv_transpose2d` and `argmax` too (in relay backend).

Now adding `scatter_add` to TVM.

### DIFF
--- a/myia/compile/backends/relay.py
+++ b/myia/compile/backends/relay.py
@@ -519,6 +519,11 @@ def relay_scatter(c, inp, dim, index, src):
     return relay.scatter(c.ref(inp), c.ref(index), c.ref(src), dim.value)
 
 
+def relay_scatter_add(c, inp, dim, index, src):
+    assert dim.is_constant(int)
+    return relay.scatter_add(c.ref(inp), c.ref(index), c.ref(src), dim.value)
+
+
 def relay_make_cell(c, v, u):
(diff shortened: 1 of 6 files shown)


myia/compile/backends/relay.py
```diff
@@ -263,7 +263,7 @@ def relay_argmax(c, v, dims):
     """Implementation of argmax for Relay."""
     v = c.ref(v)
     assert dims.is_constant(tuple)
-    return relay.cast(relay.argmax(v, axis=dims.value), "int64")
+    return relay.cast(relay.argmax(v, axis=dims.value, keepdims=True), "int64")
```
abergeron, line 266: I'm not sure an unconditional keepdims is ok here.

I'm also not sure what behaviour we expect out of argmax wrt reduced dimension for our op.

notoraptor: Hi @abergeron ! It seems `infer_shape` function for primitive argmax expects all dimensions to be here and reduced dimensions with size 1: https://github.com/mila-iqia/myia/blob/master/myia/operations/prim_argmax.py#L17

So, currently, I guess keepdims is necessary.

abergeron: ok

PR #369, opened 2020-08-11
Prepare backends as plugins.

notoraptor:


notoraptor:
Hi @abergeron @breuleux this is a PR to set backends as plugins, available as entry points under namespace `myia.backend`. What do you think ?

environment-gpu.yml
```diff
@@ -0,0 +1,23 @@
...
+- pillow<7
+- pip
+- cudatoolkit=10.2.*
+- cudnn=7.6.*
+- pip:
+  - asttokens>=2.0.3,<3.0.0
+  - prettyprinter>=0.18.0,<0.19.0
+  - black>=19.10b0,<19.11.0
+  - pydocstyle==5.0.1
+  - gym>=0.17.1,<0.18.0
```
abergeron, line 23: I don't think we need the gpu stuff for the frontend.

myia/compile/backends/__init__.py
```diff
@@ -20,46 +22,87 @@ class LoadingError(Exception):
...
 
 
-def backend_loader(pkg, name):
-    """Helper function for simple backends.
+class BackendLoader:
+    """Utility class to load a backend."""
 
-    This will return a callable that will load a module, retrieve a
-    object from its namespace and return that.
+    __slots__ = ("load_options", "load_backend")
```
abergeron, line 28: Why does this have a `__slots__` declaration?  It doesn't harm anything but I'm curious.

myia_backend_pytorch/scripts/update_env.py
```diff
@@ -0,0 +1,55 @@
...
+            if isinstance(element, dict):
+                dependencies = (
+                    dependencies[:index] + packages + dependencies[index:]
+                )
+                packages_added = True
+                break
+        if not packages_added:
+            dependencies.extend(packages)
+        data["dependencies"] = dependencies
+    data.pop("name", None)
```
abergeron, line 45: Why do you remove the env name for the update?

scripts/ci-install.sh
```diff
@@ -26,10 +26,26 @@ conda update -q conda
...
+
+# Install myia and backend plugins using poetry.
+poetry install
+cd myia_backend_pytorch
+poetry install
+cd ../myia_backend_relay
+poetry install
+cd ..
+
+# Complete installation with specific conda packages using environment files.
```
abergeron, line 45: It might be a good idea to regenerate the env files here to make sure that the pyproject files have been properly updated.

scripts/update_env.py
```diff
@@ -0,0 +1,55 @@
...
+    output = dump(data, Dumper=Dumper)
+    if args.output:
+        with open(args.output, "w") as file:
+            file.write(output)
+    else:
+        print(output)
+
+
+if __name__ == "__main__":
+    main()
```
abergeron, line 55: It might be nice to not have three copies of this script around.

environment-gpu.yml
notoraptor: Hi ! I keep everything related to pytorch-gpu because pytorch backend is still in myia main package (ie. , even if `myia_backend_pytorch` is not installed, `myia` itself would still need GPU packages so that pytorch frontend can work).

Should I better change something here ?

myia/compile/backends/__init__.py
notoraptor: Sorry, an habit ! `__slots__` helps to lock list of allowed attributes and help python to reduce memory footprint (by denying `__dict__` and allocate only fields declared in slots).

But it is useless here, I will remove it.

scripts/ci-install.sh
notoraptor: OK !

scripts/update_env.py
notoraptor: I can create a special package, something like `myia_utils`, and move this file into it.

myia_backend_pytorch/scripts/update_env.py
notoraptor: When I keep the name, the script produces following env output for relay backend:
```
dependencies:
- python>=3.7,<4.0
- numpy>=1.16,<2.0
- scipy>=1.5,<2.0
- decorator>=4,<5
- attrs>=19,<20
- pip
- mila-iqia::tvm=0.7dev1+5.*
- pip:
  - antlr4-python3-runtime==4.7.2
name: myia_backend_relay
```

The name is at the end, and, I don't know why, but tvm package is never installed in such case.
When I remove the name field, everything is installed correctly.

I guess name field is not mandatory, so I think I can remove it without issues.

notoraptor:
@abergeron Updated !

environment-gpu.yml
abergeron: But the backend has its own requirements list (which should include the gpu stuff).  I don't see why you need it here.

myia/compile/backends/__init__.py
abergeron: I knew what it does, but I was just curious why it was here.

myia_backend_pytorch/scripts/update_env.py
abergeron: It's at the end because the yaml dumper sorts keys by default.  You can disable that with `sort_keys=False`.  If you have another reason to remove it though, it's fine.

scripts/update_env.py
abergeron: Since all the packages are within the myia repo for now, we could just reuse the script.  We can revisit this if we split the repos.

myia_backend_pytorch/scripts/update_env.py
notoraptor: OK, I make further checks, and it appears that, if name is included, conda will create an environment with specified name instead of installing packages in current activated environment. Thus, if specified name is not current environment name, then current environment won't get updated.

So, I must keep removing the name field!

```
(myia) notoraptor@notoraptor-linux:~/mila/dev/git/myia/myia_backend_relay$ conda info --envs
# conda environments:
#
base                     /home/notoraptor/anaconda3
docker                   /home/notoraptor/anaconda3/envs/docker
firebase                 /home/notoraptor/anaconda3/envs/firebase
graphcore                /home/notoraptor/anaconda3/envs/graphcore
myia                  *  /home/notoraptor/anaconda3/envs/myia

(myia) notoraptor@notoraptor-linux:~/mila/dev/git/myia/myia_backend_relay$ cat environment.yml 
name: myia_backend_relay
dependencies:
- python>=3.7,<4.0
- numpy>=1.16,<2.0
- scipy>=1.5,<2.0
- decorator>=4,<5
- attrs>=19,<20
- pip
- mila-iqia::tvm=0.7dev1+5.*
- pip:
  - antlr4-python3-runtime==4.7.2
(myia) notoraptor@notoraptor-linux:~/mila/dev/git/myia/myia_backend_relay$ conda env update --file environment.yml
Collecting package metadata (repodata.json): done
Solving environment: done
Preparing transaction: done
Verifying transaction: done
Executing transaction: done
Installing pip dependencies: / Ran pip subprocess with arguments:
['/home/notoraptor/anaconda3/envs/myia_backend_relay/bin/python', '-m', 'pip', 'install', '-U', '-r', '/media/win/Users/notoraptor/mila/dev/git/myia/myia_backend_relay/condaenv.jtiyxq2z.requirements.txt']
Pip subprocess output:
Processing /home/notoraptor/.cache/pip/wheels/63/93/52/480c17789fbf685f5d0e33e83c311f4a844cd6e2f5cf036c03/antlr4_python3_runtime-4.7.2-py3-none-any.whl
Installing collected packages: antlr4-python3-runtime
Successfully installed antlr4-python3-runtime-4.7.2

done
#
# To activate this environment, use
#
#     $ conda activate myia_backend_relay
#
# To deactivate an active environment, use
#
#     $ conda deactivate

(myia) notoraptor@notoraptor-linux:~/mila/dev/git/myia/myia_backend_relay$ conda info --envs
# conda environments:
#
base                     /home/notoraptor/anaconda3
docker                   /home/notoraptor/anaconda3/envs/docker
firebase                 /home/notoraptor/anaconda3/envs/firebase
graphcore                /home/notoraptor/anaconda3/envs/graphcore
myia                  *  /home/notoraptor/anaconda3/envs/myia
myia_backend_relay       /home/notoraptor/anaconda3/envs/myia_backend_relay
```

myia/compile/backends/__init__.py
notoraptor: Ok ! I removed it anyways !

scripts/update_env.py
notoraptor: OK ! I already moved the script into a package `myia_utils` in previous commit. Should I just move it back to a simple script in myia `scripts` folder ?

notoraptor:
@abergeron updated ! I removed putorch GPU/CPU dependencies (cudatoolkit, cudnn, cpuonly) from myia package.

myia_backend_pytorch/scripts/update_env.py
abergeron: Ok

PR #370, opened 2020-08-18
Keep type annotations in graph, if available.

notoraptor:
@breuleux @abergeron This is a PR to parse annotations (for variables, function arguments, function return) and keep it in new class member `ANFNode.annotation`.

What do you think?


myia/parser.py
```diff
@@ -769,6 +783,14 @@ def process_Assign(self, block: "Block", node: ast.Assign) -> "Block":
 
         return block
 
+    def process_AnnAssign(self, block: "Block", node: ast.AnnAssign) -> "Block":
+        """Process an annotated assignment."""
+        anf_node = self.process_node(block, node.value)
+        # Get type annotation for variable.
+        anf_node.annotation = astunparse.unparse(node.annotation).strip()
```
breuleux, line 790: I think it would be preferable to evaluate the annotations (in the function's environment), rather than store them as strings. Basically something like `eval(annotation, self.function.__globals__)`. I think you can do that from the AST so it might not be necessary to unparse. Also, there should be an informative error pointing to the annotation if the eval fails.

myia/parser.py
```diff
@@ -129,6 +130,15 @@ def _fresh(node):
         return node
 
 
+def _eval_ast_node(node):
+    """Evaluate an AST node as a Python expression and return result."""
+    # eval() does not accept an AST node, so we first convert AST node
+    # to string code.
+    code = astunparse.unparse(node).strip()
+    # Then we parse it with eval().
+    return eval(code)
```
breuleux, line 139: We want to evaluate in the function's environment, though. Let's say we have this code, for example:

```python
from numpy import ndarray

def f(x: ndarray):
    return x * x
```

This is not going to work with just `eval(code)` because the symbol `ndarray` is not defined in parser.py. That's why we want to pass the function's globals (I think they would be in  `f.__globals__`) as the second argument to `eval`.

tests/test_parser.py
```diff
@@ -181,6 +182,40 @@ def j(**kwargs):
...
+        name = node.debug.debug_name
+        if name == "c":
+            assert node.annotation is tuple
+            variables_checked += 1
+        elif name == "d":
+            assert node.annotation is int
+            variables_checked += 1
+    assert variables_checked == 2
+
+
```
breuleux, line 218: We also need a test to check that the annotations are evaluated in the right scope, like the example I gave in the other comment.

notoraptor:
@breuleux code updated!

NB: I notice that this function `f` will be parsed (with imports at top of file):
```
from numpy import ndarray

def f(x: ndarray):
    return x * x
```

But parsing this function `f` will generate an error from eval because it can not find numpy (local import):
```
def wrapper_function():
  from numpy import ndarray

  def f(x: ndarray):
      return x * x
```
It's as if `f.__globals__` won't include local numpy import from container function `wrapper_function`.

breuleux:
@notoraptor That's true, it doesn't include closure variables. It's possible to get closure variables, though. Search for `ClosureNamespace` in the code, that class builds an internal dictionary of closure variables. `eval` takes a second argument for the locals, so maybe it could be made to work this way.

abergeron:
According to PEP563, the postponed evaluation explicitly doesn't support the `wrapper_function()` case.  The python compiler doesn't always keep enough information around to resolve that case so I don't think we should attempt that.

Also postponed-by-default is not until python 3.10, but can be enabled with a `__future__` import.

breuleux:
@abergeron I see. Well, in that case, I suppose we don't need to support it.

notoraptor:
@breuleux @abergeron using `ClosureNamespace` seems to fix the parsing. I added a test to check it.

myia/parser.py
```diff
@@ -255,6 +256,14 @@ def __init__(
         # Variable names that could be phi nodes
         self.possible_phis = set()
 
+    def _eval_ast_node(self, node):
+        """Evaluate an AST node as a Python expression in function context."""
+        # eval() does not accept an AST node, so we first convert AST node
+        # to string code.
+        code = astunparse.unparse(node).strip()
+        # Then we parse it with eval().
```
breuleux, line 264:
```suggestion
        # Then we evaluate it with eval().
```

myia/parser.py
```diff
@@ -402,6 +411,11 @@ def _finalize_function(self, node, function_block):
             with DebugInherit(ast=arg, location=self.make_location(arg)):
                 param_node = Parameter(function_block.graph)
             param_node.debug.name = arg.arg
+
+            if arg.annotation:
+                # Get tyoe annotation for function arg.
```
breuleux, line 416:
```suggestion
                # Get type annotation for function arg.
```

scripts/gen_conda_env_file.sh
```diff
@@ -1,9 +1,13 @@
...
+# NB: package myia_utils must be installed.
 
 # Install python dependencies required to run this script.
 pip install poetry2conda==0.3.0
 
 # Generate environment.yml from pyproject using poetry2conda
 poetry2conda pyproject.toml --dev -E pytorch environment.yml
+
+# Remove name field from environment file.
+python -m myia_utils.update_env environment.yml -o environment.yml
```
breuleux, line 13: Why?

myia/parser.py
notoraptor: OK !

myia/parser.py
notoraptor: Ok !

scripts/gen_conda_env_file.sh
notoraptor: If name is in env yml file, then conda will try to create an environment with this name and install packages in it. It was also discussed here: https://github.com/mila-iqia/myia/pull/369#discussion_r470190420

So, to avoid any future issues, I must remove name field from any environment yml file. I did it for other packages (myia_backend_relay and myia_backend_pytorch) in previous PR, but I forgot to do that for main myia package here!

PR #371, opened 2020-08-24
Move pytorch frontend to a plugin.

notoraptor:
Hi! This is a PR to move pytorch frontend to a plugin module. Now I want to check how tests will run.

@abergeron @breuleux

### DIFF
--- a/myia_frontend_pytorch/pyproject.toml
+++ b/myia_frontend_pytorch/pyproject.toml
@@ -14,6 +14,9 @@ numpy = "^1.16"
 torch = "1.5.0"
 torchvision = "0.6.0"
 
+[tool.poetry.dev-dependencies]
+isort = "~4.3.21"
+
 [tool.poetry2conda]
 name = "myia_frontend_pytorch"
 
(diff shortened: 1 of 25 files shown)


myia_frontend_pytorch/pyproject.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[tool.poetry2conda]
+name = "myia_frontend_pytorch"
+
+[tool.poetry2conda.dependencies]
+torch = { channel = "pytorch", name = "pytorch" }
+torchvision = { channel = "pytorch" }
+
+[tool.isort]
+known_first_party = "myia_backend_pytorch"
```
breuleux, line 25:
```suggestion
known_first_party = "myia_frontend_pytorch"
```

Also, I don't think this section should be here unless `isort` is in the dev requirements for this package.

Thinking about it, the same comment applies to the `pyproject.toml` of the backends.

IMO, while these packages are in subdirectories of the main one, the `isort` section should only be specified for the main `myia` package.

PR #374, opened 2020-09-10
Check that node annotations are rexpected (when available).

notoraptor:
@breuleux @abergeron This is a PR to check node annotations in graph. What do you think ?


notoraptor:
@breuleux Updated ! I successfully moved annotation checking into step_infer. However, I still us a specific annotation checking function, instead of `amerge`, to avoid many problems. In particular, `amerge` either returns an unspecific type (if forced is false), or will complain about type mismatch (if forced is true), while I need to, either return the more specific type (e.g. `tuple` vs `tuple[int, int]`), or check if types match perfectly (e.g `tuple[int, int]` vs `tuple[int, float]`).

breuleux (commented):
Yeah, this is a harder problem than I thought.

Thinking about it some more, I think the issue essentially is that the annotation is _not_ supposed to be merged like normal, because it is a hint and not the actual type of some value in the program: if the application of a function has type `AbstractTuple(ANYTHING)` that means the function can literally return an arbitrary tuple, but the same type as an annotation doesn't mean that, it just means we expect the variable to have a _compatible_ type.

On the other hand, an annotation like `i: int = f(x)` should disambiguate the type of `f(x)` so that if it could be either `float` or `int` at that moment, we now know it is `int`, so the check can't be 100% passive.

If you look at the definition of `typecheck` in `amerge.py` it does `amerge(model, abstract, forced=True, bind_pending=False)`. `model` here would be the annotation, so I'd propose trying `amerge(annotation, abstract, forced=True, bind_pending=True)` (so, the same thing, but we do execute bind_pending). If that doesn't work, I think we might have to stick to `typecheck(annotation, abstract)`.

notoraptor:
@breuleux Hi ! I updated the code with your suggestions (using `amerge(annotation, abstract, forced=True, bind_pending=True)`). It fixed many issues, but I still had to add an overload to `amerge` to manage a specific case, when an annotation `list` (without arguments) is compared to a list (`AbstractADT`).

In such cases, `type_to_abstract(list)` returns an `AbstractUnion`, so `AbstractUnion` vs `AbstractADT` will fail, so I added a `amerge(self, x1: AbstractUnion, x2: AbstractADT)` to handle that.

What do you think?

breuleux:
@notoraptor Good observation, I hadn't thought about that. This is a more general issue, I think. For instance, if the user annotates a variable `x: Union[int, ndarray]`, should it be compatible with an inferred type of `int` or `ndarray`? I suppose that might correspond to user expectation in many cases, but at the same time, are we supposed to allow `x.dot(y)` if we can prove that `x` is an ndarray, even though the user explicitly declared that they think `x` could possibly be an `int`?

It's debatable.

Regarding amerge: it's not clear to me whether merging `U[Empty, Cons]` with `Cons` should yield `U[Empty, Cons]` or raise a type error (currently it does the latter, as you've experienced). However, what it should *not* do is yield `Cons`, because remember that we're also using this to merge branches of conditionals, so the result has to be the union of all possibilities. This is a case where merging the annotation might differ from the normal merging behavior.

Here's something we could try, which is a lot easier now that `ovld` supports multiple dispatch:

```python
def amerge(self, x: AbstractUnion, y: AbstractValue):
    return self(x, AbstractUnion([y]))

def amerge(self, x: AbstractValue, y: AbstractUnion):
    # Note: should fail if forced=True
    return self(AbstractUnion([x]), y)
```

Basically, we'd always get a Union out. I'm curious how that plays out and if all the tests would still pass. I'm a bit hesitant about it, though.

An alternative, if we want to go the other way, would be to do this:

```python
@amerge.variant
def annotation_merge(self, x: AbstractUnion, y: AbstractValue):
    # Reproduce the code you wrote for AbstractUnion and AbstractADT
```

`annotation_merge` would behave just like `amerge` except in this case, and it would not interfere with the existing merge behavior that's used for branches of `if` statements.

I would argue for the second solution, for the time being.

notoraptor:
@breuleux hi ! New commit, I used the `annotation_merge` suggestion.

myia/abstract/amerge.py
```diff
@@ -451,6 +459,33 @@ def amerge(self, x1: object, x2: object, forced, bp):
...
+    an union (union of empty and list of anything), and then
+    amerge(annotation: AbstractUnion, abstract: AbstractADT) will fail.
+
+    This variant is intended to check such cases.
+    """
+    # We check if at least one option in x1 matches x2.
+    merged = None
+    for option in set(x1.options):
+        try:
+            merged = amerge(option, x2, forced, bp)
```
breuleux, line 477:
```suggestion
            merged = self(option, x2, forced, bp)
            break
```

myia/abstract/infer.py
```diff
@@ -60,6 +64,60 @@
...
+def validate_annotation(annotation, abstract):
+    """Check if abstract is allowed by given annotation."""
+    try:
+        dict_annotation = _annotation_as_dict(annotation)
+        if dict_annotation is not None:
+            # type_to_abstract could not generate a suitable
+            # AbstractDict from a type hint because it won't
+            # provide dict keys and values, so we need
+            # a specific check for dict types.
+            _check_dict_annotation(abstract, dict_annotation)
```
breuleux, line 108: Could `type_to_abstract` simply be extended to fix this issue?

notoraptor: Hi @breuleux ! I implemented `pytype_to_abstract` to support `dict` type hints. Annotation tests pass, however, abstract dict need to know keys in some other places in myia code. So, I added a kind of `placeholder` class to set AbstractDict entries when keys are unknown, and I updated `amerge(AbstractDict, AbstractDict)` to support the placeholder. Could you check it ? (last commit)

notoraptor:
@breuleux Updated!

myia/abstract/amerge.py
```diff
@@ -412,8 +415,22 @@ def amerge(self, x1: AbstractClassBase, x2: AbstractClassBase, forced, bp):
...
 @ovld  # noqa: F811
 def amerge(self, x1: AbstractDict, x2: AbstractDict, forced, bp):
-    args1 = (x1.entries, x1.values)
-    args2 = (x2.entries, x2.values)
+    entries_1 = x1.entries
+    entries_2 = x2.entries
+    if isinstance(entries_1, DictDesc) and isinstance(entries_2, DictDesc):
+        # No available keys, use empty dicts.
+        entries_1 = {}
+        entries_2 = {}
```
breuleux, line 423: I think this new definition could be moved to `annotation_merge`, since ultimately `DictDesc` should only be found in annotations (I think?). Furthermore, since the annotation is always going to be the first argument, I think you can simply remove the non-covered branches.

notoraptor:
@breuleux Done ! PR updated.

notoraptor:
Coverage seems OK, I don't know why Travis is failing.

breuleux:
The Travis error looks like it's just a matter of retrying:

```
CondaHTTPError: HTTP 504 GATEWAY_TIMEOUT for url <https://conda.anaconda.org/mila-iqia/linux-64/tvm-0.7dev1 5.f11abf26-py37ha8d69ae_0.tar.bz2>
Elapsed: 01:00.013945
CF-RAY: 5d64a4b7cd47cf5c-IAD
An HTTP error occurred when trying to retrieve this URL.
HTTP errors are often intermittent, and a simple retry will get you on your way.
```

breuleux (approved):

PR #381, opened 2020-11-09
Rewrite convert

breuleux:
This rewrites the ConversionResource and infer_constant in the inferrer to be simpler. The tests seem to take 10-20% less time to execute overall with these changes. For example, for the RNN, `step_infer` is about 30% faster because it doesn't reroute to new constants when the constants already contain converted value (also, it doesn't set a `_converted` field in nodes anymore, which had always been a dirty hack).

I also fixed the debugger when using buche.

### DIFF
--- a/debug/bupdb.py
+++ b/debug/bupdb.py
@@ -1,0 +1,26 @@
+import os
+import pdb
+from buche import BucheDb
+
+
+global_interactor = None
+
+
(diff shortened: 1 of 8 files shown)


abergeron (approved):
Travis is unhappy, but I am.

breuleux:
Everything passes now.

PR #382, opened 2020-11-19
Fix for intern WeakValueDirectory leak

breuleux:
This fixes a leak in the WeakValueDirectory used in intern.py: the keys in that directory are Wrapper instances that themselves contain a weakref to the value which is used in eq/hash computations. This seems to cause an interference that prevents collection of dead values. I have written a new class, `CanonStore`, that fixes this issue.

It doesn't seem to change the test suite performance much if at all. I tried to find ways to reduce the overhead of intern besides that, but nothing really worked, so that's that.

Oh, I also added a `--bw` option for debugging with breakword more easily.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -52,12 +52,21 @@ def pytest_addoption(parser):
         default=False,
         help="Import Myia debug functions",
     )
+    parser.addoption(
+        "--bw",
+        action="store",
+        dest="breakword",
+        default=None,
+        help="Use the given breakword",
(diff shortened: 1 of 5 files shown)


abergeron (approved):

abergeron:
We still need to make sure jenkins runs on this.

breuleux:
@abergeron Jenkins and the rest of the tests succeeded.

PR #383, opened 2020-11-24
Allow tests to explicitly skip backend for unsupported op categories

notoraptor:
Hi @breuleux @abergeron this is a PR to allow to skip tests for backends that don't support some specified op categories.

- I added a file to define op categories (named `computations.py`) into `myia.compile.backends`
- I updated Backend base class and derived class in each backend with a method to check if an op category is supported
- I added a testing decorator `bt` (for `backend testing`), which can receives a list of required op categories and will explicitly skip test for if backend does not support all required categories.
- I used `bt` in `test_universal.py` as example.

What do you think ?


notoraptor:
@abergeron @breuleux PR updated ! Though Jenkins does not start ...

myia/compile/backends/prim_groups.py
```diff
@@ -0,0 +1,144 @@
...
+        P.scalar_tanh,
+        P.scalar_trunc,
+        P.scalar_uadd,
+        P.scalar_usub,
+    ],
+)
+
+
+# Ignored primitives:
+# P.make_handle
```
abergeron, line 137: This should be in the universe group.

myia/api.py
```diff
@@ -48,11 +48,6 @@ def __init__(
         pipeline=standard_pipeline,
     ):
         """Initialize a MyiaFunction."""
-        # Change this once relay becomes the default backend.
-        if use_universe and backend not in ("python", "relay"):
-            raise RuntimeError(  # pragma: no cover
-                "Universe is only supported for the relay backend."
-            )
```
abergeron, line 55: This check should probably stay as it is not for the tests but for end users.  Unless I'm missing something in this PR.

abergeron (approved):
I'm overall ok, with two minor nits

abergeron:
We still have to make sure that Jenkins runs.

notoraptor:
@abergeron Updated (see latest commit). Jenkins launched !

notoraptor:
Hi @abergeron tests succeed on Jenkins, but coverage fail: https://jenkins.server.mila.quebec/blue/organizations/jenkins/Mila%2Fmyia/detail/PR-383/3/pipeline

Failing files should be ignored based on `codecov.yml` file, but they appear on report: https://github.com/mila-iqia/myia/blob/master/codecov.yml

notoraptor:
Oh, ultimately, all tests passed (even coverage) !

myia_backend_relay/myia_backend_relay/relay.py
```diff
@@ -1071,6 +1071,13 @@ def compile(self, graph, argspec, outspec):
             graph, self.context, self.target, self.exec_kind
         )
 
+    def supports_prim_group(self, prim_group):
+        if prim_group is G.universe_operations:
+            # Relay does not explicitly implement primitive make_handle
+            # but still supports universe operations.
+            return True
```
breuleux, line 1078: That's a brittle check, we should accept any primitive group that contains `make_handle`. A better hardcoded fix would be to write `MAP.get(prim) or prim is P.make_handle` as the condition in the line below. Alternatively, set `P.make_handle: True` in `SIMPLE_MAP`, or have an additional set of supported primitives for Relay.

notoraptor: Updated @breuleux with `MAP.get(prim) or prim is P.make_handle` !

PR #385, opened 2020-11-30
Add Python backend option `pdb: bool = False`

notoraptor:
Hi @abergeron @breuleux ! This is a PR that adds a Python boolean backend option `pdb`. If set to True, then code will be run with PDB. Tested successfully, e.g. with `pytest -xvvs tests/frontends/test_all_ops.py::test_conv2d_no_dil[python-cpu-grad0]` (pdb will run, then I need to type `continue` to fully run the code, then `quit` to close PDB, and test finishes correctly). However, I don't know how to test it through travis/jenkins.


notoraptor:
I added a test for pdb.

myia_backend_python/myia_backend_python/python.py
```diff
@@ -462,6 +462,82 @@ def nested_list_to_code_string(structure, indentation=""):
...
+        final_code = self.code + pdb_code
+        try:
+            # Save script.
+            with open(code_path, "w") as code_file:
+                code_file.write(final_code)
+            # Save main args.
+            with open(args_path, "wb") as input_file:
+                pickle.dump(args, input_file)
+            # Run script.
+            subprocess.run(["python", "-m", "pdb", code_path])
```
breuleux, line 525: This will start debugging at the entry of the module you are generating, which isn't super useful: it will start at `import math` and you'll have to do next a bunch of times and then step inside `main`. Ideally we want to start debugging inside `main` directly.

I think a better way would be something like this:

```python
# Create a globals dictionary
glb = {}

# Execute the code for the function using glb as the globals
exec(self.code, glb)

# Run pdb at the call, assuming the main function is called "main"
pdb.runcall(glb["main"], *args)
```

I'm not sure how to write a test for that, but it's somewhat less important to test debug stuff, since you're not going to run it in production.

breuleux (changes requested):

notoraptor:
Hi @breuleux updated, using `runcall`, it seems to work:
```
(myiagpu) notoraptor@notoraptor-linux:~/mila/dev/git/myia$ pytest -xvvs myia_backend_python/tests/test_python_pdb.py 
=========================================================================================== test session starts ===========================================================================================
platform linux -- Python 3.7.9, pytest-6.1.2, py-1.9.0, pluggy-0.13.1 -- /home/notoraptor/anaconda3/envs/myiagpu/bin/python
cachedir: .pytest_cache
rootdir: /media/win/Users/notoraptor/mila/dev/git/myia, configfile: pytest.ini
plugins: cov-2.10.1, xdist-2.1.0, forked-1.3.0
collected 1 item                                                                                                                                                                                          

myia_backend_python/tests/test_python_pdb.py::test_pdb 
import math
import numpy as np
from myia.utils import RandomStateWrapper
from myia.lib import TaggedValue
from myia.utils.universe import HandleInstance
import myia_backend_python.implementations as IMPL
def main(_a, _b, _c, _d):
    _scalar_pow = _a ** _b
    _scalar_div = type(_scalar_pow)(_scalar_pow / _d)
    _scalar_mul = _scalar_pow * _c
    _scalar_add = _scalar_mul + _scalar_div
    return _scalar_add + _scalar_pow

  4  	from myia.lib import TaggedValue
  5  	from myia.utils.universe import HandleInstance
  6  	import myia_backend_python.implementations as IMPL
  7  	def main(_a, _b, _c, _d):
  8  	    _scalar_pow = _a ** _b
  9  ->	    _scalar_div = type(_scalar_pow)(_scalar_pow / _d)
 10  	    _scalar_mul = _scalar_pow * _c
 11  	    _scalar_add = _scalar_mul + _scalar_div
 12  	    return _scalar_add + _scalar_pow
 13  	
[EOF]
PASSED

============================================================================================ warnings summary =============================================================================================
myia_backend_python/tests/test_python_pdb.py::test_pdb
myia_backend_python/tests/test_python_pdb.py::test_pdb
  /home/notoraptor/anaconda3/envs/myiagpu/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: numpy.ufunc size changed, may indicate binary incompatibility. Expected 192 from C header, got 216 from PyObject
    return f(*args, **kwds)

-- Docs: https://docs.pytest.org/en/stable/warnings.html
====================================================================================== 1 passed, 2 warnings in 0.86s ======================================================================================
```

notoraptor:
Well, I don't know why, but coverage seems broken, with many uncovered lines even in non-backend and non-tests files.

notoraptor:
Aaah, ultimately, tests passed.

It seems the problem was because PDB is imported during testing. To prevent PDB import in main pytest run, I make sure function compiled with `pdb=True` is executed in a separate process, by using `multiprocessing`. It helped make coverage work again, but there were still some un-covered lines: the ones just after the call to `pdb.runcall`: https://github.com/notoraptor/myia/blob/pdb-like/myia_backend_python/myia_backend_python/python.py#L500 , as if PDB call "broke" something in coverage report. I simply decide to make those lines with `# pragma: no cover`.

So, it seems PDB (either import or run) is not compatible with `pytest-cov`.

@abergeron @breuleux

breuleux (approved):

PR #386, opened 2020-12-18
Improve convert_type for constants in all backends.

notoraptor:
Hi @breuleux this is a small PR to improve `convert_type` for constants in backends.

Initially, I should have made it only for Python backend, but I was unable to write a good test, because either raw types are deleted from graph if not used, or generated compilation error because they are not handled by all backends.

So, I try to improve type support (at least as constant) in all backend, and I add a simple test that return a type.

### DIFF
--- a/myia/simplify_types.py
+++ b/myia/simplify_types.py
@@ -491,6 +491,11 @@ def from_canonical(self, arg, orig_t: AbstractUnion):
         raise AssertionError(f"Badly formed TaggedValue")
 
 
+@ovld  # noqa: F811
+def from_canonical(self, arg, orig_t: AbstractType):
+    return arg
+
+
 __consolidate__ = True
 __all__ = [
     "from_canonical",
--- a/myia/simplify_types.py
+++ b/myia/simplify_types.py
@@ -493,7 +493,10 @@ def from_canonical(self, arg, orig_t: AbstractUnion):
 
 @ovld  # noqa: F811
 def from_canonical(self, arg, orig_t: AbstractType):
-    return arg
+    if isinstance(orig_t.element, AbstractHandle):
+        return HandleInstance
+    else:
+        return arg
 
 
 __consolidate__ = True
--- a/myia/simplify_types.py
+++ b/myia/simplify_types.py
@@ -493,10 +493,7 @@ def from_canonical(self, arg, orig_t: AbstractUnion):
 
 @ovld  # noqa: F811
 def from_canonical(self, arg, orig_t: AbstractType):
-    if isinstance(orig_t.element, AbstractHandle):
-        return HandleInstance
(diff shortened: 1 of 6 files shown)


tests/test_api.py
```diff
@@ -437,3 +437,17 @@ def f(x, y):
 
     assert _flag1[0] is True
     assert _flag2[0] is True
+
+
+@bt()
+def test_convert_type(backend):
+    @myia(backend=backend)
+    def f(v):
+        return tuple
```
breuleux, line 446: Did you mean not to use the argument?

myia_backend_python/myia_backend_python/python.py
```diff
@@ -441,9 +441,14 @@ def convert_tuple(self, v: tuple, t: AbstractTuple):
...
-        myia_type = t.element.xtype()
         # Return type name as a string.
-        return f"'{type_to_np_dtype(myia_type)}'"
+        myia_type = t.element.xtype()
+        if myia_type is None:
+            if isinstance(v, AbstractHandle):
+                return "HandleInstance"
+        if myia_type is Tuple:
+            return "tuple"
+        return f"np.{type_to_np_dtype(myia_type)}"
```
breuleux, line 451: I think the code is a bit oddly structured here.

```suggestion
        if isinstance(v, AbstractHandle):
            return "HandleInstance"
        else:
            myia_type = t.element.xtype()
            if myia_type is Tuple:
                return "tuple"
            else:
                return f"np.{type_to_np_dtype(myia_type)}"
```

notoraptor:
@breuleux Updated !

breuleux (approved):
Looks good. Jenkins queued the jobs because the GPUs are not available apparently.

notoraptor:
@abergeron @breuleux tests passed. Previous test (yesterday) failed because of coverage, as conversion of HandleInstance type was not tested, so, I added a supplementary commit to test it and make sure HandleInstance is well managed. A new review might be required!

myia/simplify_types.py
```diff
@@ -493,7 +493,10 @@ def from_canonical(self, arg, orig_t: AbstractUnion):
 
 @ovld  # noqa: F811
 def from_canonical(self, arg, orig_t: AbstractType):
-    return arg
+    if isinstance(orig_t.element, AbstractHandle):
+        return HandleInstance
+    else:
+        return arg
```
breuleux, line 499: Why is it necessary to have a special case here?

notoraptor: Indeed, useless. I just needed to fix HandleInstance conversion in Pytorch backend, instead.

notoraptor:
@breuleux updated

breuleux (approved):

PR #389, opened 2021-02-09
Improvement to print_graph, add print_node and better inference errors.

abergeron:


breuleux (approved):

abergeron:
Now with more coverage!

PR #390, opened 2021-02-16
(WIP) Try to fix f'' when a switch is used.

notoraptor:
Hi! This is just my current code trying to make work second order gradient on a function using an `if`. Well, after 2 weeks, I am still unable to figure out why it fails, so I'm currently blocked.

One of my current issues is that, now, graph contains some `J(apply)` nodes, ie. not only `J(constant)` nodes. I'm not sure if it's normal, and how to handle it.

This is my current output:

```
$ pytest -xvvs tests/test_grad.py::test_second_order_on_if[python] > myia_debug/out.log
```

[Output file `out.log`](https://gist.github.com/notoraptor/66f9322c9738b9336b155e5137764c2a)

@abergeron @breuleux

### DIFF
--- a/debug/j_utils.py
+++ b/debug/j_utils.py
@@ -1,0 +1,107 @@
+from myia.ir import Graph
+from myia.operations import Primitive, primitives as P
+
+
+class NodeDescriptor:
+    """Node descriptor class, used to keep inputs and output for a node."""
+
+    PARAM = 0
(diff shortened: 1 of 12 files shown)


notoraptor:
I close this PR for the moment. Gradient will be implemented later in new myia `zero` branch.

PR #392, opened 2021-02-23
[WIP] Current IR proposition

abergeron:


myia/ir/node.py
```diff
@@ -0,0 +1,202 @@
...
+        self.varargs = False
+        self.kwargs = False
+        self.defaults = {}
+        self.kwonly = 0
+        self.location = None
+
+    @property
+    def output(self):
+        if not self.return_ or 0 not in self.return_.edges:
+            raise ValueError("Graph has no input")
```
breuleux:
```suggestion
            raise ValueError("Graph has no output")
```

myia/ir/node.py
```diff
@@ -0,0 +1,202 @@
...
+        return self.return_.edges[0].node
+
+    @output.setter
+    def output(self, value):
+        if self.return_:
+            self.return_.edges[0].node = value
+        else:
+            self.return_ = self.apply("return_", value)
```
breuleux, line 42: That probably shouldn't be a string.

myia/ir/node.py
```diff
@@ -0,0 +1,202 @@
...
+                    edge.node = mapping_seq[edge.node]
+                elif edge.node in mapping:
+                    edge.node = mapping[edge.node]
+                if edge.node and edge.node.is_apply():
+                    todo.append(edge.node)
+
+    def __str__(self):
+        if self.name is not None:
+            return self.name
+        return "<unamed_graph>"
```
breuleux:
```suggestion
        return "<unnamed_graph>"
```

myia/ir/node.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def is_constant(self, cls=object):
+        return False
+
+    def is_constant_graph(self):
+        return False
+
+    def ensure_info(self):
+        if self.info is None:
+            self.info = SimpleNamespace()
```
breuleux: Why not create `self.info` in the constructor?

myia/ir/node.py
```diff
@@ -0,0 +1,202 @@
...
+    def add_edge(self, label, node):
+        assert label not in self.edges
+        e = Edge(label, node)
+        self.edges[label] = e
+
+    def append_input(self, node):
+        i = 0
+        while i in self.edges:
+            i += 1
+        self.add_edge(i, node)
```
breuleux: Is this used anywhere?

myia/ir/node.py
```diff
@@ -0,0 +1,202 @@
...
+    def inputs(self):
+        i = 0
+        res = []
+        while i in self.edges:
+            res.append(self.edges[i].node)
+            i += 1
+        return res
+
+    def __str__(self):
+        return f"{self.fn}({self.inputs})"
```
breuleux: That representation risks being a bit misleading, because it may not show all edges. If e.g. you have inputs 0, 1 and 3 but not 2, it'll only show 0 and 1. If some edges correspond to keyword arguments, they won't show up either. More importantly it will crash if edge FN is not present, and it can be annoying to debug a problem with a node you can't print.

tests/test_parser.py
```diff
@@ -0,0 +1,1123 @@
...
+}
+"""
+
+
+def test_boolop():
+    def f(a, b, c):  # pragma: nocover
+        return a and b or c
+
+    assert str_graph(parse(f)) == """graph f(%a, %b, %c) {
+  %_apply0 = universe_setitem(%_apply1, %b)
```
breuleux: What's `%_apply1`? Seems to be `make_handle` but it should appear here.

myia/ir/node.py
abergeron: Yep, but we don't have operations.

myia/ir/node.py
abergeron: This is a possibly useless space optimization.

tests/test_parser.py
abergeron: This was fixed

myia/ir/node.py
```diff
@@ -0,0 +1,183 @@
...
+            if node in seen:
+                continue
+            seen.add(node)
+            for edge in node.edges.values():
+                if edge.label is SEQ and edge.node in mapping_seq:
+                    edge.node = mapping_seq[edge.node]
+                elif edge.node in mapping:
+                    edge.node = mapping[edge.node]
+                if edge.node and edge.node.is_apply():
+                    todo.append(edge.node)
```
breuleux: I'd like some documentation about what this function does and why it exists.

It might also be a good idea to add an `iter_edges` method that iterates over edges and use it to implement `replace`. Pretty sure it'd be useful in other situations.

myia/ir/node.py
```diff
@@ -0,0 +1,183 @@
...
+        return False
+
+    def is_constant(self, cls=object):
+        return False
+
+    def is_constant_graph(self):
+        return False
+
+    def add_annotation(self, annotation):
+        self.annotation = annotation
```
breuleux: Might as well remove this method and write `node.annotation = x` directly.

myia/utils/info.py
```diff
@@ -54,6 +70,12 @@ class DebugInfo:
 
     """
 
+    def __new__(cls, *args, **kwargs):
+        if not _debug.get():
+            return None
+        else:
+            return super().__new__(cls)
+
```
breuleux: It should be possible to instantiate `DebugInfo` without setting a contextvar.

This should be moved to a `maybe_debug` function (or some other name, I don't mind) that `Node`, `Graph`, etc. can call instead of the class constructor.

myia/parser.py
```diff
@@ -0,0 +1,958 @@
...
+from .utils.info import debug_inherit, get_debug
+
+
+class Location(NamedTuple):
+    """A location in source code.
+    Attributes:
+        filename: The filename.
+        line: The line number.
+        column: The column number.
+    """
```
breuleux: Missing line_end and column_end

myia/parser.py
```diff
@@ -0,0 +1,958 @@
...
+            function_block.graph.parameters.append(param_node)
+            function_block.write(arg.arg, param_node)
+            if dflt:
+                # TODO If there are no parents (the initial function),
+                # then evaluate in the global env
+                if block is None:
+                    raise MyiaSyntaxError(
+                        "default value on the entry function",
+                        self.make_location(arg),
+                    )
```
breuleux, line 512: Could we simply use `_eval_ast_node` here?

myia/parser.py
```diff
@@ -0,0 +1,958 @@
...
+
+        after_block = self.process_statements(function_block, node.body)
+
+        if after_block.graph.return_ is None:
+            after_block.returns(Constant(None))
+
+        if node.returns:
+            function_block.graph.return_.add_annotation(
+                self._eval_ast_node(node.returns)
+            )
```
breuleux: Do we want to annotate `return_` or `output`? The latter might be more accurate if we decide to add more return values to `return_` like the universe etc. but I don't know.

myia/parser.py
```diff
@@ -0,0 +1,958 @@
...
+
+        return function_block
+
+    def make_condition_blocks(self, block, tn, fn):
+        with debug_inherit(name="if_true", location=self.make_location(tn)):
```
breuleux: I suppose we can leave it for another PR because it'd require some adjustments in the tests, but it'd be better to use `about(block.graph, relation="if_true", location=...)` because it would let us generate labels like `f.if_true` instead of just `if_true`, which isn't as informative.

myia/ir/node.py
abergeron: Yes, I did that.

myia/utils/info.py
abergeron: Ok I will do that.

myia/parser.py
abergeron: Maybe, but I don't think it would work because that might need a bunch of context that we don't have.

I would prefer to leave it like that for the merge, but we can revisit after.

myia/parser.py
abergeron: It is technically more accurate here since the annotation is on the function, not the output expression.

Also, the extra return values should be on different edges so we can separate them.

myia/parser.py
abergeron: Yes

myia/parser.py
abergeron: Done

myia/ir/node.py
abergeron: I did add some documentation in the latest commit, tell me if it's good enough.

As for iter_edges, I'm not sure.  I'm not against the idea, but I don't see the advantages over plain for loop.

breuleux: Why are there two mapping dictionaries? That's the main interrogation I have right now.

I meant an iter_edges method on Graph to perform a bfs of all of the graph's edges. I suppose we can refactor when we need to, but I suspect we will need to do it at some point.

myia/parser.py
breuleux: Does "yes" mean you're doing it, or that we'll leave it for another PR?

myia/ir/node.py
abergeron: Because sometimes you want to replace a node with another one, but then sequence chain should flow differently.

So mapping_seq allows to override (or define) replacements for the sequence chain specifically.

myia/parser.py
abergeron: I can do it in this PR since it's not a big change.

myia/ir/node.py
breuleux: Okay, I think I see what you mean. Replacing node 1 by node 2 in the data flow might not mean you want to perform the same replacement in the sequence chain.

abergeron:
The only thing that fails is the coverage in 3.9 because some AST nodes have been removed.

I want to use codecov for coverage, but I would rather do that in another PR.  This one is already enormous and have been in the works for too long.

So we can either merge as-is or remove the coverage check and reintroduce it later with codecov.

PR #393, opened 2021-02-28
fix inconsistency package dependendy of poetry install

strint:


abergeron:
I've started a manual CI run for this PR.  Thanks for the changes!

PR #394, opened 2021-03-02
More python backend

notoraptor:
Improve Python backend compilation for a graph just after parsing.

- Support `if`
- Support while loop
- Support for loop with `range`
- Support recursions
- Support `grad` and `value_and_grad` using finite differences
- Inline `operation.resolve` (use resolved value instead of node name)
- Support importation of global symbols in compiled module (e.g. this allows to import an external function returned by operation.resolve)

I have not yet tested all operations that exist in `myia.operations` module, so I am not sure if everything is already well handled (I guess at least macros won't work, except `user_switch`). Maybe I will  do final full test in a 3rd PR ?

Also, Python backend still generates string code, not AST.

@breuleux 
@abergeron


myia/compile/backends/python/python.py
```diff
@@ -197,6 +211,59 @@ def python_universe_getitem(c, universe, handle):
...
+def op_user_switch(cond, if_true, if_false):
+    """Implementation for operation `user_switch`."""
+    return if_true if cond else if_false
+
+
+def op_range(start, stop=None, step=None):
+    """Implementation for operation range."""
+    from myia.operations.op_range import range_
+
+    return range_(start, stop, step)
```
breuleux, line 254: I think this may be better supported by implementing a `myia_iter/next/hasnext` function that can handle range (and possibly generalizes over Python's `iter/next`).

myia/compile/backends/python/python.py
```diff
@@ -903,6 +1083,9 @@ def run(self, graph, backend):
         # reference: https://stackoverflow.com/a/19850183
         compiled = compile(final_code, "", "exec")
         module = ModuleType("mod")
+        # Add required global variables into module.
+        for name, value in self.globals.items():
+            module.__dict__[name] = value
         exec(compiled, module.__dict__)
         return getattr(module, "main")
```
breuleux, line 1090: You don't need to create a module:

```suggestion
        exec(compiled, self.globals)
        return self.globals["main"]
```

myia/compile/backends/python/python.py
```diff
@@ -297,17 +364,68 @@ def python_universe_getitem(c, universe, handle):
...
+            # Function will be available as global symbol in compiled function.
+            fn = FUNCTION_MAP[resolved]
+            symbol_name = fn.__name__
+            c.register_global(symbol_name, fn)
+            code = symbol_name
+        elif isinstance(resolved, FunctionType):
+            # Resolved is a function.
+            # Function will be available as global symbol in compiled function.
+            symbol_name = resolved.__name__
+            c.register_global(symbol_name, resolved)
```
breuleux, line 400: Two different functions may have the same `__name__`, in which case they should be registered under different global names.

myia/compile/backends/python/python.py
```diff
@@ -297,17 +374,58 @@ def python_universe_getitem(c, universe, handle):
...
+            if impl is not None:
+                code = c.register_global(f"operation_{resolved.name}", impl)
+        elif isinstance(resolved, ModuleType):
+            # Module imported, probably to call an external function
+            # (e.g. `torch.argmax`)
+            code = c.register_global(resolved.__name__, resolved)
+        elif callable(resolved):
+            # Resolved is a callable.
+            # Callable will be available as global symbol in compiled function.
+            code = c.register_global(resolved.__name__, resolved)
```
breuleux, line 414: A callable can be any object that defines `__call__`. They may not have a `__name__`.

tests/python_backend/test_graph_compilation.py
```diff
@@ -1,18 +1,263 @@
...
+            return self.value, HalfIterator(self.value // 2)
+
+    @parse_and_compile
+    def f(v):
+        ret = 0
+        for x in HalfIterator(v):
+            ret = ret + x
+        return ret
+
+    print(f(10))
```
breuleux, line 162: Test if it returns the correct value.

myia/compile/backends/python/python.py
```diff
@@ -511,7 +629,12 @@ def convert_default(self, v):
         # (e.g. for inputs of operations.resolve apply nodes).
         from myia.utils.misc import Namespace
 
-        if isinstance(v, (str, Namespace, Operation)):
+        if isinstance(v, str):
+            # Return a ConstantString so that given string will be inlined.
+            # Given string is wrapped inside 2 double quotes.
+            assert '"' not in v, v
+            return ConstantString(f'"{v}"')
```
breuleux, line 636: You can use `repr(v)` to quote the string properly even if it contains `"`.

Also I am not entirely sure what the purpose of `ConstantString` is, can't you return the string directly?

notoraptor: Hi ! Ok for `repr` !

About `ConstantString`:

As I currently compile by generating code string, each constant node may be converted to a string.

For example, `my_node = (1, 2, 3)`. I will save `(1, 2, 3)` as a string code,

But I can not inline all constant nodes (e.g. constant tuple nodes), so I need a way to make distinction between constant nodes than must be defined (ie. `constant_node_name = constant_node_value`), and constant nodes than can be inlined (ie. use `constant_node_value` everywhere `constant_node_name` appears). As `constant_node_value` may be a string in both cases, I can not just decide to inline all consant nodes converted to strings, so I must separate strings that can be inlined and ones that can not. I use `ConstantString` to do that, by wrapping strings that can be inlined, and then inline any constant node associated to a `ConstantString`.

breuleux (approved):

PR #397, opened 2021-04-30
New python backend for zero branch

notoraptor:
Hi @breuleux @abergeron , this is my PR for new Python backend in `zero` branch.

### DIFF
--- a/myia/compile/backends/python/code_generator.py
+++ b/myia/compile/backends/python/code_generator.py
@@ -1,15 +1,24 @@
+"""Code generator used to generate compiled code."""
+
 import builtins
 import operator
 from collections import Counter
 
 from myia.compile.backends.python.directed_graph import DirectedGraph
-from myia.compile.backends.python.implementations import MakeHandle, myia_hasnext, myia_iter, myia_next, typeof, \
-    Universe
+from myia.compile.backends.python.implementations import (
+    MakeHandle,
+    Universe,
+    myia_hasnext,
+    myia_iter,
+    myia_next,
+    typeof,
+)
 from myia.ir import Node
 from myia.utils.info import Labeler
 
 
-def default_str(c, format_string, nodes):
+def default_formatter(c, format_string, nodes):
+    """Generate code using given format string and input nodes."""
     return format_string.format(*[c.label(node) for node in nodes])
 
 
@@ -29,16 +38,26 @@ def str_make_list(c, *inputs):
 
 
 def str_make_dict(c, *inputs):
+    """Formatter for apply `make_dict`."""
     assert not len(inputs) % 2
     pairs = [c.rvalue(inp) for inp in inputs]
-    return '{' + ', '.join('{}: {}'.format(pairs[2 * i], pairs[2 * i + 1]) for i in range(len(pairs) // 2)) + '}'
+    return (
+        "{"
+        + ", ".join(
+            "{}: {}".format(pairs[2 * i], pairs[2 * i + 1])
+            for i in range(len(pairs) // 2)
+        )
+        + "}"
+    )
 
 
 def str_apply(c, fn, args, kwargs):
+    """Formatter for apply `apply`, which seems to represent a call with args and kwargs."""
     return f"{c.label(fn)}(*{c.rvalue(args)}, **{c.rvalue(kwargs)})"
 
 
 def str_getattr(c, *params):
+    """Formatter for apply getattr."""
     assert len(params) in (2, 3)
     if len(params) == 2:
         obj, symb = params
@@ -88,7 +107,13 @@ COMPLEX_MAP = {
 
 
 class NodeLabeler:
+    """Node labeler used in Python backend.
+
+    Combine a Labeler (from myia.utils.info) and a node cache to generate default names when necessary.
+    """
+
     def __init__(self):
+        """Initialize."""
         self.cache = {}
         self.default_name_counter = Counter()
         self.lbl = Labeler(
@@ -99,7 +124,9 @@ class NodeLabeler:
         )
 
     def __call__(self, node):
+        """Generate label for given node."""
         if isinstance(node, Node) and node.is_constant_graph():
+            # Use labeler for graph in constant node.
             return self.lbl(node.value)
         else:
             # Use labeler for anything else.
@@ -180,7 +207,9 @@ class CodeGenerator:
         fn = node.fn
         if fn.is_constant():
             if fn.value in SIMPLE_MAP:
-                return default_str(self, SIMPLE_MAP[fn.value], node.inputs)
+                return default_formatter(
+                    self, SIMPLE_MAP[fn.value], node.inputs
+                )
             elif fn.value in COMPLEX_MAP:
                 return COMPLEX_MAP[fn.value](self, *node.inputs)
             elif fn.value in self.module_implementations:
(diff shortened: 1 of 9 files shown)


myia/compile/backends/python/code_generator.py
```diff
@@ -0,0 +1,286 @@
...
+            return str(node.value)
+
+        if node.debug:
+            return None
+
+        if node not in self.cache:
+            name = f"_{type(node).__name__.lower()}"
+            self.default_name_counter.update([name])
+            self.cache[node] = f"{name}{self.default_name_counter[name]}"
+        return self.cache[node]
```
breuleux: Why not just return None here and let the labeler take care of the counters?

myia/compile/backends/python/code_generator.py
```diff
@@ -0,0 +1,286 @@
...
+            # Use labeler for anything else.
+            return self.lbl(node)
+
+    @classmethod
+    def _relation_translator(cls, rel):
+        return f"{rel}_"
+
+    @classmethod
+    def _name_generator(cls, identifier):
+        return f"_apply{identifier}"
```
breuleux: I think `_{identifier}` would be good enough, the `apply` is not very informative so it would just look like clutter.

myia/compile/backends/python/code_generator.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+class NodeLabeler:
+    """Node labeler used in Python backend.
+
+    Combine a Labeler (from myia.utils.info) and a node cache to generate default names when necessary.
+    """
+
+    def __init__(self):
+        """Initialize."""
```
breuleux: Documentation on init is not necessary anymore, so you don't need to bother if it isn't informative

```suggestion
```

myia/compile/backends/python/code_generator.py
```diff
@@ -0,0 +1,286 @@
...
+            self.default_name_counter.update([name])
+            self.cache[node] = f"{name}{self.default_name_counter[name]}"
+        return self.cache[node]
+
+
+class CodeGenerator:
+    """Helper class to convert graph and nodes to code string."""
+
+    def __init__(self):
+        """Initialize."""
```
breuleux:
```suggestion
```

myia/compile/backends/python/directed_graph.py
```diff
@@ -0,0 +1,118 @@
+"""Implementation of a directed graph.
+
+Used to convert a myia graph to a directed graph.
+Directed graph is then visited to generate Python code.
+"""
```
breuleux, line 5: Can you explain real quick why DirectedGraph exists? Why do we need another graph class? Can't we sort the nodes in Graph directly?

myia/compile/backends/python/directed_graph.py
```diff
@@ -0,0 +1,118 @@
...
+
+        :return: sequence of graph values in visited order.
+        """
+        assert self.has(None)
+        cp = self._copy()
+        sequence = []
+        todo = cp.pop(None)
+        while todo:
+            element = todo.pop(0)
+            sequence.append(element)
```
breuleux: It would make sense for this function to be a generator.

```suggestion
            yield element
```

myia/compile/backends/python/implementations.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self):
+        """Initialize."""
+        self.counter = Counter()
+
+    def __call__(self, object_type):
+        """Generate an handle for given object type."""
+        name = object_type.__name__
+        self.counter.update([name])
+        count = self.counter[name]
+        return f"{name}_h{count}"
```
breuleux: Seems a bit indirect. Why not just have a Handle class?

```suggestion
_counter = Counter()

class Handle:
    """Implementation of `make_handle`."""

    def __init__(self, object_type):
        """Generate an handle for given object type."""
        name = object_type.__name__
        _counter.update([name])
        count = _counter[name]
        self.name = f"{name}_h{count}"
        
    def __str__(self):
        return self.name
```

Technically you probably don't need the counter since every instance of Handle is unique.

myia/compile/backends/python/implementations.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class Universe:
+    """Helper class for Universe."""
+
+    def __init__(self):
+        """Initialize."""
```
breuleux:
```suggestion
```

myia/compile/backends/python/python.py
```diff
@@ -0,0 +1,227 @@
...
+        """
+        code_generator = CodeGenerator()
+        code = []
+        for directed in GraphToModule().generate_directed_graphs(graph):
+            code.extend(code_generator.directed_graph_to_code(directed))
+
+        module = code_generator.globals
+        dynamic_imports = [
+            f"# Dynamic external import: {name}" for name in module
+        ]
```
breuleux: I think it would be possible to generate static imports as well, something like

```python
[f"from {value.__module__} import {value.__name__} as {name}" for name, value in module.items()]
```

That will only work if `value` has `__module__` and `__name__` and `value.__qualname__` does not contain dots, but I think that should be most of the time. The advantage is that this would allow us to write the code to a file to create a real module. In this case I think I would have methods `generate_code`, `write` and `compile` instead of just `compile`, so that the user can choose what they want to do.

tests/python_backend/test_graph_compilation.py
```diff
@@ -0,0 +1,277 @@
...
+from myia.compile.backends.python.python import compile_graph
+from myia.parser import parse
+from myia.utils.info import enable_debug
+
+
+def parse_and_compile(function):
+    with enable_debug():
+        graph = parse(function)
+    # from myia.ir.print import str_graph
+    # print(str_graph(graph))
```
breuleux:
```suggestion
```

Or add a flag to toggle the printing, if you think that's useful to keep for debugging purposes.

myia/compile/backends/python/code_generator.py
notoraptor: Hi! Indeed, it will work. but it will use pattern `_apply{id}` everywhere (because of `name_generator`), both for function names, apply names, and parameter names. Here I try to generate different names for each kind of node, to keep a bit of information even if debug info is not available. So, functions are named `_graph{id}`, applies are named `_apply{id}`, and parameters are named `_parameter{id}`.

Should I still remove it and let labeler generate default names ?

breuleux: If we need more information, we probably want to use the debug mode anyway, so it'd be better to minimize the amount of code.

myia/compile/backends/python/directed_graph.py
notoraptor: Hi!

Indeed, I could simply sort the graph nodes. But I don't know initially all the graphs involved in the main graph to compile. I can find them only while recursively visiting the main graph. Sometimes, a closure `a` for a graph is only discoverable when visiting another closure `b` of same graph, because only `b` calls `a`. And, for each closure found, I need to correctly position closure code into parent graph code, so that closure is defined after parent variables it uses, and before parent variables/closures that use it.

So, to do that more conveniently, I chosed to use directed graph as strategy. It allows me to link users to used nodes (applies or closures) easily, and then visit nodes in an order when used nodes easily come before user nodes.

Goal was also to have an implementation logic more easy to understand, thus more easy to modify/debug later.

I think I could remove directed graph logic, but I would need a different way to keep trace of links between closures and applies that use them.

myia/compile/backends/python/code_generator.py
notoraptor: Ok! So I will just return None.

myia/compile/backends/python/directed_graph.py
breuleux: Okay, let's keep it that way.

breuleux (approved):

PR #398, opened 2021-05-03
Abstract data types and utilities

breuleux:
In preparation for the inference algorithm, here are the data structures for the type system as well as the functions to manipulate them.

### DIFF
--- a/myia/abstract/data.py
+++ b/myia/abstract/data.py
@@ -1,5 +1,3 @@
-# from ..utils import ABSENT, Atom, AttrEK, Interned, PossiblyRecursive
-from dataclasses import dataclass
 from itertools import count
 
 from hrepr import pstr
@@ -238,7 +236,7 @@ class AbstractUnion(AbstractValue):
                 *hrepr(
                     self.tracks, omit_tracks={"interface"}, bare_tracks=True
                 ).children,
-                type=f"*U",
+                type="*U",
             )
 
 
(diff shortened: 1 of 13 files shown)


myia/abstract/data.py
```diff
@@ -0,0 +1,340 @@
...
+        """Initialize an AbstractValue."""
+        super().__init__()
+        if not isinstance(tracks, Tracks):
+            tracks = Tracks(**tracks)
+        self.tracks = tracks
+
+    @property
+    def t(self):
+        """Returns the AbstractValue's tracks."""
+        return self.tracks
```
abergeron: I get that this is a shortcut, but is it really necessary?

myia/abstract/data.py
```diff
@@ -0,0 +1,340 @@
...
+                type=f"*{self.tracks.interface.__name__}",
+            )
+
+
+class AbstractUnion(AbstractValue):
+    """Base class for abstract values that are sum types."""
+
+    def __init__(self, options, tracks):
+        super().__init__(tracks)
+        self.options = list(options)  # Possibilities(options)
```
abergeron: That comment seems like leftovers.

abergeron (approved):

myia/abstract/data.py
breuleux: I suppose not

myia/abstract/data.py
breuleux: Ah, yeah

PR #399, opened 2021-05-05
First draft of the inference system

breuleux:
This is the new, generator-based inference system. It is mostly agnostic to the data structures on which it is to perform inference.


myia/abstract/data.py
```diff
@@ -318,7 +337,7 @@ def __hrepr_short__(self, H, hrepr):
 
 
 class Placeholder(GenericBase):
-    """Subtype of GenericBase used as placeholders when processing a node.
+    """Subtype of Generic used as placeholders when processing a node.
```
abergeron: It says GenericBase above...

myia/infer/algo.py
```diff
@@ -0,0 +1,286 @@
...
+        self.satisfied = {None: None}
+        self.todo = []
+
+    def _satisfy_request(self, request, node, value):
+        """Tell a request that the node's value has been inferred.
+
+        If all the request's nodes have a value, the payload and
+        request output will be appended to the todo list.
+        """
+        if request.satisfy(node, self.satisfied[node]):
```
abergeron: Can't you do this?

```suggestion
        if request.satisfy(node, value):
```

myia/infer/algo.py
```diff
@@ -0,0 +1,286 @@
...
+        self.unif = autils.Unificator()
+        self.scheduler = Scheduler()
+
+    def bootstrap(self, node):
+        """Start inference of the node.
+
+        If the node was not already inferred, set the inference result to a
+        placeholder and call `self.engine(node)` to start the inference.
+        """
+        existing = getattr(node, "abstract", None)
```
abergeron: All nodes have the abstract attribute and it's set to `None` by default.

myia/infer/infnode.py
```diff
@@ -0,0 +1,215 @@
...
+    def __init__(self, new_node):
+        self.new_node = new_node
+
+
+class InferenceEngine:
+    def __init__(self):
+        self.replacements = defaultdict(dict)
+
+    def __call__(self, node, unif):
+        if repl := self.replacements.get((None, None, node), None):
```
abergeron: Once again you don't want this to work on 3.8 :)

abergeron (commented):
Overall looks good.

I do think that we should do the replacements on-the-fly to avoid issues like we had with the old inferrer, but it seems like it would be easy to fix.

I have a few small cosmetic comments.

myia/abstract/data.py
breuleux: Ah yeah I'm not sure how that happened. Maybe a search and replace mess up.

myia/infer/algo.py
breuleux: True. That does seem to be the purpose of that argument...

myia/infer/algo.py
breuleux: The tests don't use nodes, but yeah, it may be better to require the field's presence.

myia/infer/infnode.py
breuleux: You mean 3.7? `:=` was added in 3.8, right?

PR #400, opened 2021-05-07
Add clone for graphs and subobjects and tests for it

abergeron:


myia/ir/node.py
```diff
@@ -74,6 +74,25 @@ def apply(self, fn, *inputs):
...
+        res.flags = self.flags.copy()
+        if self.return_:
+            res.return_ = self.return_.clone(self, objmap)
+        res.varargs = (
+            self.varargs.clone(self, objmap) if self.varargs else self.varargs
+        )
+        res.kwargs = (
+            self.kwargs.clone(self, objmap) if self.kwargs else self.kwargs
+        )
+        res.defaults = self.defaults
```
breuleux, line 96: Is there any chance of this dict being modified?

```suggestion
        res.defaults = dict(self.defaults)
```

myia/ir/node.py
```diff
@@ -157,6 +176,11 @@ def add_debug(self, **kwargs):
             for k, v in kwargs.items():
                 setattr(self.debug, k, v)
 
+    def _clone(self, old, objmap):
```
breuleux: Confusing name for what that method does.

breuleux (commented):
The one thing I think that's missing here is that cloning a graph should also clone its children, otherwise some nodes will refer to graphs that have parents that are not in scope.

myia/ir/node.py
abergeron: That dictionary should not be modified because it's the default values for the arguments.  The expression could be optimized, but that should be done once replaced in the call expression, not in  there.

myia/ir/node.py
abergeron: I'm open to better suggestions.

breuleux: `_copy_fields`? Or `_finalize_clone`?

abergeron: I'll go with `_copy_fields`

abergeron (commented):
I've added recursive copying for child graphs (those whose parent is self).

myia/ir/node.py
```diff
@@ -74,6 +74,28 @@ def apply(self, fn, *inputs):
 
         return Apply(self, *edges)
 
+    def clone(self, objmap=None):
+        """Make a copy of this graph."""
+        res = Graph(self.parent)
+        if objmap is None:
+            objmap = {self: res}
+        else:
+            objmap[self] = res
```
breuleux: We should probably check if the graph is already in the objmap, if e.g. we use it more than once.

```suggestion
        if objmap is None:
            objmap = {self: res}
        elif self in objmap:
            return objmap[self]
        else:
            objmap[self] = res
```

PR #401, opened 2021-05-11
Optimize universe nodes and handle r-values in Python backend.

notoraptor:
Hi @abergeron @breuleux This is a PR to optimize universe nodes in Python backend.

I also fix some bugs related to compilation of literal strings, which were previously handled as symbol names (`string` instead of `"string"`).

### DIFF
--- a/myia/compile/backends/python/code_generator.py
+++ b/myia/compile/backends/python/code_generator.py
@@ -12,96 +12,8 @@ from myia.compile.backends.python.implementations import (
     myia_iter,
     myia_next,
 )
-from myia.ir import Node
+from myia.ir import Constant
 from myia.utils.info import Labeler
-
-
-def default_formatter(c, format_string, nodes):
-    """Generate code using given format string and input nodes."""
-    return format_string.format(*[c.rvalue(node) for node in nodes])
-
-
-def str_user_switch(c, cond, if_true, if_false):
-    """Formatter for apply `user_switch`."""
-    return f"{c.label(if_true)} if {c.label(cond)} else {c.label(if_false)}"
-
-
-def str_make_tuple(c, *inputs):
-    """Formatter for apply `make_tuple`."""
-    return f"({', '.join(c.rvalue(inp) for inp in inputs)})"
-
-
-def str_make_list(c, *inputs):
-    """Formatter for apply `make_list`."""
-    return f"[{', '.join(c.rvalue(inp) for inp in inputs)}]"
-
-
-def str_make_dict(c, *inputs):
-    """Formatter for apply `make_dict`."""
-    assert not len(inputs) % 2
-    pairs = [c.rvalue(inp) for inp in inputs]
-    return (
-        "{"
-        + ", ".join(
-            "{}: {}".format(pairs[2 * i], pairs[2 * i + 1])
-            for i in range(len(pairs) // 2)
-        )
-        + "}"
-    )
-
-
-def str_apply(c, fn, args, kwargs):
-    """Formatter for apply `apply`, which seems to represent a call with args and kwargs."""
-    return f"{c.label(fn)}(*{c.rvalue(args)}, **{c.rvalue(kwargs)})"
-
-
-def str_getattr(c, *params):
-    """Formatter for apply getattr."""
-    assert len(params) == 2
-    obj, symbol = params
-    return f"{c.label(obj)}.{c.label(symbol)}"
-
-
-SIMPLE_MAP = {
-    operator.add: "{} + {}",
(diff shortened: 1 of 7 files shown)


myia/compile/backends/python/code_generator.py
```diff
@@ -190,22 +215,22 @@ def _register_global(self, name, value):
 
     def _node_to_expr(self, node):
         """Convert an apply node to an expr."""
-        fn = node.fn
+        node = self.replace.get(node, node)
+        fn = self.replace.get(node.fn, node.fn)
+        inputs = [self.replace.get(inp, inp) for inp in node.inputs]
```
breuleux: Is it possible that a node's replacement might itself have a replacement?

myia/compile/backends/python/code_generator.py
```diff
@@ -224,14 +249,15 @@ def _node_to_line(self, node):
 
     def label(self, node):
         """Get name for given node."""
-        return (
-            self.inline_nodes[node]
-            if node in self.inline_nodes
-            else self.lbl(node)
-        )
+        node = self.replace.get(node, node)
```
breuleux: Is this necessary? I assume that we already got the node's replacement in the caller to this function.

myia/compile/backends/python/code_generator.py
```diff
@@ -224,14 +249,15 @@ def _node_to_line(self, node):
...
+        node = self.rename.get(node, node)
+        if node in self.inline_nodes:
+            return self.inline_nodes[node]
+        return self.lbl(node)
 
     def rvalue(self, node):
         """Inline constant strings and get name for other nodes."""
+        node = self.replace.get(node, node)
```
breuleux: Same question as for label.

myia/compile/backends/python/optimizer.py
```diff
@@ -0,0 +1,105 @@
...
+                self.nonlocals.setdefault(dg.data, []).append(n_make_handle)
+
+    def _optimize_universe_getitem(self, dg: DirectedGraph):
+        """Optimize `universe_getitem` nodes for given directed graph."""
+        nodes_universe_getitem = []
+        for element in dg.value_to_node:
+            if isinstance(element, Apply):
+                fn = element.fn
+                if fn.is_constant() and fn.value == UNIVERSE_GETITEM:
+                    nodes_universe_getitem.append(element)
```
breuleux: The utility function `element.is_apply(UNIVERSE_GETITEM)` should take care of both conditions here.

myia/compile/backends/python/optimizer.py
```diff
@@ -0,0 +1,105 @@
...
+    def _optimize_universe_setitem(self, dg: DirectedGraph):
+        """Optimize `universe_setitem` nodes for given directed graph."""
+        # Dictionary mapping typename to myia node to assignation object
+        nodes = {TYPEOF: set(), MAKE_HANDLE: set(), UNIVERSE_SETITEM: set()}
+        for element in dg.value_to_node:
+            if isinstance(element, Apply):
+                fn = element.fn
+                if fn.is_constant() and fn.value in nodes:
+                    assert element not in nodes[fn.value]
+                    nodes[fn.value].add(element)
```
breuleux: I think the design for this could be made somewhat more generic, by e.g. having a function like `self.collect_nodes({TYPEOF, MAKE_HANDLE, UNIVERSE_SETITEM})` return the nodes dict you have here. That function could be reused for `_optimize_universe_getitem`, with a different argument.

myia/compile/backends/python/code_generator.py
notoraptor: Hi! Currently no, it's just one replacement by a new node, no recursion.

I don't think I will have to handle recursive recursion in the future, unless very complex optimization to do in backend.

myia/compile/backends/python/optimizer.py
notoraptor: Ok !

myia/compile/backends/python/optimizer.py
notoraptor: Ok !

notoraptor:
Hi @breuleux , PR updated !

I re-organized the `code_generator` code for a better follow-up of function usages.
- `rvalue()` indeeds always receives an already-replaced node. It is now a private method `_rvalue()`.
- `label()` may be called from outside code generator (in Python backend to get main graph name), so it still replaces node if necessary (although I currently replace only nodes, not graphs). Anyway, I added a method `_label()`, for specific labeling of already-replaced nodes. `label()` uses `_label()`.

PR #403, opened 2021-05-12
Add relations to the parser

abergeron:
This add more `about` relations to the output of the graph.


abergeron:
I've rebased and fixed this.

PR #404, opened 2021-05-18
Graph visualization for new IR

notoraptor:
Hi ! This is my current code for graph visualization using hrepr. Code logic is into a hrepr mixin.

For function:
```python
def f(x):
    while x:
        x = x - 1
    return x
```

I currently get this output:

output

FN edges are red, input edges are default gray, SEQ edges are dashed.
Parameter nodes are green, apply nodes are blue, constant nodes are pink, return nodes are orange.

TODO:
* Improve labeling. In grave above, sometimes, parameter is named `x.2` and `make_handle` is named `x`. It should be the opposite. I guess I need to define a deterministic visit order when generating node labels.
* Maybe do not display all constant nodes. For example, I think I could remove the `node.fn` constants if they are just constant strings like `universe_setitem`, as apply node already prints it.

@breuleux


notoraptor:
Hi @breuleux , PR updated ! Should be ready for review !

I added a GraphPrinter with 3 boolean parameters:
- `show_constants` to show constant nodes (other than constant graphs). Constant graphs are displayed in any case.
- `link_fn_graphs` to show edges from apply FN slots to constant graphs (e.g. for `#1 = while_header()` node).
- `link_inp_graphs` to show edges from apply input slots to constant graphs (e.g. for `user_switch` inputs).

I added a test that generates hrepr HTML outputs for all combinations of these 3 parameters, and then compares outputs to expected. Expected outputs are in HTML files in test folder, and can also be used as visualization examples: https://github.com/notoraptor/myia/tree/snekgraph/tests/visualization . Function `f0` correctly demonstrates output changes among parameters as it uses constant graphs and user_switch. Function `f1` is just here to test constants.

HTML code also seems to be deterministic now.

myia/ir/visualization/graph_printer.py
```diff
@@ -0,0 +1,277 @@
...
+                },
+                export="make_graph",
+            ),
+        ]
+
+    def __hrepr__(self, H, hrepr):
+        # Generate identifiers to make hrepr output deterministic.
+        elements = self.graphs + self.nodes
+        identifiers = {
+            element: indice for indice, element in enumerate(elements)
```
breuleux: Probably better to convert them all to strings upfront.

```suggestion
            element: str(index) for index, element in enumerate(elements)
```

tests/visualization/test_graph_printer.py
```diff
@@ -0,0 +1,58 @@
...
+    with enable_debug():
+        graph = parse(function)
+    gp = GraphPrinter(
+        graph,
+        show_constants=bool(show_constants),
+        link_fn_graphs=bool(link_fn_graphs),
+        link_inp_graphs=bool(link_inp_graphs),
+    )
+    html = hrepr.page(gp)
+    expected_filename = f"{function.__name__}_{show_constants}{link_fn_graphs}{link_inp_graphs}.html"
```
breuleux, line 54: Do you have a utility function to regenerate the HTML pages, if we want to make some changes to the graph generation and update the tests? That would be handy.

myia/ir/visualization/graph_printer.py
notoraptor: Ok !

tests/visualization/test_graph_printer.py
notoraptor: Ok !

notoraptor:
@breuleux Updated !

PR #406, opened 2021-05-27
Use functions instead of strings in function nodes

breuleux:
This replaces the strings `"return_"`, `"universe_getitem"`, and so on that are currently in some constant nodes produced by the parser by builtin functions when appropriate, or functions in `myia.basics` when they are Myia-specific. Tests and representations are adjusted accordingly.


myia/basics.py
```diff
@@ -0,0 +1,131 @@
...
+        self.value = value
+
+
+def make_handle(object_type):
+    """Create a handle of the given type."""
+    return Handle(object_type)
+
+
+def global_universe_getitem(h):
+    """Set the value of a handle."""
```
notoraptor: I think it should be "Get the value of a handle" here.

myia/basics.py
```diff
@@ -0,0 +1,131 @@
...
+    return h.value
+
+
+def global_universe_setitem(h, value):
+    """Get the value of a handle."""
```
notoraptor: And "Set" here.

notoraptor (commented):
I only checked changes in python backend, it seems good for me! But there is a typo in documentation for `global_universe` methods.

myia/basics.py
breuleux: Oh, right, haha.

myia/ir/print.py
```diff
@@ -131,7 +149,7 @@ def __init__(self):
         )
 
     def __call__(self, node):
-        if node.is_constant_graph():
+        if isinstance(node, Node) and node.is_constant_graph():
```
abergeron, line 152: This should always be called with a Node so that check should not necessary

abergeron (approved):

abergeron:
I didn't really look at the html pages, but the rest looks good.

myia/ir/print.py
breuleux: It is now used for graph_printer as well, which passes Graphs sometimes.

abergeron: Ok then.

PR #407, opened 2021-05-27
Use phi arguments instead of universe for local closures

abergeron:
This is complete and working in all cases of local closures that I could find.

It may have trouble in the future with `IfExp` if we add support for the `:=` operator.


abergeron:
This PR will conflict with #406 for the parser tests.

abergeron:
I'll fix the failing print tests once we merge #406 to avoid doing the work twice.

breuleux (approved):
Looks good to me

notoraptor (approved):
Looks good to me!

PR #408, opened 2021-05-27
Add utility functions toposort(), get_node_users(), and use them to optimize universe_getitem after graph parsing

notoraptor:
Hi ! This is a PR to optimise `universe_getitem` nodes after parsing. As I needed `toposort()` and `get_node_users` to do it easily, I implemented these both utility functions here too. Commits are correctly separated, so I could split PR if necessary.

Testing is also added for utility functions and optimization.

@abergeron @breuleux


notoraptor:
PR #406 should be merged and then this one rebased before merging.

myia/compile/backends/python/optimizer.py
```diff
@@ -36,15 +36,15 @@ def optimize(self, directed_graphs):
         todo = list(directed_graphs)
         seen = set()
         while todo:
-            dg = todo.pop(0)
+            dg = todo.pop(0)  # type: DirectedGraph
```
breuleux: Might as well use an annotation:

```suggestion
            dg: DirectedGraph = todo.pop(0)
```

myia/utils/directed_graph.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def visit(self):
+        """Visit directed graph from user to used values.
+
+        :return: generator: sequence of graph values in visited order.
+        """
+        # Use a copy to visit graph, so that current graph is not modified.
+        cp = self._copy()
+        # Starts with unused nodes.
+        todo = [n for n in cp.uses if n not in cp.used_by]
```
breuleux: It would be better to use `collections.deque` instead of a list, the former is more efficient when popping from the left side.

myia/ir/graph_utils.py
```diff
@@ -0,0 +1,68 @@
+"""Utility functions for myia graph."""
+from myia.ir.node import Apply, Graph
+from myia.utils.directed_graph import DirectedGraph
+
+
+def toposort(graph: Graph, reverse=True):
+    """Return graph nodes in topological order.
+
+    :param graph: myia graph
+    :param reverse: if True, return nodes in execution order (default)
```
breuleux: I feel the meaning of `reverse` should be the opposite (so it'd be False by default and you'd do `output.reverse()` if reverse is False).

breuleux (commented):
I wrote a few review comments last week but it appears that I forgot to post them. I haven't gone through everything yet.

myia/ir/graph_utils.py
notoraptor: Hi ! By default, as I visit graph from `return_` node to parameters, I get the topological order from user to used nodes, which is not the execution order. As I guess people would like to get execution order, I set `reverse` to True by default, so that one must explicitly set it to `False` to get raw order (user to used), e.g. if execution order is not necessary, or to save time (skipping reversing).

Should I still set `reverse` to False by default ?

Or, for disambiguation, should I rename `reverse` to something like `execution_order` ?

notoraptor:
Hi @breuleux @abergeron PR rebased and updated ! I still keep post-parsing opt (added here) and Python backend opt, because there are still cases where phi is not used, so that graph still contains universe methods. I added example cases into `test_parser`.

What do you think ?

myia/compile/backends/python/optimizer.py
notoraptor: Ok!

myia/utils/directed_graph.py
notoraptor: OK!

myia/ir/graph_utils.py
breuleux: I meant that "reverse" being True should mean the opposite order of whatever the default is, so in this case "reverse" would be from user to used, and you'd have to set reverse to True to get the "raw" order. Basically, if a function has a `reverse` argument I expect it to be `False` by default -- the fact that you have to reverse the raw list to get the execution order is an implementation detail, it doesn't have to correspond to the function's API.

notoraptor: OK, updated !

myia/parser.py
```diff
@@ -77,14 +78,17 @@ def __init__(self, msg, loc):
 _parse_cache = {}
 
 
-def parse(func):
+def parse(func, parser_opt=False):
```
abergeron: I'm of the opinion that parser_opt should be always on.

I fail to see a case where one would want to genuinely turn it off.

myia/parser_opt.py
```diff
@@ -0,0 +1,90 @@
...
+    mapping = {node: new_node}
+    mapping_seq = {node: node.edges[SEQ].node} if SEQ in node.edges else {}
+    seen_graphs = set()
+    todo_graphs = deque([root_graph])
+    while todo_graphs:
+        g = todo_graphs.popleft()
+        if g in seen_graphs:
+            continue
+        seen_graphs.add(g)
+        g.replace(mapping, mapping_seq, recursive=False)
```
abergeron: Why are you writing a complex wrapper to do what recursive does?

myia/parser_opt.py
```diff
@@ -0,0 +1,90 @@
...
+            # To optimize, make_handle must be used by:
+            # - only 1 universe_setitem
+            # - at least 1 universe_getitem to replace
+            if nodes_getitem and len(nodes_setitem) == 1:
+                (n_setitem,) = nodes_setitem
+                n_value = n_setitem.inputs[1]
+                n_typeof = n_make_handle.inputs[0]
+                # Replace universe_getitem nodes with associated value.
+                for n_getitem in nodes_getitem:
+                    _replace_apply_node(graph, n_getitem, n_value)
```
abergeron, line 66: Please use `Graph.replace` directly and add a dictionary to replace all the getitems at once.  With this code you are scanning the graph once per getitem node which is unnecessary and slow.

myia/parser_opt.py
```diff
@@ -0,0 +1,90 @@
...
+def _replace_apply_node(root_graph: Graph, node: Apply, new_node):
+    """Replace node by new_node in given graph.
+
+    :param root_graph: graph in which to make replacements.
+        Replacement will be recursively applied to graph and all children closures.
+    :param node: node to replace
+    :param new_node: node to replace with. May be None to delete node.
+    """
+    mapping = {node: new_node}
+    mapping_seq = {node: node.edges[SEQ].node} if SEQ in node.edges else {}
```
abergeron: If you want to delete the SEQ edge, map it to `None`.  An empty dictionary will just use the normal mapping

myia/parser_opt.py
```diff
@@ -0,0 +1,90 @@
...
+                (n_setitem,) = nodes_setitem
+                n_value = n_setitem.inputs[1]
+                n_typeof = n_make_handle.inputs[0]
+                # Replace universe_getitem nodes with associated value.
+                for n_getitem in nodes_getitem:
+                    _replace_apply_node(graph, n_getitem, n_value)
+                # Remove typeof, universe_setitem and make_handle nodes.
+                _replace_apply_node(graph, n_typeof, None)
+                _replace_apply_node(graph, n_make_handle, None)
+                _replace_apply_node(graph, n_setitem, None)
```
abergeron: This is invalid, you can't replace a node by `None`, but it works by accident because there are no actual users except the seq chain for the setitem.

Also it will break the sequence chain unless the setitem is the first element.

What you want to do is something like this:
```
prev = n_setitem.edges.get(SEQ, None)
if prev is not None:
    prev = prev.node
graph.replace({}, {n_setitem, prev})
```

tests/test_parser.py
```diff
@@ -229,6 +229,144 @@ def g():
         )
 
 
+def test_self_recursion_parser_opt():
+    def f():
```
abergeron: For coverage mark this function with `# pragma: no cover` (and all the other tests functions)

tests/test_parser.py
```diff
@@ -229,6 +229,144 @@ def g():
...
+
+graph g() {
+  #2 = g()
+  return #2
+}
+"""
+        )
+
+
+def test_universe_on_string():
```
abergeron: I'm curious what special properties strings have that integers (or any other value) doesn't.

abergeron (changes requested):

myia/parser_opt.py
abergeron: This could be simplified by a property on Apply nodes similar to `.fn` that returns the next element in the sequence chain or `None` if it's the first, but that is optional for now.

abergeron (approved):

PR #409, opened 2021-06-01
Improve visualizations

breuleux:
* `Labeler` uses methods for disambiguation etc. instead of passing functions to the constructor.
* Subclass `Labeler` into `myia.ir.print.NodeLabeler`, the labeler for nodes
* `NodeLabeler.informative` returns a more informative label for Apply nodes
* Add a global NodeLabeler: `myia.ir.print.global_labeler`
* `Node.__str__` uses `global_labeler.informative`
* Add the `show_args` parameter to `GraphPrinter`
* Use `~` instead of `.` to disambiguate names


tests/utils/test_info.py
```diff
@@ -2,8 +2,8 @@
 
 import pytest
 
-from myia.utils.info import (
-    AbbreviationTranslator,
+from myia.utils.info import (  # AbbreviationTranslator,
```
notoraptor: Should this comment be removed?

myia/ir/visualization/graph_printer.py
```diff
@@ -21,24 +20,29 @@ def __init__(
...
+        labeler=None,
     ):
         """Initialize.
 
         :param graph: graph to visualize
         :param on_node: optional function to call on printed node when clicked
-        :param show_constants: if True, display constant non-graph nodes.
-            Constant graphs are always displayed.
+        :param show_fn_constants: if True, display constant nodes when they are used
+            in the FN edge of an Apply node.
```
notoraptor: Documentation for parameter `show_args` is missing.

tests/visualization/gen_expected_files.py
```diff
@@ -10,26 +11,43 @@
 
 
 def _generate_expected_files(*functions):
+    for (
+        function,
+        show_fn_constants,
+        show_args,
+        link,
```
notoraptor, line 18: Same value `link` is used for both parameters `link_fn_graphs` and `link_inp_graphs`. Is it voluntary ?

notoraptor (commented):
Hi @breuleux ! I made few comments ! Overall seems good for me.

tests/utils/test_info.py
breuleux: Yes, good catch.

myia/ir/visualization/graph_printer.py
breuleux: True, I'll fix that.

Normally pydocstyle should spot missing parameters, I think the reason it doesn't in this case is that this is a different doc format as the one declared for pydocstyle. I might fix that as well.

tests/visualization/gen_expected_files.py
breuleux: Yes, I didn't want to generate an exponential number of tests in the number of configuration parameters, and these seemed similar enough.

notoraptor: Ok !

notoraptor (approved):

breuleux:
Rebased

PR #410, opened 2021-06-07
Parse positional-only arguments

notoraptor:
Hi @abergeron , this is a PR that adds support for positional-only arguments to parser. I also updated the Python backend to correctly generate function parameters list.

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -5,6 +5,22 @@ from myia.parser import MyiaSyntaxError, parse
 from myia.utils.info import enable_debug
 
 
+def test_posonly_args():
+    def f1(a, b, c, d, e):  # pragma: no cover
+        pass
+
+    g1 = parse(f1)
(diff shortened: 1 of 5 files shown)


tests/test_parser.py
```diff
@@ -5,6 +5,22 @@
 from myia.utils.info import enable_debug
 
 
+def test_posonly_args():
+    def f1(a, b, c, d, e):  # pragma: no cover
+        pass
+
+    g1 = parse(f1)
+    assert len(g1.parameters) == 5
+    assert g1.posonly == 0
```
abergeron: There are already tests for normal parameters so I don't think we need to test it again.

Also, it would be nice if this test would go with the other `test_def` below.

notoraptor:
@abergeron updated ! Is it what you expect, for the test ?

abergeron (approved):

abergeron:
Yep, that's good.

PR #412, opened 2021-06-08
Fix problems with closure variables defined in branches

abergeron:
This also fixes some issues around variable classification in analyze and phis that could reach out of the function.

Finally, I've added some failing tests of weird python corner-case name resolution issues that we might want to error out on.


notoraptor:
This looks good to me ! Do you have something else to add, or could we merge this PR ?

notoraptor:
Rebase needed!

notoraptor (approved):

PR #414, opened 2021-06-10
Add backlinks to Node and Edge

abergeron:
This adds backlinks and uses them for replaces, removing all the now-unnecessary code.

This contains a commit that is duplicated with #412 (Stop using None as a node value), so there might be some small rebase work to do depending on which one is merged first.

### DIFF
--- a/myia/ir/node.py
+++ b/myia/ir/node.py
@@ -123,46 +123,39 @@ class Graph:
 
         A node can be in either mapping or mapping_seq or both.
         """
-        self._replace(mapping, mapping_seq, recursive, self, set())
+        for node, repl in mapping_seq.items():
+            if repl is None:
+                self.delete_seq(node)
+            else:
+                self.replace_node(node, SEQ, repl, recursive=recursive)
 
-    def _replace(
-        self, mapping, mapping_seq, recursive, root_graph, seen_graphs
-    ):
-        if self in seen_graphs:
-            return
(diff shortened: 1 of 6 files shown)


myia/ir/node.py
```diff
@@ -216,11 +181,12 @@ class Node:
       annotation: Defined type for this node, optional
     """
 
-    __slots__ = ("abstract", "annotation", "debug", "__weakref__")
+    __slots__ = ("abstract", "annotation", "users", "debug", "__weakref__")
 
     def __init__(self):
         self.abstract = None
         self.annotation = None
+        self.users = weakref.WeakSet()
```
breuleux, line 189: The order isn't deterministic, right? It might end up working just fine, but this could potentially lead to issues during optimization (chiefly if there are subtle bugs where the order of replacement accidentally matters).

myia/ir/node.py
```diff
@@ -299,11 +268,33 @@ def clone(self, g, objmap):
...
+        if self._user:
+            return self._user()
+
+    @user.setter
+    def user(self, node):
+        self._user = weakref.ref(node)
+
 
-def _edgemap(edges):
+def _edgemap(edges, self):
```
breuleux: It'd be best not to name that argument `self`, I think.

abergeron: ok

breuleux (approved):

PR #415, opened 2021-06-22
Support assignment expressions.

notoraptor:
Hi ! This is the PR to support assignment expressions. It seems to work well with addition of just ast.NamedExpr support. I am still looking for corner cases.

What do you think ? @abergeron @breuleux

### DIFF
--- a/myia/parser.py
+++ b/myia/parser.py
@@ -954,6 +954,11 @@ class Parser:
 
         return block
 
+    def _process_NamedExpr(self, block, node: ast.NamedExpr):
+        val = self.process_node(block, node.value)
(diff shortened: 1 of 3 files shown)


breuleux:
Looks good, maybe just add a test of the form `f(x, x := y, x)` to make sure the order of operations is properly preserved, it should be equivalent to `f(x, y, y)`.

notoraptor:
@breuleux done. I don't know why, but CI is not triggered.

notoraptor:
CI tests ultimately succeed.

breuleux (approved):

PR #418, opened 2021-07-15
Support augmented assignments.

notoraptor:
@abergeron @breuleux


myia/parser.py
```diff
@@ -958,6 +958,20 @@ def _process_Assign(self, block, node):
 
         return block
 
+    def _process_AugAssign(self, block, node: ast.AugAssign):
+        # Replace `a {op}= b` with `a = a {op} b`
```
breuleux: That's not how Python translates it. `a += b` translates to `a = operator.iadd(a, b)`. Most of the operators in the `ast_map` have an `iop` equivalent that should be used for augmented assignment, and those that don't cannot be used in augmented assignment.

breuleux (changes requested):

myia/compile/backends/python/code_generator.py
```diff
@@ -83,12 +83,26 @@ def __init__(self, *, skip=None, nonlocals=None, replace=None, rename=None):
             operator.ge: "{} >= {}",
             operator.getitem: "{}[{}]",
             operator.gt: "{} > {}",
+            operator.iadd: "{} + {}",
```
breuleux: That's incorrect. The codegen for `operator.iadd` should be `operator.iadd`.

breuleux (changes requested):

breuleux (approved):

PR #420, opened 2021-07-30
Use string edges for keyword arguments

abergeron:


abergeron:
I have a problem with the python backend for Apply.  It doesn't parse applies correctly and assumes that it's `Apply(fn, args, kwargs)`.

I'll try to fix that next week.

myia/basics.py
```diff
@@ -143,3 +142,11 @@ def return_(x):
...
     """Resolve the name in the given namespace."""
     return ns[key]
+
+
+def concat(*args):
+    """Contatenate the arguments in a tuple.
+
+    This is the same functionally as make_tuple, but used for macros.
+    """
+    return args
```
breuleux: `concat` should concatenate the arguments, i.e. `reduce(operator.add, args, ())`. I'm not sure why you would concatenate the arguments in `apply` rather than have `concat` do it.

myia/compile/backends/python/code_generator.py
```diff
@@ -331,7 +339,14 @@ def _str_make_dict(self, *inputs):
...
         """Formatter for apply `apply`, which seems to represent a call with args and kwargs."""
-        return f"{self._label(fn)}(*{self._rvalue(args)}, **{self._rvalue(kwargs)})"
+        params = []
+        l, _ = args.args()
+        for e in l:
+            params.append(f"*{self._rvalue(e)}")
+        l, _ = kwargs.args()
+        for e in l:
+            params.append(f"**{self._rvalue(e)}")
+        return f"{self._label(fn)}({', '.join(params)})"
```
breuleux: Requiring `apply` to have two `concat` as arguments (and not checking that they are `concat`) is somewhat fragile. I would find it less ad hoc to have `concat` actually do something and generate `f(*tuple_concat(a, b), **dict_concat(c, d))`, or `f(*(*a, *b), **{**c, **d})`. Then `apply` can take arbitrary expressions and still work.

myia/ir/print.py
```diff
@@ -19,8 +19,12 @@ def _print_node(node, buf, nodecache, offset=0):
...
     print(f"{nodecache(node.fn)}(", end="", file=buf)
+    args, kwargs = node.args()
+    args = [nodecache(a) for a in args]
+    args.extend(f"{k}={nodecache(v)}" for k, v in kwargs.items())
+
     print(
-        ", ".join(nodecache(a) for a in node.inputs),
+        ", ".join(args),
         end="",
         file=buf,
```
breuleux:
```suggestion
        file=buf
```

I believe that'll make `black` collapse the call into a single line.

myia/compile/backends/python/code_generator.py
abergeron: I'm ok with that but then how is that really different from star and dstar?

breuleux: I suppose it's not? I'm not sure what you mean.

abergeron:
This has been updated to reflect review comments.

myia/basics.py
```diff
@@ -72,19 +72,18 @@ def global_universe_setitem(h, value):
...
     kwargs = {}
-    for g in groups:
-        if isinstance(g, tuple):
-            args.extend(g)
-        else:
-            kwargs.update(g)
+    for a in args_:
+        args.extend(a)
+    for k in kwargs_:
+        kwargs.update(k)
```
breuleux: You forgot to fix `basics.apply`.

abergeron: Yep

breuleux (approved):

PR #421, opened 2021-08-02
Inference tests from master branch

notoraptor:
I reopend my PR here, as @breuleux PR is now merged.
- Still WIP, I will fix more tests today.
- Numpy dependencies removed.
- Basic support for dict added.
- currently 187 failed, 170 passed, 3 xfailed among added tests


notoraptor:
Update.

For added tests:

`pytest -vvs tests/operations/ tests/test_infer.py`:
- `131 failed, 202 passed, 26 xfailed in 22.92s`

I think I won't improve it too much from now to meeting, so I stop here for now.

What do you think ? @breuleux

notoraptor:
Well, I guess it's ready for a review and possible merging now, @breuleux !

myia/abstract/data.py
```diff
@@ -273,6 +279,33 @@ def __hrepr__(self, H, hrepr):
...
+
+    @property
+    def keys(self):
+        """Get list of key types."""
+        return self.elements[::2]
+
+    @property
+    def values(self):
+        """Get list of value types."""
+        return self.elements[1::2]
```
breuleux: I would rather avoid having a specific `AbstractDict` type here. The basic idea behind Abstract is that all the subtypes are the elements, and the interface contains the necessary information to figure out what to do with them. In this case, for e.g. `{"a": 1, "b": 2.3}` I would use something like `AbstractStructure((int, float), {"interface": dict_with_keys("a", "b")})`. The int and float are the things we can't throw away because they are the types of the actual things we need to put in the dictionary, whereas the keys "a" and "b" we can eventually throw away along with the interface when we replace the dict by a simple tuple.

myia/infer/inferrers.py
```diff
@@ -47,52 +63,486 @@ def partial_inferrer(node, args, unif):
...
+        raise AssertionError(
+            f"getattr can currently only be used for methods, got {interface}.{result} (type {type(result)})"
+        )
         # new_node = Constant(result)
     yield Replace(new_node)
     res = yield Require(new_node)
     return res
 
 
+def getitem_inferrer(node, args, unif):
```
breuleux, line 94: Instead of special casing various interfaces, this can dispatch to `x.__getitem__(y)` through `Replace`, and the various builtin types' getitem methods can have inferrers.

myia/infer/inferrers.py
```diff
@@ -47,52 +63,486 @@ def partial_inferrer(node, args, unif):
...
+    return data.AbstractStructure(elements, {"interface": list})
+
+
+def make_dict_inferrer(node, args, unif):
+    """Inferrer for the make_dict function."""
+    arg_types = yield RequireAll(*args)
+    return data.AbstractDict(arg_types)
+
+
+def len_inferrer(node, args, unif):
```
breuleux, line 202: This can just dispatch to `x.__len__()`.

myia/infer/inferrers.py
```diff
@@ -47,52 +63,486 @@ def partial_inferrer(node, args, unif):
...
+        b_casted = node.graph.apply(cls, b_node)
+        new_node = node.graph.apply(getattr(cls, bin_op), a_node, b_casted)
+        yield Replace(new_node)
+        res = yield Require(new_node)
+        return res
+
+
+def operator_add_inferrer(node, args, unif):
+    """Inferrer for the operator.add function."""
+    return _bin_op_inferrer("__add__", "__radd__", node, args, unif)
```
breuleux: It would be more economical to have `_bin_op_inferrer("__add__", "__radd__")` return the inferrer, and write that directly in the inferrers directly. (Same for below).

tests/infer/from_master/master_inferrers.py
```diff
@@ -0,0 +1,87 @@
+"""Supplementary inferrers for testing.
```
breuleux: Could you keep `from_master` in a different branch? I don't see a purpose in merging these in.

notoraptor: Done!

myia/infer/inferrers.py
notoraptor: Done!

myia/infer/inferrers.py
notoraptor: Shouldn't `x.__len__()` return an integer in any case ? If so, I think we could just check that interface has the `__len__` method, then assume return type is a `int`.

breuleux: It is true that `len` forces its result to be an integer, but if the user defines their own `__len__` method and it returns something else than an integer, we would like to be able to report an error.

myia/infer/inferrers.py
notoraptor: Done!

myia/infer/inferrers.py
notoraptor: OK, done !

myia/abstract/data.py
notoraptor: OK ! I removed AbstractDict and just use an AbstractStructure instead. I keep the same code logic (ie. interface as `dict`, and elements as list of flattened key-value types, corresponding to what `make_dict` would receive).

This allows a basic support for dicts, but I guess dict support will be modified/updated later for more complex cases.

notoraptor:
@breuleux PR updated!

myia/abstract/map.py
```diff
@@ -366,6 +369,12 @@ def abstract_map2(  # noqa: F811
     if len(x.elements) != len(y.elements):
         raise MapError(x, y, reason="Structures have different lengths")
 
+    if x.tracks.interface is dict:
+        # Just check keys, tracks will be checked below.
+        for kx, ky in zip(x.elements[::2], y.elements[::2]):
+            if kx is not ky:
+                raise MapError(kx, ky, "Different dict keys")
+
```
breuleux: The keys should be in the interface, though (e.g. a class `DictWithKeys("x", "y")`), and if the keys are different, the interfaces would be different, so there would be no need to special case anything.

myia/abstract/to_abstract.py
```diff
@@ -20,11 +21,35 @@ def to_abstract(self, xs: tuple):
...
+    # Override `to_abstract` instead of `precise_abstract` so that
+    # dict string keys are abstracted will string values in any case.
+    return data.AbstractAtom({"interface": type(x), "value": x})
+
+
+@ovld
+def to_abstract(self, x: dict):  # noqa: F811
+    return data.AbstractStructure(
+        [self(el) for item in x.items() for el in item], {"interface": dict}
+    )
```
breuleux: They keys should not be in the structure's elements.

notoraptor:
@breuleux PR updated about dicts, anyway !

I'm working on splitting it.

PR #426, opened 2021-08-12
Improve inference for apply nodes if FN node is from a parameter

notoraptor:
Use case:
```python
def cast(t, x):
    return t(x)
```
In such case, `t` comes with an abstract type and never goes through inference logic (as it's a parameter node, not a constant nor an apply). So, inference for apply `t(x)` won't be able to get inference function for `t`. This PR tries to fix it.

- Check apply's FN node abstract to get either inference function if available, or lookup inference in inferrers if FN has an abstract type.
- Add a utility method dispatch_inferences() to set multiple signatures for an inference.
- use dispatch_inferences() for constructors inferences: bool, int, float.
- Add a test for this case (test_scalar_cast).

Extracted from #421 

@breuleux @abergeron

### DIFF
--- a/myia/infer/infnode.py
+++ b/myia/infer/infnode.py
@@ -4,8 +4,6 @@ import types
 from dataclasses import dataclass
 from typing import Dict, Sequence, Tuple
 
-from ovld import OvldMC
-
 from myia.ir.node import SEQ
 
 from ..abstract import data, utils as autils
@@ -103,7 +101,7 @@ def signature(*arg_types, ret):
     return inference_function(_infer)
 
 
-class InferenceDefinition(metaclass=OvldMC):
+class InferenceDefinition:
     """Helper class to represent a signature."""
 
     __slots__ = "arg_types", "ret_type"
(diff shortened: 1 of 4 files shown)


myia/infer/infnode.py
```diff
@@ -89,14 +92,70 @@ def _infer(node, args, unif):
...
+        )
+        self.ret_type = _type_to_abstract(ret_type)
+
+    @classmethod
+    def without_value(cls, abstract: data.AbstractAtom):
+        """Get pure type from given abstract, without associated value.
+
+        Useful to lookup node abstract in registered signatures.
+        """
+        return data.AbstractAtom({"interface": abstract.tracks.interface})
```
breuleux: You should be able to do this more generally with something like this:

```python
@abstract_map.variant
def without_value(x: ValueTrack):
    return ValueTrack(ANYTHING)
```

That way, it also works recursively.

myia/infer/infnode.py
```diff
@@ -89,14 +92,70 @@ def _infer(node, args, unif):
...
+            f"expected {len(arg_types)}, got {len(inp_types)}"
+        )
         for inp_type, expected_type in zip(inp_types, arg_types):
             autils.unify(expected_type, inp_type, U=unif)
         return autils.reify(return_type, unif=unif.canon)
 
     return inference_function(_infer)
 
 
+class InferenceDefinition(metaclass=OvldMC):
```
breuleux: What does it need `OvldMC` for?

myia/infer/infnode.py
notoraptor: In fact, it's useless in current branch, so I removed it. I may apply your suggestions in next extracted PRs.

myia/infer/infnode.py
notoraptor: Usless here, indeed. Removed.

notoraptor:
@breuleux PR updated.

breuleux (approved):

PR #428, opened 2021-08-17
Build abstract types for arbitrary objects

breuleux:
* `to_abstract` on an arbitrary object generates an `AbstractStructure` with an interface that lists the object's fields, and the types are in the elements.
* `recursivize` generalizes instances into recursive abstract types, e.g. `Cons(int, Cons(int, empty))` => `#1=Cons(int, Union(#1, empty))`. It is not part of the pipeline for the time being.

### DIFF
--- a/tests/abstract/test_utils.py
+++ b/tests/abstract/test_utils.py
@@ -1,4 +1,5 @@
 from myia.abstract import data
+from myia.abstract.to_abstract import to_abstract
 from myia.abstract.utils import (
     MapError,
     canonical,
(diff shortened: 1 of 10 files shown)


tests/abstract/test_utils.py
```diff
@@ -182,3 +184,66 @@ def test_merge():
...
+
+    pt = Point(Point(1, 2), 3)
+    pt2 = Point(Point(Point(4, 5), 6), 7)
+
+    apt = A(pt)
+    apt2 = A(pt2)
+
+    rapt = recursivize(apt, parents={}, tagger=tg_p)
+    rapt2 = recursivize(apt2, parents={}, tagger=tg_p)
+    assert rapt is rapt2
```
notoraptor, line 249: Hi ! Could you add here an additional assert for how `rapt` is printed ?

breuleux: Done!

notoraptor (approved):
Looks good to me!

PR #429, opened 2021-08-17
Improve inference for builtin oeprators.

notoraptor:
Extracted from #421 

@breuleux @abergeron

### DIFF
--- a/myia/infer/inferrers.py
+++ b/myia/infer/inferrers.py
@@ -172,11 +172,18 @@ def _bin_op_dispatcher(bin_rop, *signatures):
     return dispatch_inferences(*signatures, default=_bin_rop_inferrer(bin_rop))
 
 
-def _bin_op_right_cast_inference(cls, bin_op):
+def _bin_op_right_cast_inference(cls, bin_op, ret=None):
     """Create a generic inference for binary op/rop methods.
 
     Replace right operand with a cast to left type if necessary.
+
+    If both operands have same interface cls,
+    return abstract atom with either cls or ret (if not None) as interface.
     """
+    if ret is None:
+        ret = cls
+    else:
+        assert isinstance(ret, type)
 
     def inf(node, args, unif):
         """Inferrer for binary op/rop methods (e.g. float.__mul__).
@@ -190,7 +197,7 @@ def _bin_op_right_cast_inference(cls, bin_op):
         b_interface = b_type.tracks.interface
         assert a_interface is cls, f"expected {cls}, got {a_interface}"
         if b_interface is cls:
-            return data.AbstractAtom({"interface": cls})
+            return data.AbstractAtom({"interface": ret})
         else:
             b_casted = node.graph.apply(cls, b_node)
             new_node = node.graph.apply(getattr(cls, bin_op), a_node, b_casted)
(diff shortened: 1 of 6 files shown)


notoraptor:
Rebased and updated.

myia/infer/inferrers.py
```diff
@@ -82,22 +82,168 @@ def getattr_inferrer(node, args, unif):
...
+        """Generic inferrer for builtin binary operator functions."""
+        a_node, b_node = args
+        a_type = yield Require(a_node)
+        b_type = yield Require(b_node)
+
+        if isinstance(a_type, data.GenericBase) or isinstance(
+            b_type, data.GenericBase
+        ):
+            # If there is any generic in operands, just unify them.
+            return autils.unify(a_type, b_type)[0]
```
breuleux, line 116: This assumes that both operands have the same type (otherwise they would not be unifiable), which is not always true. For instance, it is untrue with `==`.

This is a delicate problem to solve and it might require to implement the inference subsystem that I mentioned a few weeks ago, where the inference of anything that involves generics is delayed as much as possible. The thing is, if `a_type` is a generic, trying to unify it with `b_type` is basically a guess, and we'd like to avoid ever having to guess.

To be clear I'm not asking for this feature in this PR, but tests that require this unification will have to be marked xfail.

myia/infer/inferrers.py
```diff
@@ -82,22 +82,168 @@ def getattr_inferrer(node, args, unif):
...
+
+        a_interface = a_type.tracks.interface
+        b_interface = b_type.tracks.interface
+        if hasattr(a_interface, bin_op):
+            new_node = node.graph.apply(
+                getattr(a_interface, bin_op), a_node, b_node
+            )
+        elif hasattr(b_interface, bin_rop):
```
breuleux, line 126: Just adding a TODO to make sure we don't forget to adjust this when we have the capability (not yet).

```suggestion
        # TODO: In order to support user-defined graphs for bin_op, the case where
        # it returns NotImplemented will have to be handled.
        elif hasattr(b_interface, bin_rop):
```

myia/infer/infnode.py
```diff
@@ -80,6 +80,18 @@ def _type_to_abstract(el):
     return type_to_abstract(el)
 
 
+def _without_value(abstract: data.AbstractAtom):
```
breuleux: Missing the decorator, also this is basically the `broaden` function that we had before.

```suggestion
@abstract_map.variant
def broaden(abstract: data.AbstractAtom):
```

myia/infer/infnode.py
```diff
@@ -80,6 +80,18 @@ def _type_to_abstract(el):
...
+    """Get pure type from given abstract, without associated value.
+
+    Useful to lookup node abstract in registered signatures.
+
+    NB: We must remove value track, not keeping it with ANYTHING value,
+    because abstract atom with only interface won't have same hash
+    as abstract atom with interface + ANYTHING value.
```
breuleux, line 92: True, but that may need fixing...

myia/infer/infnode.py
```diff
@@ -80,6 +80,18 @@ def _type_to_abstract(el):
...
+    """
+    return data.AbstractAtom({"interface": abstract.tracks.interface})
```
breuleux: You're removing all tracks that are not `interface`, not just `value`.

myia/infer/infnode.py
```diff
@@ -134,12 +152,27 @@ def inference(node, args, unif):
...
+                    inp_types, inf_def.arg_types
+                ):
+                    autils.unify(inp_type, expected_type, U=unif)
+                return autils.reify(inf_def.ret_type, unif=unif.canon)
+        else:
+            assert default, (
+                f"No inference for node: {node}, "
+                f"signature: {list(inp_types)}"
+            )
+            return (yield from default(node, args, unif))
```
breuleux: It may not be a bad idea to cache the result in `def_map`.

tests/ir/test_print.py
```diff
@@ -105,47 +105,47 @@ def test_format_exc():
...
        ^
 
-File ./tests/ir/test_print.py, line 85
-In add:clone(x::*int(), y::*float())
-85     return x + y
-              ^^^^^
+File ./tests/test_parser.py, line 2024
+In and_:clone(x~2::*int(), y::*float())
+2024     return f"value is {x}"
+         ^^^^^^^^^^^^^^^^^^^^^^
```
breuleux: Why is it highlighting a line in `test_parser`?

myia/infer/inferrers.py
notoraptor: Ok. If I deactivate these lines, tests with `fact` will fail in `x * fact(x - 1)`. It seems it's because of `fact(x - 1)`, whose type is still Generic at this step.

I may just add a TODO here with your remarks, while keeping current code. Is it ok, or should I mark related tests as xfail in this PR ?

myia/infer/inferrers.py
notoraptor: Ok!

myia/infer/infnode.py
notoraptor: Ok!

myia/infer/infnode.py
notoraptor: Ok, I add a TODO here.

myia/infer/infnode.py
notoraptor: Indeed, fixed, only value track removed.

myia/infer/infnode.py
notoraptor: Ok !

tests/ir/test_print.py
notoraptor: I still don't know, neither.

notoraptor:
Hi @breuleux , I add a commit to try to fix `format_exc` parsing. What do you think ?

myia/utils/info.py
```diff
@@ -40,6 +44,8 @@ def top(self):
 def enable_debug():
     """Enable debugging for a context."""
     tok = _debug.set(True)
+    # Clear stack each time we call enable_debug.
+    _stack.clear()
```
breuleux: I think the problem happens because when an exception is raised in `with debug_inherit(...):`, the stack pop does not happen, so a better solution would be to add a `try: finally:` around the `yield` in `debug_inherit`, with the cleanup code in the `finally` block.

notoraptor: OK, done. I still have 2 uncovered lines in `myia/ir/print`.

notoraptor:
Coverage fixed!

